PR #2, opened 2020-03-30
Darodrig.c

d-r-e:


d-r-e (commented):
Fixed '#' prefix

d-r-e (commented):
Added a description of the code.

clafoutis42:
This is a good solution, however you could have allocated directly the size of the wanted string, and modify it instead of using `strdup`.
Also you should have commented the solution in the pull request too.

clafoutis42:
The file was not entered in the good subdirectory. Your PR will be reverted.

PR #3, opened 2020-03-30
Solution

alvrodri:


alvrodri:
Ejemplo con 255, 255, 255:
· Muevo 'r' 16 bits a la izquierda (ff0000), muevo 'g' 8 bits a la izquierda (ffff00), muevo 'b' 0 bits a la izquierda (ffffff)
· Hago un itoa base para convertir el número decimal a hexadecimal

clafoutis42:
That's a good solution, and a great use of `itoa_base`, however you should always comment your code in english. As well as you should comment the PRs in english.  
Any further admission in an other language won't be accepted.

PR #4, opened 2020-03-30
Pull mcarrete

mcarrete42:
To solve this challenge I first had to figure out how the hex value is done, that even if I work on a daily basis with, Ihad never asked myself that question (I would just copy over the code!).
Basically, it is the int to hex function of each of the values (r,g,b) adding a # in fornt of it. One "if" is that if the hex value "strlen" is less than 2 (i.e is from 0 to 9), you need to add a 0 to that hex. 
 What I have done is use the itoa_hex function which converts the int to hex and saves it in a string, for each number, and then check if the strlen of any of those is < 2. If so, add a 0. 
Then I have  joined the "#" with char *r, the result with char *g, and the result with char *b.


clafoutis42:
The solution seems to work, however it is really greedy in memory, using multiple strjoin and never freeing them is really bad.  
Also using multiple iterating function your solution might be really slow on a greater problem.

mcarrete42:
Thank you for your feedback Pistache!!!

PR #5, opened 2020-03-30
chall00 - gbudau

gbudau:
My solution was to allocate memory for the string we have to return (8 characters, including terminating null character)

Then create the final color using the RGB values  by shifting the bits of each additive color so red will be on second byte, green on the third byte and blue on the forth byte of an integer, assuming each individual color (red, green, blue) is one byte each, 0 to 255 in decimal value.

Since were interested in the last 3 bytes of the color in base 16 -hexadecimal- we use modulo to get the last digit.

If the remainder we get is more than 9, we substitute by 10 and add the result to the character 'a' to represent  it as a letter, if is between 0-9 we add it by '0' to convert it to a digit.

Then divide the color by 16 and repeat the  modulo/divide process 6 times to get all of the last 3 bytes.

Add '#' character and '\0' to end the string.

Since the string  will be in reverse order because we started from the end, we have to reverse it to display properly.

### DIFF
--- a/chall00/gbudau.c
+++ b/chall00/gbudau.c
@@ -1,0 +1,55 @@
+#include <stdlib.h>
+
+/*
+** simple swap function
+*/
+static char	*str_swap(char *str, int start, int end)
+{
+	char	tmp;
+
+	while (start < end)
+	{
+		tmp = str[start];
+		str[start++] = str[end];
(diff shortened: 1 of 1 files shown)


clafoutis42:
Great explanation :)

PR #9, opened 2020-03-30
chall00 - gsanchez

GonzSanch:
Submitted function solution using memory, stupid hard way to review concepts


GonzSanch:
In hv_rgb2hex function firts at all check valid values for rgb then create the returned char* allocating enough memory and started with '#' char and ended with Null. The rgb values to hex are procesed in an aux  function, this function takes the value, char* to fill and the index. If the value is greater than 16 the function will handle recursively else will add a zero before the value minor than 16.

GonzSanch:
Probably I should replace "error" to NULL in L22 to handle error if rgb are negative or greater than 255..

GonzSanch:
Changed return NULL instead "error", make flag static to work with recursivity and ajusted correct size of char* in allocation.

anna-dominguez:
Too bad you've done a push after the limit hour...  Of what I can see you have done case of errors, protection of malloc. The function is failing on the output, you're missing the 0 written before a number.  I think you're close to the good answer !

PR #11, opened 2020-03-30
chall00-fulldemo

fulldemo13:
To solve it I converted the binary numbers to hexadecimal and I have been concatenating them all in the string "res". For this I used functions learned in 42. I paid attention to the limits of the numbers too.

### DIFF
--- a/chall00/chall00-fulldemo.c
+++ b/chall00/chall00-fulldemo.c
@@ -1,0 +1,94 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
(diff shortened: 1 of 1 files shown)


clafoutis42:
Good solution. However concatenating the string each time is really greedy with memory and execution time.  
A better solution would have been to allocate the memory for the all string and fill it later with the hexadecimal conversion.

fulldemo13:
Thanks for the tips, next time will be better ;-)

clafoutis42:
I just noticed that the filname was not of the format `xlogin` your PR will be reverted.

fulldemo13:
Don't worry nobody is perfect, hahaha

PR #12, opened 2020-03-30
chall00 add

dalba-de:
Para resolver el ejercicio convierto los 3 colores dados a un unico valor RGB mediante la formula:

r * 256 * 256 + g * 256 + b.

Ese número lo convierto a hexadecimal mediante una función itoa base y lo devuelto en una cadena.


clafoutis42:
The solution is working but does not respect the expected output. `#ff` should be returned as `#0000ff`.  
Also you should always submit your PRs in english. For the next challenges, any PR in any other language than english won't be accepted.

PR #13, opened 2020-03-30
Antomart

miyomovil:
To solve it I decided to make and "int_to_hex" function (recicling my old itoa_base) . Then I customized my own strjoin to join the 3 values and add a '#' on the string. The function only accepts valid values for the input.


clafoutis42:
The solution seems correct. However you making multiple joins and using multiple iterating functions is really greedy in memory and execution time.  
A better solution would have been to allocate the expected size (as it is always the same) and fill the string with the conversion.

PR #14, opened 2020-03-30
vde-dios

vdedios:
The challenge is solved by applying ternary operators and bitwise operations. Although it may seem a little bit intimidating at first glance, it's very straight forward: based on the idea that a byte can be seen as 8 single bits or as 2 "hex bytes". So, masking each byte with 0x0F will get the lowest 4 bits and the same with 0xF0 and the upper ones. No need to divide by 16!


clafoutis42:
The solution seems good, and this is a great use ouf binary operations.  
However you should prefer to return `NULL` so it is more readable by other users.

PR #18, opened 2020-03-30
Challenge 00: Ojo de Sauron / RGB to Hexadecimal by eligero

eligero:
/*
** Ojo de Sauron (Chall00) / RGB to Hexadecimal
**
** Solved with embedded system mindset.
** Integer focus range is [0, 255], it means 2 ascii characters. That is the
** output string always will have 7 characters ('#' + 3 x 2 character), so I
** created dinamically an array of 8 positions, in order to null terminating the
** string.
** I did a fast bit-level algorithm to translate number to hex character.
** By right displacing four times, at bit level, de variable is the same as
** divide by 2^4 = 16 the variable, but faster, e.g from 0xAF i get 0xA.
** By doing the bit level AND with 0x0F i get the remainder, e.g from 0xAF
** i get 0xF. In order to store any hex digit 4 bits are needed.
** Finally returning the dynamic allocated string.
*/

### DIFF
--- a/chall00/eligero.c
+++ b/chall00/eligero.c
@@ -1,0 +1,54 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   eligero.c                                          :+:      :+:    :+:   */
+/*                                                    +:+ +:+         +:+     */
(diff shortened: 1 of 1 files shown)


aguimbao:
Only one allocation, explained bitwise operations flow, code is simple and it just works, gg.

PR #19, opened 2020-03-30
chall00 - mmartin-

Mariomm-marti:
Problem approach:
Converted the number to HEX using the function [sprintf](https://www.gnu.org/software/libc/manual/html_node/Formatted-Output-Functions.html) and storing the result in heap using [strdup](http://man7.org/linux/man-pages/man3/strdup.3.html)
How I did it:
- Checked whether the input was valid
- Converted the whole input to a single int by bit shifting, then apply format using sprintf (with a precision of 6 chars in case the input is too low)
- Duplicate result in heap

### DIFF
--- a/chall00/mmartin-.c
+++ b/chall00/mmartin-.c
@@ -1,0 +1,45 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   mmartin-.c                                         :+:      :+:    :+:   */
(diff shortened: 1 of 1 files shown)


clafoutis42:
The solution seems great, however it seems to cause a segmentation fault :(

aguimbao:
Btw next time, write your PR explanation in english 🇺🇸 @Mariomm-marti

PR #20, opened 2020-03-30
Edramire - Ojo de Sauron

edramir18:
### Edramire - Challenge 00
**Description**
The function take every argument and convert to hex using bitwise operators. Each integer number between 0 and 255, can be represented by two hex number on hex notation; using the bit operator AND and a the mask 0xF, for the rightmost number, the bit operator SHIFT and AND with the mask 0xF on the leftmost operator, i can get the representation from each hex digit. In case a number is greater than 9, that number is added to the ascii character 87, to generate the letter on lower case, otherwise the number is added to the ascii character 48, to generate the numbers between 0 and 9. 
**Return**
The final string is created with malloc and returned to the calling process. 
In case any number are negative or the function is unable to reserve enough memory for the string, the NULL value is returned as result.


clafoutis42:
The solution seems to work, great use of the binary operators!

PR #22, opened 2020-03-30
El ojo de Sauron

fjimenez81:
Primero he convertido los números a hexadecimal procurando que si el numero es menor de 16 meter un cero delante, luego meto el # y uno cada una de las cadenas resultantes en orden.


aguimbao:
Too many allocations, code doesn't seem too performant. You are running out of bounds when provided with (0, 0, 255) or similar because of a lack of checks.

Next time, write your PR explanation in english, as everything else will be discarded.

PR #23, opened 2020-03-31
fgata-va solution

FranciscoJesusGata:
## How do I solved it
- Fist I check if the input values are between 0 and 255, if not the function will return `NULL`.
- Then I allocate the string that the function will return while I check the `malloc` output.
- The first character output '#', is assigned and followed by a NULL (if it doesn't strcat returns a wrong value).
- I pass the each input value into a function multiple times that transforms the value to hexadecimal and concatenate it to the final string that the function will return.

## How my code works
The code includes 4 functions, the `hv_rgb2hex` function and other 3 extra functions:
- The function `hv_rgb2hex` checks the input and allocates the string the string that will later return.
- The function `join_n_free` concatenates the existing output string with the number we received converted into hexadecimal and free unused memory.
- The function `dec_to_hex` converts the number received in decimal to hexadecimal as a string.
- The function `ft_rev_str` is used to reverse the resulting string that the `dec_to_hex` loop returns.

### DIFF
--- a/chall00/fgata-va.c
+++ b/chall00/fgata-va.c
@@ -1,0 +1,86 @@
+#include <unistd.h>
+#include <stdlib.h>
+#include <string.h>
+
+char			*ft_rev_str(char *str)
+{
+	char *result;
+	int i;
+	int j;
+
+	if(!(result= (char*)malloc(sizeof(char) * strlen(str)))){
+		return(NULL);
+	}
+	i = 0;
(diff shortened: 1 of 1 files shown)


clafoutis42:
The solution seems to work. However it is really greedy in memory allocation. A better solution would have been to allocate the size of the string first, and then fill it with the conversion.

PR #24, opened 2020-03-31
chall 00 (Ojo de Sauron)

misterjsr:
I have basically solved the exercise, passing the RGB values to hexadecimal and concatenating them into a string. Before I have made sure that the values are correct and there are no memory problems.

### DIFF
--- a/chall00/jeserran.c
+++ b/chall00/jeserran.c
@@ -1,0 +1,51 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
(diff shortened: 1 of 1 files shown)


anna-dominguez:
You should always submit your PRs and your code errors in english. For the next challenges, any PR in any other language than english won't be accepted.
But this is working great, doing verification of the function inputs, of malloc. Good job !

misterjsr:
Ok but the exercice's name was "sauron's eye" in spanish...  thanks for all¡¡¡¡

PR #28, opened 2020-03-31
Add files via upload

victorsempere:
Put together in a single 4B int the 3 params in a row. Truncated to 255. Then a simplified itoa function will produce the expected string


aguimbao:
The deadline was yesterday, at 11:42 a.m. :( 

So your PR was like 12h late. 
Stay tuned for the next challenge though !

PR #29, opened 2020-04-01
Out of Deadline But works ;)

byrd92:
I made math change RGB to number, then itoa the number to Hex (size stablished to hex RGB number (8)).


aguimbao:
Neat solution 👍🏻

Sorry for not being in time though :(

PR #30, opened 2020-04-02
chall01-jbaringo

baringo11:
Primero compruebo que las dos strings tengan la misma longitud, luego voy comprobando 1 por 1 que la segunda string siga el mismo patrón que la primera y cuando coincida devuelvo 1 y si no coincide ninguna devuelvo 0.

### DIFF
--- a/chall01/jbaringo.c
+++ b/chall01/jbaringo.c
@@ -1,0 +1,56 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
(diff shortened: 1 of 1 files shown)


clafoutis42:
Good working solution. However you could have optimized by avoiding repetitive strlens. Especially knowing that both strings should have the same length.  
As said in slack, this is the last time we will accept contribution in any other language than english, on github you should always contribute in english so that anyone can understand you.

clafoutis42:
Also, comparing a string and "" (e.g `str == ""`) will always return false, you are not comparing the chars of the string, but the pointer values.

PR #31, opened 2020-04-02
vde-dios.c

vdedios:
The resolution is straight forward. Considering for example the word "hello", a correct necklace would be:

hello
o-hell
lo-hel
(..)

In my  solution i just compare the original string (s1) against the  second doing all the permutations possible. If no mismatches doing comparison are encountered it means that we have a correct necklace. Otherwise, if  any mismatches is encountered and we end with all possible comparison, it will be a wrong necklace.


vdedios:
This pull request is deprecated since I chose the wrong branch to merge with. Just made a new good one to master. Sorry for that!!

PR #32, opened 2020-04-02
vde-dios.c

vdedios:
The resolution is straight forward. Considering for example the word "hello", a correct necklace would be:

hello
o-hell
lo-hel
(..)

In my solution I just compare the original string (s1) against the second (s2) making all the correct possible permutations (like the example above). If no mismatches are encountered while comparing a single combination it means that we have a correct necklace. Otherwise, if any mismatch is encountered and we have ended with all possible combinations it means that s2 does not come from s1 and it will be a wrong necklace.


clafoutis42:
Good working solution. However you should split your code in multiple functions to make it more readable.  
Also you could have used a solution like modulo or something to rotate while comparing the strings, instead of making 2 loops.

vdedios:
All right, I will consider it for the  next time! Thanks  for your time

PR #34, opened 2020-04-02
estina

xplore93:
The main idea is that all the characters should be in the same order. The structure of the program is:

- Check the size of both strings
- Check if the strings are empty
- Starting from 0 we check both strings while s1 is not ended.
So, for example, if we have 'Hello' and 'lloHe', we will check each character while they are not equal. When 'H' is found in the second string, we will check in order all the next characters, and if we end the second string while the first one is not ended and they are equal, we assign 0 to our s2's index and continue the comparison while s1 is not ended.

Return 0 if the strings are not equal, 1 if equal.

### DIFF
--- a/chall01/estina.c
+++ b/chall01/estina.c
@@ -1,0 +1,63 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   estina.c                                           :+:      :+:    :+:   */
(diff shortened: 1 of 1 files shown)


aguimbao:
Simple solution, works flawlessly and you work on the strings directly, gg.

I think there could be a slightly more performing solution with some clever tricks, but this just works, gg.

PR #35, opened 2020-04-02
gmartine

GabrielMartinezRodriguez:


GabrielMartinezRodriguez:
Busco en el string 2 la primera letra del string 1 y compruebo si circularmente es igual. Si no coincide repito hasta terminar el string 2.
I look in string 2 for the first letter of string 1 and check if it's circularly the same. If it does not match I repeat until I finish string 2.

anna-dominguez:
Works perfectly ! A good practice to have is to decompose a little your function, the norme isn't mandatory, but it helps a lot to understand your code. Well done not using `malloc`.
Be careful not to use spanish in the next challenge.

PR #36, opened 2020-04-02
Chall 01 antomart

miyomovil:
I decided to create a function that switches the letter of a string "as the necklace does" (swithing the last position of the string to the first position). The function compares the 2 strings and "switches" one of them. If there is no diferences between the strings (after swithing positions in one of them) both necklaces are the same. My function also compares the len of both strings.

### DIFF
--- a/chall01/antomart.c
+++ b/chall01/antomart.c
@@ -1,0 +1,102 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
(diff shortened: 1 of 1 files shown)


aguimbao:
You could have spared a strlen() call in hv_necklace(), you already had the len value of str1.

It's a decent approach, but not the most performant for edge cases: if the string is really big, or if str2 is cut at the end. You use strdup(), so malloc(), and the exercise could be completed without malloc !

Nonetheless, it works well, good job :)

PR #37, opened 2020-04-02
chall01-pcuadrad

pabloocg:
Function int hv_necklace(char *s1, char *s2)
  vars:
      - is1: counter of the s1 string
      - is2: counter of the s2 string
      - lens1: length if the s1 string 
      - saves1: number of the last difference between characters in the string
      - i: auxiliar counter
  return value (int):
      1 if both strings are equal independently of how they start
      0 if not are equal
  Explanation:
      We compare the first string with the second one until we have a match, 
      once it matches we continue comparing until we reach the end of the second string.
      Once we reach the end we see if something has been saved in the variable
      save1 (it would be that something has not previously matched), if so, we start
      to compare the first string again from the beginning to the character where
      they naturally started to match.

### DIFF
--- a/chall01/pcuadrad.c
+++ b/chall01/pcuadrad.c
@@ -53,14 +53,14 @@ int hv_necklace(char *s1, char *s2)
     int     saves1;
     int     i;
 
-    is1 = 0;
+    is1 = -1;
     saves1 = -1;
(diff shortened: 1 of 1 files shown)


anna-dominguez:
Well done ! Good explanation and work perfectly. Good implementation too 👍

PR #38, opened 2020-04-02
Edramire - Freddiemercury 17473

edramir18:
### Description
In order to resolve the problem, i assume two string are equivalent only if both are the same length and the start of the first string has an offset in the second string.
To calculate the offset in the second string, i need to find the position of the first character of the first string; using pointer arithmetic i get the index and compare the second string from that index with the first one. In case the compare are equal both string are equivalent and return 1 to the calling process, otherwise i try the next occurrence of the first character until i get NULL on the searching function.

### Return
in case the two string are equivalent, the function return 1, otherwise return 0.


clafoutis42:
Really great solution, that is the best one I've seen (so far).  
You could have stored just one of the length, as they are supposed to be the same :)

PR #39, opened 2020-04-02
Chall01

Dapalm:
This function makes use of some of the functions from my Libft (ft_strlen, ft_strdup, ft_strcmp, ft_substr). 

First we check if s1 and s2 are equal using ft_strcmp. If they are we return 1. Else, since s1 and s2 could be string literals, we create a temporal copy of s1 using ft_strdup.

Using a while loop, for every character in the original string we create another string using a slighty modified ft_substr (it receives the character that should be placed in the last position of our new string). On each iteration we check if s1 and s2 are equal, and if they are we return 1. Else, loop will end and 0 will be returned.

Checked using valgrind: no memory leaks.


Dapalm (commented):
Removed unused variables after compiling wit flags Werror, Wextra, Wall.
Made some formatting changes

clafoutis42:
This is a working solution. However it is greedy in memory, you are constantly doing substring and duplications.  
You could have made a rotative comparison using a modulo or something likely, and avoid allocation memory.

PR #40, opened 2020-04-02
fgata-va's solution

FranciscoJesusGata:
# How do I solved it
- First I check if both strings have the same length and are not ` NULL`.
- Then I allocate a src string I'll use to copy the first string and slice n times until it matches with the second output.
- The function iterates as many times as chars the string has (if both strings are equal check first before slice the first character, if not the function will slice all the chars until get the solution).
# How my code works
Because of the previous function I took care of memory usage, this time I have only two functions:
 - The `hv_necklace` function: allocates the `src` string, copy the first string input, iterates over the second function, check if both strings are equal and free the allocated memory.
- The `ft_slide_char` function: define an auxiliar string with the src length, slice the first character and copy the auxiliar string into the src string.

### DIFF
--- a/chall01/fgata-va.c
+++ b/chall01/fgata-va.c
@@ -1,0 +1,45 @@
+#include <string.h>
+#include <stdlib.h>
+
+void		ft_slide_char(char *dst, size_t len)
+{
+	int		j;
+	char	aux[len];
+
+	j = 0;
+	while (dst[j + 1])
+	{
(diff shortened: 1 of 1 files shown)


clafoutis42:
Interesting solution, however you could have used a rotative comparison (using the length as a modulo, or something likely) to optimize your solution and avoid using a malloc.  
Watch out for the empty string, they're a necklace too ;)

PR #41, opened 2020-04-02
Added Chall01.

byrd92:
I malloc max size of string (strlen * 2). 
Make a copy on aux and start while compare. 
If string are the same ret 1, else aux get first world at last and move *aux 1++ and repeat!

/* en español

No se si tengo que quitar del pull request el antiguo (chall 00), tampoco se muy bien como hacerlo.
Salut!

### DIFF
--- a/chall00/egarcia-.c
+++ b/chall00/egarcia-.c
@@ -1,0 +1,38 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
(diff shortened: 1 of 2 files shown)


clafoutis42:
That's a good solution. It's just a little greedy on memory and operations as you might make a swap for each characters of the string if the match comes late.  
A better solution would have been using a rotative comparison, with a modulo or something likely.

PR #43, opened 2020-04-02
solve hv_necklace

javrodri42:


clafoutis42:
Good working solution. However it might be long to rotate your string in some cases where the string might be long and the match might me the last rotation.  
You could have made a comparison using modulos, for example.

PR #46, opened 2020-04-02
Chall01 migferna

migferna42:
First we look for the character of the second string in the main string, to get the position. With the calculated position, it is checked if the characters match in an ordered way, in the case that it does not match, another position is searched. 
*Previously used positions are marked to not be used again. If no valid position is found, the function is terminated.


clafoutis42:
The solution seems to work, but with various problems:
- You will never really use the mask, as your exit conditions could be reaching `\0` in one of both strings, or a match  
  Basically with your mask you just added redundant conditions and an unnecessary malloc.
- You are using strlen more than 2 times instead of storing the length.

You should split your code in multiple functions in order to make it more readable.

PR #48, opened 2020-04-02
Chall01 fulldemo

fulldemo13:
For this exercise I look if the length of the two strings is equal, if it isn't I return 0.
Next, I look for the first coincidence in the two strings and it starts to compare char by char.
If one fails  the comparator restarts in the next coincidence.
When the counter flag is equal to the length of the string the function return the result.


clafoutis42:
That's a good solution, but it could have been implemented in a more simple way.  
You could simply loop on the occurrences of the first char, and make a rotative comparison like you did or with a modulo.

PR #50, opened 2020-04-02
chall01 dalba-de

dalba-de:
First I check if both strings are the same length. Then I go through the first string looking for a character match with the second one and from that point I try to follow the pattern in both strings. If the pattern is wrong, break and continue trying with the next position of the first string.
If the counter of the second string gets to be the same as the length of s2, it means that it has found the complete pattern and returns 1. In case of not getting it at any time, return 0. Finally I check the special case that both strings are empty.

### DIFF
--- a/chall01/dalba-de.c
+++ b/chall01/dalba-de.c
@@ -1,0 +1,35 @@
+#include <string.h>
+
+int hv_necklace(char *s1, char *s2)
+{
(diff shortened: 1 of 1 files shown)


clafoutis42:
Good working solution, however you should have stored the len to avoir repetitive repetitive strlens.

PR #53, opened 2020-04-02
PR 2

JavierCervilla:
Okey, sorry for my english.
we know we have to iterate as times as the len of the s1, both of them has to be the same lenght.
we have 2 counters to see the iterations, the first one has to be a temporal counter to see if are right, if not we have to reset this counter but the other counter has to be fix.


clafoutis42:
You modified the README in your PR. We won't accept this PR.

JavierCervilla:
I modified It because It wasnt Accept muy fork i dnt know why, i copied it
from the raw GitHub...

PR #54, opened 2020-04-02
Chall01

Top64:
Hello, my login name is cdorta-n.

Let me explain how I face this problem.

> Understand the main idea

We need to check two strings, the first one is our source, and the second one, the one to be checked. We need to see if each letter in the second string, is consecutive, equal to the first string that set the correct distributtion.

At first, the idea was check the distributtion, but is couldn't solve the examples. Then, with help from others Friends of 42, we get a solution, check each variation so the second string its the same as the first one.
If the solution is found, the program return 1.

### DIFF
--- a/chall01/cdorta-n.c
+++ b/chall01/cdorta-n.c
@@ -1,0 +1,115 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   cdorta-n.c                                         :+:      :+:    :+:   */
+/*                                                    +:+ +:+         +:+     */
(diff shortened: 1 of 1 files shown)


anna-dominguez:
It compile and is working but you've got a warning that you can resolve easily 😞 you've written your `ft_strlen` under a function calling it and trying compiling it says: 
`student.c:60:33: warning: implicit declaration of function 'ft_strlen' is invalid in C99 [-Wimplicit-function-declaration]`
You have a lot of `malloc` unnecessarily too. Playing with the two strings lengths you can resolve the problem

PR #56, opened 2020-04-02
Challenge 01: Freddiemercury 17473 / Necklace

eligero:
/*
** Freddiemercury 17473 (Chall01) / Necklace
**
** Solved as fast as possible, reusing libft, and obeying the 42 norme.
** Malloc used in order to not overflow the program stack.
** Algorithm takes track of the point of rotation on s1, by using i. Firstly It
** s1[i]...s1[len - 1] into p[0]...p[i], after that it copyies
** s1[0]...s1[i - 1] into p starting at p[i + 1]. Generat all posible rotations
** of the string s1 on the null terminated string p.
** If one of this posible rotation strings generated on p, matches the string
** s2, by using ft_strcmp, then returns 1. Algorithm finishes once occur a
** match or by it has done all posible rotations.
** In order to speed up the algorith, it only calls ft_strcmp if p[0] == s2[0]
**
*/

### DIFF
--- a/chall01/eligero.c
+++ b/chall01/eligero.c
@@ -1,0 +1,93 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   eligero.c                                          :+:      :+:    :+:   */
+/*                                                    +:+ +:+         +:+     */
+/*   By: eligero <eligero@student.42.fr>            +#+  +:+       +#+        */
+/*                                                +#+#+#+#+#+   +#+           */
+/*   Created: 2020/04/02 15:01:31 by eligero           #+#    #+#             */
+/*   Updated: 2020/04/02 17:21:49 by eligero          ###   ########.fr       */
+/*                                                                            */
+/* ************************************************************************** */
+
+#include <stdlib.h>
+
+/*
+** Freddiemercury 17473 (Chall01) / Necklace
+**
+** Solved as fast as possible, reusing libft, and obeying the 42 norme.
+** Malloc used in order to not overflow the program stack.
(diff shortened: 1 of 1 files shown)


eligero:
This is not the best solution, of course, but it depends on how you conceive the problem. If you don't reserve a portion of memory, on stack or heap, you will have redundancy on memory if you write on s1, as s1 and s2 will contain the same string if the program retuns 1, or s1 will have rubbish if the program returns 0. My solution is conceived with the approach in mind of build your own stuff in your function and leave all the given data intact. In the scenario presented by solving this problem writing over s1, without memory allocation, it's possible that s1 is on portions of memory that you don't have write permission, in wich case my solution remains working and it performs good. By the way, approach done by a 42's colleage using just strncmp it's a great solution, maybe the best.

aguimbao:
It definitely depends on how you approach and conceive the problem indeed. Usually this kind of small problems are solved with performance and simplicity in mind and with the less amount of code possible. It's true that the student that you are probably referring to didn't have in mind memory ownership, which probably won't give him any problems in general, but it's good to know and anticipate all of this issues you talk about.
I think when programming in general, security shouldn't be taken lightly, it should be as important if not more, than performance, along code simplicity. Your solution accounts for a lot of edge cases and I think that's a good thing, I left it for the end just to see what your peers had done. In the end, you need to find a balance, you can't make completely safe code, as there's generally a lot of environment variables out of your control, sometimes it's more important to aim for a less-secure, but also less-time consuming and bit riskier solution. I think your usage of malloc is justified though !

As for your code, I really like your explicit casting, pointer arithmetic and your general knowledge of keywords and C in general. You pass all the tests. :) Good job 👍

PR #57, opened 2020-04-02
hv_necklace isfernan

isabelfdez:
We have to study whether the 2nd string (s2) given is a rotation of the 1st (s1) or not. To do this, I created a new string, s3, where I copied s2 TWICE. Therefore, if s2 really is a rotation of s1 we should be able to find s1 in s3, and now I just have to use a variation of strnstr (which returns an integer; 1 if s1 appears in s3 and 0 in the other case) to finish the problem.

### DIFF
--- a/chall01/isfernan.c
+++ b/chall01/isfernan.c
@@ -1,0 +1,103 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   isfernan.c                                         :+:      :+:    :+:   */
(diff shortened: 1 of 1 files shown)


anna-dominguez:
Working great, the idea is good but you could have think a lot more simple. Using `malloc` is not an obligation in the resolution and make higher the difficulty and slower the result. Comparing the two strings length and then search where the rotation is starting in the initial string untill the end is a better and simpler way to achieve it 😉

PR #58, opened 2020-04-02
Necklace 1.0

jtrancoso:
First I get rid of the simplest cases like different strings and the same strings. Then I look if it has only been one move and then I start looking for the rest.

It doesn't work in all the cases yet, I'll try to fix it before deadline.


anna-dominguez:
You still have some work to do, your code isn't working on all cases. Try with this one `"rabcr", "bcrra"`. Good luck !

PR #59, opened 2020-04-02
chall01 - gbudau

gbudau:
***How i solved the problem.***
Compare the end of the first string with the beginning of the second, then compare the beginning of the first with the end of the second. If necklace is found the elements at last indexes for both strings should be equal to `'\0'`.

***Description of the code.***
Iterating first string using `i` as an index to find the part that coincides with the second string.
When we find the part that coincides we use `j` as an index for the second string and `i + j` as index for the first string to compare them.

If the first or second string index points at the null character, exit the  loop.
Else reset `j` (second string) and increment `i` (first string) and keep checking.

 Since we already compared the last part of the first string with first part of the  second string we now need to do the oposite.

Save the length compared until now to `checked` variable and reset the index of first string.

 If the elements at last indexes for both strings are equal to `'\0'` `return 1` if not `return 0`.

### DIFF
--- a/chall01/gbudau.c
+++ b/chall01/gbudau.c
@@ -1,4 +1,4 @@
-int check_necklace(char *s1, char *s2)
+statric int check_necklace(char *s1, char *s2)
 {
 	int	i;
        	int	j;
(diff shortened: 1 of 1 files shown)


anna-dominguez:
Your solution is working perfectly. I love the way you proceeded to resolve this and the way you've code it, it's very simple 😄Good job !

gbudau:
Thank you! :)

PR #60, opened 2020-04-02
mpernia-

mmiguelx:
Hello.

1. Check if s1 or s2 aren't null strings and both strings have the same length, return 0 if any of those statements are false.
2. To avoid use unnecessary computing power, if both strings are the same return 1.
3. Find the sub-string inside s1 with an aux pointer which starts on s2, if you find it compare the rest of the string.

### DIFF
--- a/chall01/mpernia-.c
+++ b/chall01/mpernia-.c
@@ -1,0 +1,95 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   mpernia-.c                                         :+:      :+:    :+:   */
+/*                                                    +:+ +:+         +:+     */
+/*   By: mpernia- <mpernia-@student.42madrid.com    +#+  +:+       +#+        */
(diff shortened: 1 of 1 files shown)


aguimbao:
Nice ! Not the most performing but an easy and straightforward solution, no allocations needed.

Good job :)

mmiguelx:
Thank you for your review. Do you have any suggestions to improve my perfomance skills?

aguimbao:
Well I saw this really nice approach: 

You malloc a new str with 2 times the size of one of the original strings after doing the initial checks. And you duplicate the string. So `necklace` becomes `necklacenecklace`. And you check with one of the two strings against this one, this way you never have to go back to the index 0 when you reach the end of the string, and you don't need to constantly swap letters. I don't know if it was more performant as I didn't do any benchmarking.
I took another look at your code and it's honestly very well written, much better than the average C user in 42.

mmiguelx:
Ofc, so you have to check only once. That's very good! 
Thank you for all your answers, @abguimba.

PR #61, opened 2020-04-02
mrosario chall01

mikirosario:
Ok, so basically I use ft_strncmp and ft_strlen. ^_^ (When copying ft_strncmp from my libft I renamed it ft_strcmp at some point... err... but really it is ft_strncmp xD)

First, I check to make sure that both of the string pointers passed as arguments are not void and that both strings have the same length - otherwise I return 0. I save the length in the len variable.

Then I check to see if the first character of s2 is the same as the first character of s1, and if they are the same I use ft_strncmp directly, and if it returns 0 then obviously I return 1 because that's a match!

If they are not the same, then I use an index (i) to advance through the characters of s2 until I find a match for the first character of s1. Then I just take advantage of the index.

First, I pass ft_strncmp the address of the first char of s1, the address of the s2 char that matches the first char of s1 (s2[i]), and then I tell ft_strncmp to only check len - i characters, so I will subtract however many characters I jumped in s2 from the number of characters I will check against s1.

Then if this is a match, I will do precisely the inverse, I will pass ft_strncmp the address of s1 in the position len - i (s1[len - i]) - that is, the position of the character after the last character of s1 to be checked in the first check - and the address of the first character of s2, and I tell ft_strncmp to check for precisely index characters, so now I compare all the characters I missed in the first comparison.

So basically, I simply look for all the instances of the char at s1[0] in s2, and then I divide the ft_strncmp comparison into two halves. The first half looks for a match in the string starting from the character in s2 that matches the first character in s1 and until the end of s2, and the second half looks for a match in the string starting from the first character in s2 and starting from the character in s1 after the last one to be compared in the prior check, and until the end of s1.

If the strings are matching in both checks, then I return 1. If they are not, I keep looking until there are no further instances of the first character in s1 within the string s2. If I cannot find a match this way, I return 0, of course.

That's it. ^_^

### DIFF
--- a/chall01/mrosario.c
+++ b/chall01/mrosario.c
@@ -1,0 +1,68 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   mrosario.c                                         :+:      :+:    :+:   */
+/*                                                    +:+ +:+         +:+     */
+/*   By: mikiencolor <mikiencolor@student.42.fr>    +#+  +:+       +#+        */
+/*                                                +#+#+#+#+#+   +#+           */
+/*   Created: 2020/04/02 21:21:08 by mikiencolor       #+#    #+#             */
+/*   Updated: 2020/04/02 21:21:16 by mikiencolor      ###   ########.fr       */
+/*                                                                            */
+/* ************************************************************************** */
+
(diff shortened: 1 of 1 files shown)


anna-dominguez:
Thank you for your explanation of every steps of your code 😉 It's working perfectly !

PR #62, opened 2020-04-02
mmateo-t

MiguelTolino:
At first, I checked the error handling. In case one of the strings have different lengths, the function returns 0. Then I created a checking algorithm whose mission is to verify that all characters are in the correct order.
hv_necklace will return 1,if not, it will return 0.


anna-dominguez:
Working fine, but I think it could be simplified a little in your `checking` loop 😄

PR #63, opened 2020-04-02
chall01 - gsanchez

GonzSanch:
First we check if s1 and s2 are null, that will be a valid necklace later we exclude all that have different length.Found the start of s1 in s2 and then compare s2 in index start with s1 and the following chars of each string if the index of s2 exceed length just put a zero and end the comparison. If any dont match will return 0 and back to the iteration of index start that will compare again with start position plus one untill reach the length, if a match its completed then the comparison function will return one.

### DIFF
--- a/chall01/gsanchez.c
+++ b/chall01/gsanchez.c
@@ -1,0 +1,46 @@
+int	ft_strlen(char *str)
+{
+	int	len;
+
+	len = 0;
+	while (str[len])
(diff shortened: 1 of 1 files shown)


anna-dominguez:
You forgot your header 😞 

Joking. It's working perfectly. Your functions are simple and your program is responding fast. Good job !

PR #64, opened 2020-04-02
Necklace

victorsempere:
Recursive call.
Starting everytime from first char in str1.
Look for first occurence from (x) in str2 of the str1[0] and then try to complete necklace.


anna-dominguez:
I don't think it was the more simple way to resolve this doing it with recursive. But it's working 😄 
Take care to suppress your file of the old challenge next time

PR #65, opened 2020-04-03
ssacrist - Freddie's necklace

esepunto:
A necklace around the feLi, the erseUniv, and verythingE.

(Translated with blood, toil, tears and sweat...)

Well... 

The function takes the strings received as beaded necklaces. The goal is to check if they have the same beads (letters) in the same order, or not. In the first case, it returns 1. In the second, it returns 0.

The function uses three variables: 1 to count and 2 to scroll through the necklaces (strings).

1º It counts the beads of each necklace: if they are different, it returns 0.

2º Checks if both necklaces have no beads. Only check one necklace, because we already know that they have the same number of beads. 

3º Compare the necklaces.

3.a) If the first beads of each necklace are different, look at the next bead of the first one and compare again. In case there is no match for a whole round, return 0.

3.b) If there is a match, check the next bead of each necklace again and again as long as there is a match. 

3.b.1. If there are as many coincidences in a row as there are beads in the necklace, return 1.

3.b.2. If there is a discrepancy, it starts the comparison => 3º Compare the strings.

4º If the function has compared all the possibilities and there is no complete match, it returns 0.

### DIFF
--- a/chall01/ssacrist.c
+++ b/chall01/ssacrist.c
@@ -1,0 +1,37 @@
+#include <string.h>
+
+int hv_necklace(char *s1, char *s2)
+{
+    size_t  i;
+    size_t  j;
+    size_t  c;
+
+    if (strlen(s1) != strlen (s2))
+        return (0);
+    if (strlen(s1) == 0)
+        return (1);
+    i = 0;
+    j = 0;
+    while (i < strlen(s2))
+    {
+        while (s1[i] != s2[j])
+            i++;
+        c = 0;
+        while (s1[i] == s2[j])
+        {
+            if (i + 1 == strlen(s2))
+                i = 0;
+            else
+                i++;
+            if (j + 1 == strlen(s2))
+                j = 0;
+            else
+                j++;
+            if (c != strlen(s2))
+                c++;
+            else 
+                return (1);
+        }
+    }
+    return (0);
+}
--- a/chall01/ssacrist.c
+++ b/chall01/ssacrist.c
@@ -19,11 +19,11 @@ int hv_necklace(char *s1, char *s2)
         c = 0;
         while (s1[i] == s2[j])
         {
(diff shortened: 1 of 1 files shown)


esepunto (commented):
(deleted)

esepunto (commented):
(deleted)

chall01/ssacrist.c
```diff
@@ -19,11 +19,11 @@ int hv_necklace(char *s1, char *s2)
         c = 0;
         while (s1[i] == s2[j])
         {
-            if (i + 1 == strlen(s2))
+            if (!s1[i + 1])
```
esepunto, line 22: To make the program lighter and more understandable.

esepunto, line 22: To make the program lighter and more understandable.

chall01/ssacrist.c
```diff
@@ -19,11 +19,11 @@ int hv_necklace(char *s1, char *s2)
...
                 i = 0;
             else
                 i++;
-            if (j + 1 == strlen(s2))
+            if (!s1[j + 1])
```
esepunto: To make the program lighter and more understandable.

aguimbao:
I really like this solution. It's smart, fast, and easy to understand. Only performance improvement I could see is the redundant call to strlen(s2) in every iteration of the main comparation loop:

https://stackoverflow.com/questions/11365039/will-strlen-be-calculated-multiple-times-if-used-in-a-loop-condition

Sadly, you have a possible global-buffer-overflow in line 17 and your solution won't pass the tests, be careful next time and remember to test thoroughly !

**Example overflows:**
hv_necklace("a", "b");
hv_necklace("AAA", "aaa");

esepunto:
Many thanks for your comments, Abraham. I've seen my mistakes, I understand and corrected them.

PR #66, opened 2020-04-03
hv_necklace func

Antogor:


anna-dominguez:
It crashed on the first case I tried `"nicole", "icolen"`. 
Please do a description of your code in english to explain how you resolve it, or how you thought resolving it. You made a very difficult code, think of it in more simple way. Write how to do it would help you a lot simplify, for exemple you didn't need `malloc` at all. Work with the strings and the length of them will do the job. 
You should practice on this 😄

PR #67, opened 2020-04-03
chall01 - mmartin-

Mariomm-marti:
**Problem approach:**
* I thought about several solutions, including swapping one by one strings but I thought it was high time consuming so I dropped it
* Also thought on doing substr and joining results, but again high time consumption
* Finally, I thought of a "creating" a loop where the string was like a necklace and it repeated again. So nicole could be every solution without having to generate them

**Code implementation:**
1. First of all, I checked for invalid input or same string twice
1. I duplicated the string, "nicole" being "nicolenicole
2. I loop through the string until reaching length of `s1`, that would mean I started again
3. If I compare `duplicated + current iteration` and `duplicated + current iteration + length of s1` with `s2` and they are the same, that means I reached a valid solution

**Example:** (looking for `coleni` in `nicolenicole`)
> nicolenicole
> nicole------
> -icolen-----
> --coleni----

### DIFF
--- a/chall01/mmartin-.c
+++ b/chall01/mmartin-.c
@@ -6,7 +6,7 @@
 /*   By: mmartin- <mmartin-@student.42.fr>          +#+  +:+       +#+        */
 /*                                                +#+#+#+#+#+   +#+           */
 /*   Created: 2020/04/02 16:46:55 by mmartin-          #+#    #+#             */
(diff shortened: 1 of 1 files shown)


anna-dominguez:
That’s a good way to implement your program.
`char strc[strlen(s1) * 2 + 1];` I don't really like to set a variable like this, and that’s what make you use your `strlen` of `s1` more than one time. but your program is still responding fast 👏

PR #68, opened 2020-04-03
chall01 pablomar

PalindromoMeszaros:
1- Firstly I check the length of the two strings. Any pair of strings which hasn't the same length is directly consider invalid.

2 - Then we specify simple, direct conditions for the cases of two empty strings and equal one-character strings.

3 - From there, for a general case, we will be increasing the index in the s2 string and compare it with the first character of the s1 string. If we find a coincidence, we will be increasing both strings character by character. 

4 - From here some different things can happen: If s2 reaches the empty string in this character by character comparison, we restart its index to 0 and check if the equality between characters is still happening. In the positive case we continue comparing till s1 string reaches its end. That case would mean that we have a full coincidence and we would return 1.

5 - It could also happen that the two strings are equal for some characters but suddenly a different one appears. In that case we may restart s1 index to 0, and continue checking in s2 till we find a new coincidence. If we find one, we start the process explained in the previous paragraph again. If eventually s2 reaches the empty character in the coincidence searching process (and not inside the character-by-character process) it means that not more first-character coincidences were found and therefore the two strings are not equal and we return 0.


anna-dominguez:
Good comment about your work. But I have a problem with this example: `assert_results("ror", "ror", 1);` it's because you're iterating your index `j` before using checking `compare(s1,s2,j)` in your last loop

PR #69, opened 2020-04-03
jgalian- chall01

jgalian:
- If s1 or s2 are empty or have different length, return 0.
- If s1 and s2 are the same (even if both are empty), return 1.
- If none of these conditions apply, put first letter of s2 at the end and check if now both strings are the same. Repeat until:
* Both strings are the same (return 1)
* Finish changing all the letters in s2 (return 0)

### DIFF
--- a/chall01/jgalian-.c
+++ b/chall01/jgalian-.c
@@ -1,0 +1,105 @@
+#include	<unistd.h>
+#include	<stdlib.h>
+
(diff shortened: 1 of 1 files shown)


aguimbao:
Similar solution to some of your peers. It's a nice approach, but you could gain some performance by not using malloc that often and comparing your strings on the fly. 

It works though, good job !

PR #70, opened 2020-04-03
Necklace, rarias-p

TH0RlN:


aguimbao:
Nice standard solution, no allocations and no redundant calls to strlen(). 

Even if your code is self-explanatory like in this case, please add a little description of the PR in english, it's good practice :)

PR #71, opened 2020-04-03
mcarrete solution

mcarrete42:
First thing I checked, is if both strings are same lengh. If not, just return 0 and ciao. 
If they are the same...
Locate the first char of s1 in s2 and compare chars ahead and before that same char in s2 string.
If the first char in s1 is the last one no w in s2, dont compare ahead! 
Code commented.


anna-dominguez:
Just in time !
It doesn't work with this exemple `hv_necklace("rabcr", "abcrr")` You should try to resolve it with it. By reading your comments you have certainly confused s1 and s2. You should look for s2 in s1 and not the other way around.

PR #74, opened 2020-04-03
chall01 rprieto-

RafaPC:
It first check that both string are the same length, then it starts to check characters from both string.
On the first string it always starts checking from the first character but on the second string it starts from the first the first time, from the second the second time and so on until the second starts at the null character.
When characters are being checked and the second string reaches the null character, its index is set to 0 so it continues to being check at the start.

### DIFF
--- a/chall01/rprieto-.c
+++ b/chall01/rprieto-.c
@@ -1,0 +1,33 @@
+#include <string.h>
+
+int hv_necklace(char *s1, char *s2)
+{
+	int index1;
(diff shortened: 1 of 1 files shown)


aguimbao:
Standard solution, works well :) Some redundant calls to strlen() that could not be optimised depending of compiler. 

You PR came after the deadline though, so sorry. Stay tuned for the next challenge, I'm sure you will be in time.

PR #75, opened 2020-04-03
Changed approach!

mcarrete42:
Thank you for your correction! 
I had to change the code completely, as I was looking for the starting char, and that obviously didn't work if the char were repeated in the string... aiaiaiiiii!!! 
Now I changed the approach, and it is iterating through s2, and comparing it to s1, it now the same i++ in s2 and start comparing from s1[0]...

Hope it works!


aguimbao:
Simple but effective approach, no malloc, no redundant strlen() calls, and it's nice that you understood why your code was flawed last time, gg

You pass all the tests (You can check it yourself as the test battery is included in the repo), but sadly the commit is out of the deadline so I have to close it :( Good luck for the next challenge !

PR #76, opened 2020-04-03
final update

ccardozom:
This is a final update, I know it's out of time but I wish you could watch it. and see if I can finish the next one on time. thanks
- work with 6 variables
- I allocate memory space and copy a string
- compare the received strings if they are equal or have the same length
- I create a condition to rotate one of the string and check if it meets the conditions
- return 1 is it ok or 0 is wrong


clafoutis42:
You should review your way of submitting, there are conflicts between the master branch and yours that have not been resolved.  
Also it has never been indicated that you can modify the `README.md` as it serves as subject.  

Your solution however, seems to work but is far from perfect:
- You forgot to include the header file to use malloc, you are compiling with warnings
- You are using strlen multiple times, you could have optimized by storing the length as it should never change
- Allocation memory in order to compare strings is a bad practice in a case like this, you could have simply used a modulo or another method to do rotative comparison
- You should separate your code in multiple functions, it will help you and others read it, understand it, and see it with more clarity

PR #77, opened 2020-04-06
chall02 - mmartin-

Mariomm-marti:
**Important note:**
Challenge asks for python3 executable script. I assumed python is in the /usr/bin/env path and that the script has executing permission
Tested on WSL `python2.7` and `python3`

**Challenge approach:**
* I decided to simply create a dictionary which stores each letter and it value
* Then mapping each character to its respective value and appending it to the final string
* Space is simply space
* Any other character that is not a-zA-Z is discarded

**Code implementation:**
* Check for valid input (arg number, arg size, etc).
* Check for each character to be in range of a-z, A-Z or to be a space
* Parse string to convert_to_morse, which has a map for each character with its respective value
* Finally, with a lambda, I map each character ("x") in given string ("to_convert") into a value, and join it to final string which is displayed with print (adding automatically a line break)


clafoutis42:
Good solution. Although you might want to use formatted strings to format your usage output. e.g:
```python3
print("usage: %s <a-zA-Z string>" % sys.argv[0]) # printf style formatting with modulo operator
print("usage: {} <a-zA-Z string>".format(sys.argv[0])) # format string formatting with .format method
print(f"usage: {sys.argv[0]} <a-zA-Z string>") # format string formatting ( python >=3.6 )
```

Also you could have made a `usage` function to avoid repetitive coding.

To verify that a character or string is alpha don't forget about the `.isalpha()` string method!

Usage of `map` is a good idea but you could have used comprehension to do the same, it is more pythonic and I find it personally more readable.

Mariomm-marti:
@clafoutis42 
I used at first the `.isalpha()`, method, but it would also pass some characters that are out of range like `ñ`, or `á` which are unhandled.
Thanks for explaining different ways of string formatting, I knew only about formatting  with `f""`, and I'll definitively use comprehension next time

clafoutis42:
In that case you could have done something like `"".join(c for c in str if c in morse.keys()) == str` to verify that all the string is in the dictionary scope ;)

PR #78, opened 2020-04-06
Chall02-migferna

migferna42:
First let's check the errors of the program, that it has only 1 argument and that it exists. If it meets the first filter, it enters the encoding function and applies the filters that each character of the string is alphabetical, if it is not, it is checked if it is a space. Spaces are accepted as word breakers. If you pass all the filters, the character is mapped with the morse table.

### DIFF
--- a/chall02/migferna.py
+++ b/chall02/migferna.py
@@ -1,0 +1,45 @@
+import sys
+
+MORSE_CODE_DICT = { 'A':'.-', 'B':'-...', 
+                    'C':'-.-.', 'D':'-..', 'E':'.', 
+                    'F':'..-.', 'G':'--.', 'H':'....', 
+                    'I':'..', 'J':'.---', 'K':'-.-', 
+                    'L':'.-..', 'M':'--', 'N':'-.', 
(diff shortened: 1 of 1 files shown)


FranciscoJesusGata:
I didn't thought of using a dictionary, nice idea!

clafoutis42:
The logic of your solution is correct. However the solution in itself is invalid:
- Your file is not executable
- Your file is missing the sheebang indicating which python binary to use
- Your code is in a deprecated version of python (python2.7.X and lower are deprecated since January the 1st of 2020, see this [link](https://pythonclock.org/))
- You are also printing the usage with `./xlogin.py`, although your file is named `migferna.py` which makes your usage incorrect

Other problem:
You are coding in python, however you express the logic as if you were in C, I advise you to read the book [Fluent Python](https://www.amazon.com/Fluent-Python-Concise-Effective-Programming/dp/1491946008) to understand how to be more pythonic, and thus have a really optimized python code writing.

clafoutis42:
After reflection we won't accept PRs that are in the wrong language (here `python2.7.x`). Wanted language is `python3.x` as specified in the README).  
I have to revert this PR

PR #79, opened 2020-04-06
chall02 - gbudau

gbudau:
***How i solved the problem.***
I created a dictionary having the ASCII letters as keys and Morse codes as value, converted the string to lowercase and created a new string using the values of the dictionary.

***Description of the code.***
First check for errors to be sure the command line arguments are valid (one argument, not empty, containing  only lowercase, uppercase and space characters) or exit if not.

Then convert the argument to lowercase and generate a new string using the dictionary we created.

Finally print the newly generated string and exit.

### DIFF
--- a/chall02/gbudau.py
+++ b/chall02/gbudau.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python3
+# String to Morse: Display <a-zA-Z string> as morse code
+import sys
+
+morse_dictionary =    {'a' : ".-", 'b' : "-...", 'c' : "-.-.", 'd' : "-..", \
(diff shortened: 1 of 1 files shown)


clafoutis42:
Your solution is correct, however it is not pythonic.

You are coding in python, however you express the logic as if you were in C, I advise you to read the book [Fluent Python](https://www.amazon.com/Fluent-Python-Concise-Effective-Programming/dp/1491946008) to understand how to be more pythonic, and thus have a really optimized python code writing.

You could use `.isalpha()` to check if the string is alpha, you could use comprehension instead of looping... You're halfway there when you use `.lower()`, or the format strings!

gbudau:
Thank you for the feedback and book recomendation, i will check it out.

PR #81, opened 2020-04-06
Chall 02 solution by rarias-p

TH0RlN:
First of all I state a dictionary with all the letters and their transformations to morse. Then I check that I only get one argument and that it is not empty. The next part is a loop that checks if the string given is valid while transforms the letters to morse and adds them to the string which will be printed at the end.


aguimbao:
Your file wasn't executable (644 permissions). 
As stated in the README: `You have to submit a python3 executable script to encode a string to morse.`, it should be executable as `./your_script.py` . 
To achieve this, you would also need a shebang in the first line of the file like -> `#!/usr/bin/env python`. And you should have specified the python version in the shebang like -> `#!/usr/bin/env python3`

As for the code:
 

```python
if __name__ == "__main__":
    main()
```
- `sys.exit()` > `os.exit()/exit()/quit()/exit`
- There's some redundant "usage" code, you could have implemented a function for this.
- You store every character two times in the dictionary, one in lowercase and one in uppercase, you could have simply used the method `.lower()` on the string argument to convert the whole string to the same case and then only have to store the alphabet once, in lowercase.

Anyway, because of no execution rights your script didn't pass the tests. Best of luck for the next challenge :)

clafoutis42:
I have to revert this PR.

PR #82, opened 2020-04-06
darodrig.py

d-r-e:
First, errors are checked, so that there is just one parameter and it just contains alpha characters and spaces.
Then characters are printed by finding them on the dictionary. Then a new line is printed.

### DIFF
--- a/chall02/darodrig.py
+++ b/chall02/darodrig.py
@@ -1,0 +1,30 @@
+#!/usr/bin/python
+import sys
+
+def main():
+	chars = { 'A':'.-', 'B':'-...', 
+                    'C':'-.-.', 'D':'-..', 'E':'.', 
+                    'F':'..-.', 'G':'--.', 'H':'....', 
+                    'I':'..', 'J':'.---', 'K':'-.-', 
+                    'L':'.-..', 'M':'--', 'N':'-.', 
+                    'O':'---', 'P':'.--.', 'Q':'--.-', 
+                    'R':'.-.', 'S':'...', 'T':'-', 
+                    'U':'..-', 'V':'...-', 'W':'.--', 
+                    'X':'-..-', 'Y':'-.--', 'Z':'--..'}
+	if len(sys.argv) != 2 or sys.argv[1] == "":
+		print("usage: ./xlogin.py <a-zA-Z string>")
+		return
+	for each in sys.argv[1]:
+		if each.isalpha() == 0 and each != ' ':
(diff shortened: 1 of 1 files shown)


chall02/darodrig.py
```diff
@@ -0,0 +1,30 @@
...
+                    'C':'-.-.', 'D':'-..', 'E':'.', 
+                    'F':'..-.', 'G':'--.', 'H':'....', 
+                    'I':'..', 'J':'.---', 'K':'-.-', 
+                    'L':'.-..', 'M':'--', 'N':'-.', 
+                    'O':'---', 'P':'.--.', 'Q':'--.-', 
+                    'R':'.-.', 'S':'...', 'T':'-', 
+                    'U':'..-', 'V':'...-', 'W':'.--', 
+                    'X':'-..-', 'Y':'-.--', 'Z':'--..'}
+	if len(sys.argv) != 2 or sys.argv[1] == "":
+		print("usage: ./xlogin.py <a-zA-Z string>")
```
clafoutis42, line 15: Your file is named `darodrig.py` and could be named differently too, therefore your usage is incorrect.  
You could have made it dynamic this way:
```suggestion
		print("usage: %s <a-zA-Z string>" % sys.argv[0])
```

chall02/darodrig.py
```diff
@@ -0,0 +1,30 @@
...
+		return
+	for each in sys.argv[1]:
+		if each.isalpha() == 0 and each != ' ':
+			print("usage: ./xlogin.py <a-zA-Z string>")
+			return
```
clafoutis42, line 20: Could have been written
```suggestion
        if not sys.argv[1].replace(" ", "").isalpha():
            return print("usage: %s <a-zA-Z string>" % sys.argv[0])
```

chall02/darodrig.py
```diff
@@ -0,0 +1,30 @@
...
+	for each in sys.argv[1]:
+		if each == ' ':
+			sys.stdout.write(each)
+		else:
+			sys.stdout.write(chars[each.upper()])
+	sys.stdout.write('\n')
+	return
```
clafoutis42, line 27: Could have been written
```suggestion
        print("".join(chars.get(c, " ") for c in sys.argv[1].upper()))
```

clafoutis42 (commented):

clafoutis42:
I have to revert this PR.

d-r-e:
At least I learned something new...
I hope this is taken into account at least, for my coalition, and not completely discarded :(
Thank you.

PR #83, opened 2020-04-06
Python morse converter

Dapalm:
- First we make a dictionary with all the morse code (upper case letter, morse code). 
- Then we read the string to be converted from the command line, using argv, (import sys) and convert it to uppercase (in case our original string had any lowercase letter).
- We create an empty string
- For every letter (A-Z, spaces and other symbols are not included) in the original string, we search it's morse value in our dictionary, and append it to our empty string.
- Print the morse string

### DIFF
--- a/chall02/dapalmer.py
+++ b/chall02/dapalmer.py
@@ -3,42 +3,46 @@
 import sys
 
 def main():
-    morseCode = {
-        'A': '.-',
-        'B': '-...',
-        'C': '-.-.',
-        'D': '-..',
-        'E': '.',
-        'F': '..-.',   
-        'G': '--.',
-        'H': '....',
-        'I': '..',   
-        'J': '.---',
-        'K': '-.-',
-        'L': '.-..',   
-        'M': '--',
-        'N': '-.',
-        'O': '---',   
-        'P': '.--.',
-        'Q': '--.-',
-        'R': '.-.',
-        'S': '...',
-        'T': '-',
-        'U': '..-',
-        'V': '...-',
-        'W': '.--',
-        'X': '-..-',
-        'Y': '-.--',
-        'Z': '--..',       
-    }
-    origString = sys.argv[1].upper()
-    morseString = ''
+    if len(sys.argv) == 2:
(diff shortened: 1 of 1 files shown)


chall02/dapalmer.py
```diff
@@ -0,0 +1,49 @@
...
+        origString = sys.argv[1].upper()
+        morseString = ''
+
+        for char in origString:
+            if char >= 'A' and char <= 'Z' or char == ' ':
+                morseString += morseCode[char]
+        print morseString
+    
+    else:
+        print 'Usage:', sys.argv[0], '<a-zA-Z>'
```
clafoutis42, line 45: [Python2 is deprecated](https://pythonclock.org/), you should use python3 as of January the 1st of 2020.

Also the usage you print does not correspond to the one expected in the subject.

```suggestion
        print('usage:', sys.argv[0], '<a-zA-Z string>')
```

chall02/dapalmer.py
```diff
@@ -0,0 +1,49 @@
...
+            'Z': '--..',
+            ' ': ' '       
+        }
+        origString = sys.argv[1].upper()
+        morseString = ''
+
+        for char in origString:
+            if char >= 'A' and char <= 'Z' or char == ' ':
+                morseString += morseCode[char]
+        print morseString
```
clafoutis42, line 42: Your code won't work if there are unexpected characters in the string.

Also you should use comprehension to make your code more pythonic and clearer.

```suggestion
        if not origString.replace(" ", "").isalpha():
            print('usage:', sys.argv[0], '<a-zA-Z string>')
        print("".join(morseCode[c] for c in origString))
```

clafoutis42 (commented):
You are on the right path, however your solution does not work and is not very pythonic. You are writing your code as if you were using C, I suggest you read [Fluent Python](https://www.amazon.com/Fluent-Python-Concise-Effective-Programming/dp/1491946008) if you want to keep progressing in python!

Also, your code script isn't executable.

clafoutis42:
I have to revert this PR.

PR #84, opened 2020-04-06
fgata-va's solution

FranciscoJesusGata:
# How do I solved it
I just iterated over the first argument and making a nested conditionals for each character of the string. I controlled the number of arguments, the length of them, if the first argument have multiple sentences and if all character are alphabetics.

# How my code works
First, I define a function to exit the program everytime something is wrong. Then, the program controls the number of arguments and the lenght of the first argument (the one that matters) and if the chars are alphabetic.


chall02/fgata-va.py
```diff
@@ -0,0 +1,72 @@
...
+    elif (c == 'w'):
+        morse += '.--'
+    elif (c == 'x'):
+        morse += '-..-'
+    elif (c == 'y'):
+        morse += '-.--'
+    elif (c == 'z'):
+        morse += '--..'
+    elif (c == ' '):
+        morse += ' '
```
clafoutis42, line 71: Remember that you are using python, you could have used a dictionary to store the correspondances of each character.  
[Here](https://www.w3schools.com/python/python_dictionaries.asp) is a little link to learn more the basic about python dicts.

chall02/fgata-va.py
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/python3
+import sys
+
+def errorCode():
+    print("usage: ./xlogin.py <a-zA-Z string>")
```
clafoutis42, line 5: Your file won't always be named `xlogin.py`, therefore this usage is incorrect.  
You could have made it dynamic like this:

```suggestion
    print("usage: %s <a-zA-Z string>" % sys.argv[0])
```

clafoutis42 (commented):
You are on the right path, however your solution does not work and is not pythonic. You are writing your code as if you were using C, I suggest you read [Fluent Python](https://www.amazon.com/Fluent-Python-Concise-Effective-Programming/dp/1491946008) if you want to keep progressing in python!

PR #85, opened 2020-04-06
thats it :)

JavierCervilla:
well, in this case i did a dictionary whith the key leter and the content .-..-.-- hehe.
i search each lettter in the dict and concatenate in a string. the program only work whith an only argument. i hope you like my solution.


JavierCervilla:
well, in this case i did a dictionary whith the key leter and the content .-..-.-- hehe.
i search each lettter in the dict and concatenate in a string. the program only work whith an only argument. i hope you like my solution.

JavierCervilla:
well, first the error checking, cannot be more than 2 arg in this case i did a dictionary whith the key leter and the content .-..-.-- hehe.
i search each lettter in the dict and concatenate in a string. the program only work whith an only argument. i hope you like my solution.pd i learned to do a fork :)

JavierCervilla:
well, first the error checking, cannot be more than 2 arg in this case i did a dictionary whith the key leter and the content .-..-.-- hehe.
i search each lettter in the dict and concatenate in a string. the program only work whith an only argument. i hope you like my solution.pd i learned to do a fork :)

JavierCervilla:
well, first the error checking, cannot be more than 2 arg in this case i did a dictionary whith the key leter and the content .-..-.-- hehe.
i search each lettter in the dict and concatenate in a string. the program only work whith an only argument. i hope you like my solution.pd i learned to do a fork :)

chall02/jcervill.py
```diff
@@ -0,0 +1,82 @@
...
+def error():
+	print("usage: " + sys.argv[0] + " <a-zA-Z string>")
+	exit(1)
+
+def is_morse(dictionary, strin):
+	for j in strin:
+		if not (j.isspace() or j.isalpha()):
+			error()
+		if not(dictionary[j]):
+			error()
```
clafoutis42, line 69: This part of the code is never reached, as your previous condition is englobes the content of the dictionary.
```suggestion
```
Also, this is not the correct way of doing this check, for various reason:
- `dictionnary[j]` will raise a `KeyError` exception when the value is absent
- `not dictionnary[j]` will be true if `dictionnary[j]` equals `""`, `0`, `{}`, `[]` and many other values

The correct form would have been:
```suggestion
		if dictionnary.get(j) is None:
			error()
```
Assuming `None` is never to be part of the dictionary's values.

chall02/jcervill.py
```diff
@@ -0,0 +1,82 @@
...
+	return 1
```
clafoutis42, line 70: A more pythonic way of doing this would have been a oneliner:
```suggestion
def is_morse(strin):
    return strin.replace(" ").isalpha()
```

If you want to keep your function's form and make the dictionnary useful:
```suggestion
def is_morse(dictionary, strin):
    return "".join(c for c in strin if c in dictionary.keys()) == strin
```

chall02/jcervill.py
```diff
@@ -0,0 +1,82 @@
...
+
+if (len(sys.argv) == 2 and len(sys.argv[1]) > 0):
+	if (is_morse(morse, sys.argv[1]) == 1):
```
clafoutis42, line 73: In python you have booleans, it would have been more correct to make it like this
```suggestion
	if is_morse(morse, sys.argv[1]) is True:
```
Or even
```suggestion
	if is_morse(morse, sys.argv[1]):
```

Note that I removed the parentheses from your condition, in python they are redundant, it is more readable to remove them.

chall02/jcervill.py
```diff
@@ -0,0 +1,82 @@
...
+		str = ""
+		while (i < len(sys.argv[1])):
+			str = str + (morse[sys.argv[1][i]])
+			i += 1
+		print (str)
```
clafoutis42, line 78: It would have been more pythonic to write this this way:
```suggestion
		print("".join(morse[c] for c in sys.argv[1]))
```

chall02/jcervill.py
```diff
@@ -0,0 +1,82 @@
+import sys
```
clafoutis42, line 1: You forgot the sheebang, even if the file is executable, the system won't know how to execute the script!
```suggestion
#!/usr/bin/env python3
import sys
```

clafoutis42 (commented):
You have a working solution. However your code is missing the sheebang and therefore will not be executed in python.  
Also you could have made a code more pythonic, read [Fluent Python](https://www.amazon.com/Fluent-Python-Concise-Effective-Programming/dp/1491946008) to keep progressing in python!

JavierCervilla:
Thanks Charles,
I am not used yo the computar being so clever...

clafoutis42:
I have to revert this PR.

JavierCervilla:
Okey....
I wont Answer you anymore 🙄😒

PR #86, opened 2020-04-06
chall02 - gsanchez

GonzSanch:
First at alll we handle the errors related to arguments provided if there are more than one or the string is empty we exited the program and print the usage message. Then using the map in python that is a key value list, the program iterate every char turns to lower from the word to translate in the key/value in morse code, adding the space for multi-word process, first step validate the word and then repeating the process printing the morse code with a new line at the end.


chall02/gsanchez.py
```diff
@@ -0,0 +1,56 @@
+# **************************************************************************** #
```
clafoutis42, line 1: Your code is missing the sheebang, therefore it won't be executed in python and will fail.
```suggestion
#!/usr/bin/env python3
# **************************************************************************** #
```

chall02/gsanchez.py
```diff
@@ -0,0 +1,56 @@
...
+	print("usage " + sys.argv[0] + " <a-zA-Z string>")
+	exit()
+for char in sys.argv[1]:
+	if char.lower() not in MORSE_CODE:
+		print("usage " + sys.argv[0] + " <a-zA-Z string>")
+		exit()
+for char in sys.argv[1]:
+	if char.lower() in MORSE_CODE:
+		sys.stdout.write(MORSE_CODE[char.lower()])
+sys.stdout.write('\n')
```
clafoutis42, line 56: Your code is not protected and will be executed even if someone just tries to import part of your code.  
To protect your code and make sure the executable part is only run when we execute the file, you can do this:
```suggestion
if __name__ == "__main__":
    if len(sys.argv) != 2:
    	print("usage " + sys.argv[0] + " <a-zA-Z string>")
    	exit()
    if sys.argv[1] == "":
    	print("usage " + sys.argv[0] + " <a-zA-Z string>")
    	exit()
    for char in sys.argv[1]:
    	if char.lower() not in MORSE_CODE:
    		print("usage " + sys.argv[0] + " <a-zA-Z string>")
    		exit()
    for char in sys.argv[1]:
    	if char.lower() in MORSE_CODE:
    		sys.stdout.write(MORSE_CODE[char.lower()])
    sys.stdout.write('\n')
```
`__name__` is the variable containing the name of the current module, the main module will always be called `__main__`.

chall02/gsanchez.py
```diff
@@ -0,0 +1,56 @@
...
+        "y": "-.--",
+        "z": "--..",
+		" ": " ",}
+
+if len(sys.argv) != 2:
+	print("usage " + sys.argv[0] + " <a-zA-Z string>")
+	exit()
+if sys.argv[1] == "":
+	print("usage " + sys.argv[0] + " <a-zA-Z string>")
+	exit()
```
clafoutis42, line 48: This is a little redundant don't you think?

Maybe try writing grouped condition, as python's conditions as lazy, the second part of a `or` group won't be evaluated if the first part is true:
```suggestion
if len(sys.argv) != 2 or sys.argv[1] == "":
	print("usage " + sys.argv[0] + " <a-zA-Z string>")
	exit()
```

chall02/gsanchez.py
```diff
@@ -0,0 +1,56 @@
...
+for char in sys.argv[1]:
+	if char.lower() not in MORSE_CODE:
+		print("usage " + sys.argv[0] + " <a-zA-Z string>")
+		exit()
```
clafoutis42, line 52: You are repeating this part of the code A LOT. Maybe you should make a `usage` function to avoid repetition and ease readability.

chall02/gsanchez.py
```diff
@@ -0,0 +1,56 @@
...
+for char in sys.argv[1]:
+	if char.lower() in MORSE_CODE:
+		sys.stdout.write(MORSE_CODE[char.lower()])
```
clafoutis42, line 55: You already verified if the string was composed by morse_code characters, it is not necessary to do it again here.

chall02/gsanchez.py
clafoutis42, line 52: You could have made this verification in a more pythonic way:
```suggestion
if "".join(c for c in sys.argv[1] if c.lower() in MORSE_CODE.keys()) == sys.argv[1]:
	print("usage " + sys.argv[0] + " <a-zA-Z string>")
	exit()
```

Also if you don't expect you dictionary to be dynamic, a simple:
```suggestion
if sys.argv[1].replace(" ", "").isalpha():
	print("usage " + sys.argv[0] + " <a-zA-Z string>")
	exit()
```
Would have made the trick.

chall02/gsanchez.py
clafoutis42, line 56: You could have made this in a more pythonic way:
```suggestion
print("".join(MORSE_CODE[c] for c in sys.argv[1]))
```

clafoutis42 (commented):
You have an almost working solution, you forgot the `:` in the error message. However your code is missing the shebang and therefore will not be executed in python.
Also you could have made a code more pythonic, read Fluent Python to keep progressing in python!

GonzSanch:
WoW! Thank you Charles! This was very usefull and excellent way to learn new language, 42-p2p style!
Feels lucky!

PR #87, opened 2020-04-06
Added chall02

byrd92:
I made a morse-dictionary in uppercase.

I check if there are only 2 argv && the argv are alpha or ' '

then i change each letter (changed to upper if neccesary) for the dictionary.
else return error usage:


anna-dominguez:
There's a few problem with your file: 

- It is not executable

- It is missing the sheebang indicating which python binary to use

- You are also printing the usage with `./xlogin.py`, although your file is named `egarcia-.py` which makes your usage incorrect

The solution might be working, you could have add the `" "` character in your options directly because you're checking all the characters type in your condition. And add a condition before if you only receive this character, in order to avoid printing a space alone.

clafoutis42:
I have to revert this PR.

PR #88, opened 2020-04-06
morse encoder done

dbelinsk:


chall02/dbelinsk.py
```diff
@@ -0,0 +1,90 @@
...
+	elif c is 'w':
+		return ".--"
+	elif c is 'x':
+		return "-..-"
+	elif c is 'y':
+		return "-.--"
+	elif c is 'z':
+		return "--.."
+	elif c is ' ':
+		return " "
```
clafoutis42, line 58: Remember that you are using python, you could have used a dictionary to store the correspondances of each character.  
[Here](https://www.w3schools.com/python/python_dictionaries.asp) is a little link to learn more the basic about python dicts.

chall02/dbelinsk.py
```diff
@@ -0,0 +1,90 @@
...
+
+def morseLoop(s):
+	code = ''
+	for c in s:
+		code += encode(c)
+	print (code)
```
clafoutis42, line 64: You could have made this more pythonic with a oneliner:
```suggestion
	print("".join(encode(c) for c in s))
```

chall02/dbelinsk.py
```diff
@@ -0,0 +1,90 @@
...
+def is_all_letter(av):
+	i = 0
+	for c in av:
+		i += 1
+		if c.isalpha() == False:
+			if c != ' ':
+				return 0
+	if i == 0:
+		return 0
+	return 1
```
clafoutis42, line 75: You could have made this more pythonic with a oneliner:
```suggestion
	return av.replace(" ", "").isalpha()
```

clafoutis42 (commented):

clafoutis42:
I just figured you did not comment your PR, I have yo revert your code, sorry.

PR #89, opened 2020-04-06
chall02-pcuadrad

pabloocg:
First I set the python environment variable so I can run like a normal script "./run" and I also give him execution permissions.
The program checks that the number of parameters entered is correct, in this case a single argument.
Then I check that the entered string consists only of alpha characters and spaces.
At the end, I create a dictionary **letter:morse**  where I go taking the value corresponding to the characters of the string and concatenate them to return it.

### DIFF
--- a/chall02/pcuadrad.py
+++ b/chall02/pcuadrad.py
@@ -1,0 +1,33 @@
+#!/usr/bin/env python
+
+import sys
+
+def finish_program():
+    print ("usage: ./pcuadrad.py <a-zA-Z string>")
+    sys.exit()
+
+def check_string(txt):
+    for char in txt:
+        if not char.isalpha() and not char.isspace():
+            return 0
+    return 1
+
+def parse_morse(txt):
+    morse = {
+        "A": ".-", "B": "-...", "C": "-.-.", "D": "-..", "E": ".", "F": "..-.",
+        "G": "--.", "H": "....", "I": "..", "J": ".---", "K": "-.-", "L": ".-..",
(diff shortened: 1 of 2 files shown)


chall02/pcuadrad.py
```diff
@@ -0,0 +1,28 @@
...
+        "A": ".-", "B": "-...", "C": "-.-.", "D": "-..", "E": ".", "F": "..-.",
+        "G": "--.", "H": "....", "I": "..", "J": ".---", "K": "-.-", "L": ".-..",
+        "M": "--", "N": "-.", "O": "---", "P": ".--.", "Q": "--.-", "R": ".-.",
+        "S": "...", "T": "-", "U": "..-", "V": "...-", "W": ".--", "X": "-..-",
+        "Y": "-.--", "Z": "--..", " ": " "
+    }
+    str_morse = ""
+    for char in txt:
+        str_morse += morse.get(char)
+    return str_morse
```
clafoutis42, line 22: You could have made this more pythonic with this oneliner:
```suggestion
   return "".join(morse[c] for c in txt)
```

chall02/pcuadrad.py
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env python
+
+import sys
+
+def check_string(txt):
+    for char in txt:
+        if not char.isalpha() and not char.isspace():
+            return 0
+    return 1
```
clafoutis42, line 9: You could have made this more pythonic with this oneliner:
```suggestion
    return txt.replace(" ", "").isalpha()
```
Remember that in python, you have booleans, you should use `True` and `False` instead of `1` and `0`

clafoutis42 (commented):

PR #90, opened 2020-04-06
chall02 pablomar

PalindromoMeszaros:
1 - I import the sys library which will be useful for working with command line arguments

2 - For the letter-morse encryption I use a dictionary, a useful python structure that allows us to pair a data -the morse code- with a key -the letter-

3 - With the "sys.argv" method we work with command line arguments. The program checks the validity of our prompt. We define rules for only allowing calls with 2 arguments (program name + string to translate), and to avoid empty strings as an argument.

4.1 - "morse_conversion" function is in charge of the coding process. Firstly we define a "morse_string" empty string variable which we will use to incorporate the morse code of each letter.
4.2 - The function iterates in a loop each character of a string, if the character is between 'A' and 'Z', it goes to the correspondent value of the dictionary an appends it to the "morse_string" till the string is finished. If the character is a space, it appends a space and if it's not any of those, it writes an error message and exits.

5 - For simplicity, instead of creating a dictionary with both uppercase and lowercase letters, we assign the method .upper to our sys.argv call and work always with capital letters as the key for the translator.


anna-dominguez:
I think your solution is working, but you have some issues with the `README` requirements:
- Your file is not executable
- Your file is missing the sheebang indicating which python binary to use

You could use a function to avoid a redundant error managing, and add the space the characters to translate to avoir another condition

PR #91, opened 2020-04-06
chall02-fulldemo

fulldemo13:
First of all I check if the arguments passed are valid. The number of arguments and if the argument is void.
Second, I pass all the character to low characters and I start to check char by char if it's a letter or a space.
I check with the dictionary "morse" where every letter has a morse string.
Ultimately, if all are letters or spaces, I return the new string and end, else I return the error.

### DIFF
--- a/chall02/fulldemo.py
+++ b/chall02/fulldemo.py
@@ -1,0 +1,27 @@
+#! /usr/local/bin/python3
+
+import sys
+num = len(sys.argv)
+if num != 2: 
+	print("usage: ./fulldemo.py <a-zA-Z string>")
+	exit(1)
+else:
+	morse = {'a' : ".-", 'b' : "-...", 'c' : "-.-.", 'd' : "-..", 'e' : ".", 'f' : "..-.", 'g' : "--.", 'h' : "....", 'i' : "..", 'j' : ".---", 'k' : "-.-", 'l' : ".-..", 'm' : "--", 'n' : "-.", 'o' : "---", 'p' : ".--.", 'q' : "--.-", 'r' : ".-.", 's' : "...", 't' : "-", 'u' : "..-", 'v' : "...-", 'w' : ".--", 'x' : "-..-", 'y' : "-.--", 'z' : "--.."}
+	res = ''
+	len = len(sys.argv[1])
+	if len == 0:
+		print("usage: ./fulldemo.py <a-zA-Z string>")
+		exit(1)
+	sys.argv[1] = sys.argv[1].lower()
+	count = 0
+	while (count < len):
+		if not sys.argv[1][count].isalpha():
+			if not sys.argv[1][count].isspace():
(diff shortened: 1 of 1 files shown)


chall02/fulldemo.py
```diff
@@ -0,0 +1,27 @@
+#! /usr/local/bin/python3
+
+import sys
+num = len(sys.argv)
+if num != 2: 
+	print("usage: ./fulldemo.py <a-zA-Z string>")
```
clafoutis42, line 6: You should make your error message more dynamic, what if we come to change the file's name?
```suggestion
	print("usage: %s <a-zA-Z string>" % sys.argv[0])
```

chall02/fulldemo.py
```diff
@@ -0,0 +1,27 @@
...
+	exit(1)
+else:
+	morse = {'a' : ".-", 'b' : "-...", 'c' : "-.-.", 'd' : "-..", 'e' : ".", 'f' : "..-.", 'g' : "--.", 'h' : "....", 'i' : "..", 'j' : ".---", 'k' : "-.-", 'l' : ".-..", 'm' : "--", 'n' : "-.", 'o' : "---", 'p' : ".--.", 'q' : "--.-", 'r' : ".-.", 's' : "...", 't' : "-", 'u' : "..-", 'v' : "...-", 'w' : ".--", 'x' : "-..-", 'y' : "-.--", 'z' : "--.."}
+	res = ''
+	len = len(sys.argv[1])
+	if len == 0:
+		print("usage: ./fulldemo.py <a-zA-Z string>")
```
clafoutis42, line 13: You wrote this part several times, maybe you cold have made a function to avoir repetitive lines.

chall02/fulldemo.py
```diff
@@ -0,0 +1,27 @@
...
+	while (count < len):
+		if not sys.argv[1][count].isalpha():
+			if not sys.argv[1][count].isspace():
+				print("usage: ./fulldemo.py <a-zA-Z string>")
+				exit(1)
+		if sys.argv[1][count].isspace():
+			res += sys.argv[1][count]
+		else:
+			res += morse[sys.argv[1][count]]
+		count = count + 1
```
clafoutis42, line 26: You could have make a more pythonic approach, with a oneliner to verify the integrity of the string:
`str.replace(" ", "").isalpha()` or a one liner to build your morse string from the dictionary `"".join(morse[c] for c in str)`.

This would have saved you many lines, and some time. It is also more optimized this way.

chall02/fulldemo.py
```diff
@@ -0,0 +1,27 @@
+#! /usr/local/bin/python3
+
+import sys
```
clafoutis42, line 3: You should always protect the executable part of your code with
```python3
if __name__ == "__main__":
    # rest of your code
```
Or else this code will be executed even when importing things from your module.

clafoutis42 (commented):

PR #93, opened 2020-04-06
szavarce.py

szavarce42:
I have created a dictionary with uppercase letters index, which represent each morse code value. Then checked if there are two arguments to continue, otherwise will get out prompting an error message. Then the program will loop through the argument received (only the arg[1] since arg[0] is the name of the script). If the uppercase char is found in our dictionary, it will add it to a variable to display afterwards. If we find something different from the letters in the dictionary it will get out printing the error message. If we made it to the end, then we print the result stored in "resultado".


chall02/szavarce.py
```diff
@@ -0,0 +1,17 @@
...
+  print("usage: ./szavarce.py <a-zA-Z string>")
+  exit()
+
+for c in sys.argv[1]:
+  if c.upper() in codigo_morse:
+    resultado += codigo_morse[c.upper()]
+  else:
+    print("usage: ./szavarce.py <a-zA-Z string>")
+    exit()
+print("{}".format(resultado))
```
clafoutis42, line 17: Why not just `print(resultado)`? :p

chall02/szavarce.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python
+import sys
+codigo_morse = {
+  "A": ".-", "B": "-...", "C": "-.-.", "D": "-..", "E": ".", "F": "..-.", "G": "--.", "H": "....", "I": "..", "J": ".---", "K": "-.-", "L": ".-..", "M": "--", "N": "-.", "O": "---", "P": ".--.", "Q": "--.-", "R": ".-.", "S": "...", "T": "-", "U": "..-", "V": "...-", "W": ".--", "X": "-..-", "Y": "-.--", "Z": "--..", " ": " "
+}
+resultado = ""
+if len(sys.argv) != 2:
+  print("usage: ./szavarce.py <a-zA-Z string>")
```
clafoutis42, line 8: The filename might change, you should make the message dynamic:
```suggestion
  print("usage: %s <a-zA-Z string>" % sys.argv[0])
```

chall02/szavarce.py
```diff
@@ -0,0 +1,17 @@
...
+  exit()
+
+for c in sys.argv[1]:
+  if c.upper() in codigo_morse:
+    resultado += codigo_morse[c.upper()]
+  else:
+    print("usage: ./szavarce.py <a-zA-Z string>")
+    exit()
```
clafoutis42, line 16: You could have made this more pythonic with those oneliners:
```suggestion
if not sys.argv[1].replace(" ", "").isalpha():
    print("usage: ./szavarce.py <a-zA-Z string>")
    exit()
resultado = "".jon(codigo_morse[c] for c in sys.argv[1])
```

chall02/szavarce.py
```diff
@@ -0,0 +1,17 @@
...
+resultado = ""
+if len(sys.argv) != 2:
+  print("usage: ./szavarce.py <a-zA-Z string>")
+  exit()
+
+for c in sys.argv[1]:
+  if c.upper() in codigo_morse:
+    resultado += codigo_morse[c.upper()]
+  else:
+    print("usage: ./szavarce.py <a-zA-Z string>")
```
clafoutis42, line 15: You wrote this multiple times, maybe you should use a function to improve readability.

clafoutis42 (commented):

However as your file is not executable, I can't accept this PR :(

szavarce42:
Hello, thanks for your comments. What do u mean it is not executable? I
could execute it in my pc... and you asked for a <login>.py file...

PR #94, opened 2020-04-06
Solution Chall02 by fvega-tr

fvega-tr:


anna-dominguez:
You didn't read the `README` correctly and forgot to make your file an executable, then I cannot test it. I didn't see a shebang too at the principle of the file. And you forgot that your file is named `fvega-tr` that is the equivalence of `xlogin` in the example, but you need to change it. Too bad you didn't wrote a comment to explain your code 😞

PR #95, opened 2020-04-06
Patch 1

mcarrete42:
This was my first python script! it has been really challenging, but more about research than anything else. It has been really useful to begin to understand Python syntax. 

The code itself is very simple, and divided in 2 parts:
1. Error checking (if more than 2 args, or if the arg[1] is empty) then print the error message. 
2. Convert the string to upper, and execute the morse translator function. This function will iterate through the uppercased string and see if each of the chars are included in my morse dictionary (A-Z plus space).  if the char is included, it will include the translation in "returnString".  if all of the chars are included in the dictionary, then print the "returnString".
If any of the chars are not included in the dictionary, print error message and return.


chall02/mcarrete.py
```diff
@@ -0,0 +1,32 @@
...
+def stringToMorse(text):
+	morse =		{'A' : '.-', 'B' : '-...', 'C' : '-.-.', 'D': '-..', 'E' : '.',
+				'F' : '..-.', 'G' : '--.', 'H' : '....', 'I' : '..', 'J' : '.---',
+				'K' : '-.-', 'L' : '.-..', 'M' : '--', 'N' :  '-.', 'O' : '---',
+				'P' : '.--.', 'Q' : '--.-', 'R' : '.-.',  'S' : '...', 'T' : '-',
+				'U' : '..-', 'V' : '...-', 'W' : '.--', 'X' : '-..-', 'Y' : '-.--',
+				'Z' : '--..', ' ' : ' '}
+	retMorse = ""
+	count = 0
+	error = "usage: ./mcarrete.py <a-zA-Z string>"
```
clafoutis42, line 14: The file name might change, you should make a more dynamic message:
```suggestion
	error = "usage: %s <a-zA-Z string>" %s sys.argv[0]
```

chall02/mcarrete.py
```diff
@@ -0,0 +1,32 @@
...
+	for a in text:
+		if a in morse:
+			retMorse += morse[a]
+			count += 1
```
clafoutis42, line 18: Where do you use this variable?

chall02/mcarrete.py
```diff
@@ -0,0 +1,32 @@
...
+		else:
+			print(error)
+			return error
+	print(retMorse)
```
clafoutis42, line 22: You could have simplified this code in a more pythonic way:
```suggestion
    if "".join(c for c in text if c in morse.keys()) != txt:
        return print(error)
    print("".join(morse[c] for c in text))
```

clafoutis42 (commented):

PR #96, opened 2020-04-06
morse

jdiaz-co:
First in our main we check if the argument is valid, that is, if there is only 1 argument and if this argument is valid, if it is not, we send an error message.

If it is valid we go to the morse function, sending it the word that they have passed us through the argument in lowercase (this is achieved with the lower () function)

In the morse function we create an array with all the letters of the alphabet, giving them the value individually with the morse code.

We look if there are coincidences between the word that we send with that array, if there is not, we send an error message. If there is a coincidence we write the morse code.


anna-dominguez:
Too bad your file is not executable, all the rest seems to work but I cannot test it.
All the good points : 

- Shebang ok 

- Think of space in the list of morse characters 

- Good message of error handling, but could do a function to avoid redundancy 

👍

PR #99, opened 2020-04-06
Challenge 02: Constelación Serpens / String to Morse

eligero:

Solved with key-value pair dictionary. The achieved solution has the mindset
of continuous time problem, i.e manage and output each letter at time, without
memory allocation. In real morse each letter is transmited over radio
connection and it can be corrupted by radio noise. I understood that every
new command line argument means a new word, so I put a space between words.


eligero:

Solved with key-value pair dictionary. The achieved solution has fits
the challenge problem's conceivement, by blacklisting wrongs behaviors.

PR #100, opened 2020-04-06
vde-dios.c

vdedios:
**Solution**

1. Check if script arguments are valid: only one argument with len > 0 will be valid.
2. With a loop we check, character by character if it is a valid one (a-z or simple space):
- If we encounter any mismatch, we exit with and error message.
- If character is valid we look for its equivalence on dictionary list using its ASCII value to get the list position. We append the character in the final string we will be returning for finally printing


anna-dominguez:
It's is an interesting way to resolve it using ascii characters but very similar to C programming. You named you variable dictionary, I think you need to see what a dictionary is in Python, it would help you a lot to understand how you could have done it differently. 
Otherwise, it works perfectly for the translation. I can see one issue with your error handling message, your output is `././vde-dios.py [...]` and not `./vde-diod.py [...]` but I think you will see where this comes from very quickly 😉 

Don't forget your shebang last time ! It's important

PR #101, opened 2020-04-06

eligero:

Solved with key-value pair dictionary. The achieved solution has fits
the challenge problem's conceivement, by blacklisting wrongs behaviors.

### DIFF
--- a/chall02/eligero.py
+++ b/chall02/eligero.py
@@ -1,0 +1,53 @@
+#!/usr/bin/python3
+#* ************************************************************************** */
+#*                                                                            */
+#*                                                        :::      ::::::::   */
+#*   eligero.py                                         :+:      :+:    :+:   */
+#*                                                    +:+ +:+         +:+     */
+#*   By: eligero <eligero@student.42.fr>            +#+  +:+       +#+        */
+#*                                                +#+#+#+#+#+   +#+           */
+#*   Created: 2020/04/02 15:01:31 by eligero           #+#    #+#             */
+#*   Updated: 2020/04/02 17:21:49 by eligero          ###   ########.fr       */
+#*                                                                            */
+#* ************************************************************************** */
+
+#
+# Constelación Serpens (Chall02) / String to Morse
+#
+# Solved with key-value pair dictionary. The achieved solution has the mindset
+# of continuous time problem, i.e manage and output each letter at time, without
+# memory allocation. In real morse each letter is transmited over radio
+# connection and it can be corrupted by radio noise. I understood that every
+# new command line argument means a new word, so I put a space between words.
+#
+
+import sys
+
+MORSE_CODE_DICT = { 'A':'.-', 'B':'-...', 'C':'-.-.', 'D':'-..', 'E':'.',
+					'F':'..-.', 'G':'--.', 'H':'....', 'I':'..', 'J':'.---',
+					'K':'-.-', 'L':'.-..', 'M':'--', 'N':'-.', 'O':'---',
+					'P':'.--.', 'Q':'--.-', 'R':'.-.', 'S':'...', 'T':'-',
+					'U':'..-', 'V':'...-', 'W':'.--', 'X':'-..-', 'Y':'-.--',
+					'Z':'--..'}
+
+def encrypt(message):
+	for char in message:
+		if (char >= 'a' and char <= 'z') or (char >= 'A' and char <= 'Z'):
+			print(MORSE_CODE_DICT[char.upper()], end = '')
+		elif char == ' ':
+			print(' ', end = '')
+		else:
+			print('#ERROR#', end = '')
+
+def main():
+	arguments = len(sys.argv) - 1
+	i = 1
+	while (arguments > i):
+		encrypt(sys.argv[i])
+		print(' ', end = '')
+		i += 1
+	encrypt(sys.argv[i])
+	print('')
+
+if __name__ == '__main__':
+    main()
--- a/chall02/eligero.py
+++ b/chall02/eligero.py
@@ -1,4 +1,5 @@
 #!/usr/bin/python3
+
 #* ************************************************************************** */
 #*                                                                            */
 #*                                                        :::      ::::::::   */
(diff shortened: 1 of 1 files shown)


eligero:
Improved solution with classes and static variables

eligero:
Improved solution of Constelación Serpens (Chall02) / String to Morse

Solved with key-value pair dictionary as static variable in a class.
The achieved solution has fitted challenge's problem conception, by
blacklisting wrong inputs by using regular expressions and, or, wrong
invocations. Memory clean.

eligero:
There is no need of classes. Speed up with local variables, as they are faster than global variables on python.

chall02/eligero.py
```diff
@@ -0,0 +1,52 @@
...
+		'x':'-..-', 'y':'-.--', 'z':'--..'}
+	cipher = ''
+	for char in message:
+		if char >= 'a' and char <= 'z':
+			cipher += morse_dict[char]
+		elif  char >= 'A' and char <= 'Z':
+			cipher += morse_dict[char.lower()]
+		elif char == ' ':
+			cipher += ' '
+	return (cipher)
```
clafoutis42, line 42: You could have made this more pythonic with this oneliner
```suggestion
	return "".join(morse_dict.get(c, " ") for c in message)
```

chall02/eligero.py
```diff
@@ -0,0 +1,52 @@
...
+
+def main():
+	if len(sys.argv) != 2 or sys.argv[1] == '' or \
+		re.search(r'[^a-zA-Z ]', sys.argv[1]):
```
clafoutis42, line 46: Using regex is a little overkill, you could have simply used the `.replace()` and `.isalpha` string methods in order to sanitize the string:
```suggestion
	if len(sys.argv) != 2 or sys.argv[1].replace(" ", "").isalpha():
```
Note that it also covers the `sys.argv[1] == ""` condition as `"".isalpha()` returns `False`

chall02/eligero.py
```diff
@@ -0,0 +1,52 @@
...
+		print ('usage: ' + sys.argv[0] + ' ' + '<a-zA-Z string>')
+	else:
+		print (encrypt(sys.argv[1]))
```
clafoutis42, line 49: You could have made the code shorter by doing this
```suggestion
		return print ('usage: ' + sys.argv[0] + ' ' + '<a-zA-Z string>')
	print (encrypt(sys.argv[1]))
```

clafoutis42 (commented):

chall02/eligero.py
eligero: That's cool. I know, but I priorized readable code. But I take note of this for future challenges.

chall02/eligero.py
eligero: That's cool to know, but memory accesses & replacements are overkill in scenarios like microcontrollers were regex performs far better. That said, I take note for future challenges.

chall02/eligero.py
eligero: This is a fault on my side. I take note.
Your comments on my code are welcome. Thank you.

chall02/eligero.py
clafoutis42: I find my code easier to read. That said, it is my opinion, but your multiple conditions aren't easy to understand.  
A more readable solution would have been to put my one_liner in a function that has a clear name about what it does, and call it in the return close.

chall02/eligero.py
clafoutis42: In this case, and many others, using regular expressions will make more memory accesses than a simple replace and isalpha.

chall02/eligero.py
eligero: Those are opinions. Both are respectable.

chall02/eligero.py
eligero: Of course, here is an scenario-dependent situation. That said, read memory accesses are faster and less power consumption than write ones. I suggest you read [Computer Architecture: A Quantitative Approach](https://www.amazon.es/Computer-Architecture-Quantitative-Approach-Kaufmann-ebook/dp/B078MFDTX4/ref=sr_1_1?__mk_es_ES=%C3%85M%C3%85%C5%BD%C3%95%C3%91&crid=1BVRNZM5TEDH2&dchild=1&keywords=computer+architecture+a+quantitative+approach&qid=1586289993&sprefix=computer+arch%2Caps%2C153&sr=8-1) if you want to get any documentation about my comments.

That said, I took note for the incomming challenges: prefer replacements over regexp.

clafoutis42: Don't forget that you're talking to a 3 year experienced python back-end developer :)  
You should really read [this book](https://www.amazon.com/Fluent-Python-Concise-Effective-Programming/dp/1491946008).

eligero: Don't forget that you're talking to a 8+ years experienced embedded systems engineer and FPGA expert :).
Yo should really read [this book](https://www.amazon.es/Computer-Architecture-Quantitative-Approach-Kaufmann-ebook/dp/B078MFDTX4/ref=sr_1_1?__mk_es_ES=%C3%85M%C3%85%C5%BD%C3%95%C3%91&crid=1BVRNZM5TEDH2&dchild=1&keywords=computer+architecture+a+quantitative+approach&qid=1586289993&sprefix=computer+arch%2Caps%2C153&sr=8-1).
The show must go on! :)

clafoutis42: I did a benchmark, I sent you in slack the complete results, but I'll just place this one here:
```
./pistache.py 'big gooood string aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa rrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrr big gooood string aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa rrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrr big gooood string aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa rrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrr'
3.407823218
./eligero.py 'big gooood string aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa rrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrr big gooood string aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa rrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrr big gooood string aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa rrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrr'
98.57283754500001
```
From the benchmark we can tell 2 things:
- My suggestions make your solution faster than your original solution
- Your solution is only faster than mine in finding errors inside small strings (and this you can say thanks to the regex)

eligero: Of course, as I said earlier: it's my fault. Your solution should perform better, as it iterates on the dict keys first not over the all string firstly. It, obviously, should perform faster & better.

That said, Python is a interpreted language, which relies on the interpreter wich, again, relies on harware. It's perfectly possible an interpreter build for a custom hardware where my solution could perform better. Although this is possible, is not real, to my knowledge, by the way standard is x86, ARM and RISC. As I said earlier, on the real world your solution has to perform far better.

Problem here is that I was wirting about cost of read operations on memory versus write operations on memory, in which case there is no posible debate nowadays.

You should read [this book](https://www.amazon.es/dp/0385347758/ref=as_li_tl?ie=UTF8&linkCode=gs2&linkId=BSMOAZWQRFJAVPRJ&creativeASIN=0385347758&tag=why059-21&creative=390957&camp=1789).

You win. By the way, there wasn't a war on my side.

Show is ended. :)

PR #102, opened 2020-04-06
mpernia-

mmiguelx:
Hello again,

1. Use the powerful python tool called dictionary indexing every character with his encryption pair.
2. Check if there's exactly 2 system arguments, the string is alphabetic or have spaces and his length is greater than 0, if not print the error string.
3. Lowercase the string to easily handle the case.
4. Print every character with a end='' to avoid the automatic end of line from python print (if character is a space ignore the dictionary searching and just print space with end='').
5. Print the last end of line.


JavierCervilla:
eres un crack

esepunto:
Really really like your approach. 

Note: maybe the "moulinette" mark your login as an error in the last line (other optiob: sys.argv[0]).

chall02/mpernia-.py
```diff
@@ -0,0 +1,43 @@
...
+	't':'-',
+	'u':'..-',
+	'v':'...-',
+	'w':'.--',
+	'x':'-..-',
+	'y':'-.--',
+	'z':'--..'
+}
+
+if len(sys.argv) == 2 and all(x.isalpha() or x.isspace() for x in sys.argv[1]) and len(sys.argv[1]) > 0:
```
clafoutis42, line 34: Nice! You are using comprehension!
That's really great!
However here there is a more optimized approach:
```suggestion
if len(sys.argv) == 2 and sys.argv[1].replace(" ", "").isalpha():
```
Note that it also covers then `len(sys.argv[1]) > 0` condition as `"".isalpha()` returns `False`

chall02/mpernia-.py
```diff
@@ -0,0 +1,43 @@
...
+	string = str.lower(sys.argv[1])
+	for char in string:
+		if char != ' ':
+			print(morse[char], end='')
+		else:
+			print(' ', end='')
+	print('')
```
clafoutis42, line 41: You could have used the comprehensions here too:
```suggestion
	print("".join(morse[c] for c in sys.argv[1].lower()))
```

chall02/mpernia-.py
```diff
@@ -0,0 +1,43 @@
...
+else:
+	print("usage: ./mpernia-.py <a--Z string>")
```
clafoutis42, line 43: The filename might change, you should do something more dynamic:
```suggestion
	print("usage: %s <a--Z string>" % sys.argv[0])
```

clafoutis42 (commented):
Good job, you have a working solution, however it is not perfectly pythonic (although you have the more pythonic approach I've seen so far). You are writing your code as if you were using C, I suggest you read [Fluent Python](https://www.amazon.com/Fluent-Python-Concise-Effective-Programming/dp/1491946008) if you want to keep progressing in python!

However, your file is not executable, I can't accept the PR :(

chall02/mpernia-.py
mmiguelx: Thank you for the review, that's very helpful. The changes you made blow my mind. Why my file couldn't be executed?

aguimbao: I'll answer for him, no execution rights on the file. It had to be executed as `./file.py`, not as `your_python_interpreter file.py`.

mmiguelx: Uhmm... I understand, but besides chmod there's something else to do? Because, I have this:
`-rwxrwxrwx 1 mmx66 mmx66  663 abr  6 16:19 mpernia-.py`

aguimbao: You only need the shebang (which you had), and the execution permissions. At the time of the last commit (826e6e7), your permissions were -rw-r--r-- (644), so it wasn't executable :)

mmiguelx: Thank you, @abguimba. You were very helpful, :)
I'm trying to understand to send the new one, we'll see if I have luck this time. >:D

PR #103, opened 2020-04-06
Add files via upload

irodrigo-42Mad:
Primer commit de este proyecto proyecto base


clafoutis42:
You did not explain your code like asked in the main README. Also you commented in spanish when it was explicitly asked during last challenge to comment in **english**.

The files were also turned-in the wrong way.

PR #104, opened 2020-04-06
String-to -morse Antomart

miyomovil:
To make this exercise i decided to
1 Create a dictionary with all morse values stored as items and uppercase alphas as keys. (including ' ' in my dictionary)
2 Control the cases of less or more arguments than demanded
3 Control isalpha or "space"
4 Turn to upper
5 for each char or space in the string write the corresponding char of our dictionary

### DIFF
--- a/chall02/antomart.py
+++ b/chall02/antomart.py
@@ -1,0 +1,51 @@
+#!/usr/bin/python3
+
+import sys
+
+# Make a dictionary widh all the morse letters Keys are alphabetic, values are morse and items are all
+
+MORSE = {
+    'A':'.-',
+    'B':'-...',
+    'C':'-.-.',
+    'D':'-..',
+    'E':'.',
+    'F':'..-.',
+    'G':'--.',
+    'H':'....',
+    'I':'..',
+    'J':'.---',
+    'K':'-.-',
+    'L':'.-..',
+    'M':'--',
+    'N':'-.',
+    'O':'---',
+    'P':'.--.',
+    'Q':'--.-',
+    'R':'.-.',
+    'S':'...',
+    'T':'-',
+    'U':'..-',
+    'V':'...-',
+    'W':'.--',
+    'X':'-..-',
+    'Y':'-.--',
+    'Z':'--..',
+    ' ': ' '
+}
+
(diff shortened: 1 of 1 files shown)


chall02/antomart.py
```diff
@@ -0,0 +1,51 @@
...
+    ' ': ' '
+}
+
+# Control the cases of less or more arguments than demanded
+# Control isalpha or spaces
+# Turn to upper
+# for each char or space in the string write the corresponding char of our dictionary
+
+def main():
+    if len(sys.argv) == 2 and sys.argv[1] != "" and all(x.isalpha() or x.isspace() for x in sys.argv[1]):
```
clafoutis42, line 43: Good use of the comprehension.  
However you could have used a more pythonic approach here:
```suggestion
    if len(sys.argv) == 2 and sys.argv[1].replace(" ", "").isalpha():
```
Note that it also covers the `sys.argv[1] != ""` condition as `"".isalpha()` returns `False`

chall02/antomart.py
```diff
@@ -0,0 +1,51 @@
...
+        str = sys.argv[1].upper()
+        for char in str:
+            if (char.isalpha() or char == ' '):
+                sys.stdout.write(MORSE[char])
```
clafoutis42, line 47: You could have used the comprehensions here too to make it more pythonic:
```suggestion
        print("".join(MORSE[c] for c in str))

```

chall02/antomart.py
```diff
@@ -0,0 +1,51 @@
...
+    else:
+        print ("usage: ./antomart.py <a-zA-Z string>")
```
clafoutis42, line 49: You could have made the error message more dynamic, the filename could change:
```suggestion
        print ("usage: %s <a-zA-Z string>", sys.argv[0])
```

clafoutis42 (commented):
Good job, you have a working solution, however it is not perfectly pythonic (although you have the more pythonic approach I've seen so far). You are writing your code as if you were using C, I suggest you read [Fluent Python](https://www.amazon.com/Fluent-Python-Concise-Effective-Programming/dp/1491946008) if you want to keep progressing in python!

However, your file is not executable, I can't accept the PR :(

PR #105, opened 2020-04-06

victorsempere:
If everything is as expected; 1 param and no other chars than spaces and letters
For each character in the input parameter I replace the representation by the morse representation and finally print the result

### DIFF
--- a/chall02/vsempere.py
+++ b/chall02/vsempere.py
@@ -1,0 +1,35 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+import sys
+import re
+
+def printUsage():
+	print('usage: ./vsempere.py <a-zA-Z string>')
+
+def changeLetterByMorseRepresentation(letter):
+	map = {'a': '.-', 'b' : '-...', 'c':'-.-.', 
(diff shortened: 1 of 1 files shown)


aguimbao:

As stated in the README: You have to submit a python3 executable script to encode a string to morse., it should be executable as `./your_script.py` .
The shebang is present in your file, but we specified we wanted python3.x, so it should have been `#!/usr/bin/env python3`.
- I don't think `# -*- coding: utf-8 -*-` is necessary in a single-encoding project scenario.
- You print the whole list of sys.argv when there's the wrong number of arguments but this wasn't asked.
```python
if __name__ == "__main__":
    main()
```
- `sys.exit()` > `os.exit()/exit()/quit()/exit`.
- regex is a bit overkill for this simple problematic, `isalpha()` or filtering the characters in the string with the ones in the dictionary should be enough.
- I don't understand the `print('a')` in line 28.
- This is subjective, I really like camelcase for languages like Go, but in Python I would much rather like to see snakecase. So `def changeLetterByMorseRepresentation(letter):` -> `def change_letter_by_morse_representation(letter):`

You don't pass some of the tests :( and I can't accept your PR cause your script wasn't executable by default, sorry !

PR #106, opened 2020-04-06

hanjelito:


clafoutis42:
There is no description, I can't accept nor review this PR. Also you pushed C files when it was not meant to be part of the challenge.

PR #107, opened 2020-04-06
dalba-de serpens add

dalba-de:
First I define the dictionary with all the letters in lowercase and uppercase. The next command is to be able to use the arguments passed to the program. Later I declare an empty string that will serve to return the final value in morse. I loop the string changing each letter for its value in morse and I manage any errors.

### DIFF
--- a/chall02/dalba-de.py
+++ b/chall02/dalba-de.py
@@ -1,0 +1,33 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+dictionary = {"a": ".-", "b": "-...", "c": "-.-.", "d": "-..", "e": ".", "f": "..-.", 
+    "g": "--.", "h": "....", "i": "..", "j": "·---", "k": "-.-", "l": ".-..", 
+    "m": "--", "n": "-.", "o": "---", "p": ".--.", "q": "--.-",
+    "r": ".-.", "s": "...", "t": "-", "u": "..-", "v": "...-", "w": ".--",
+    "x": "-..-", "y": "-.--", "z": "--..",
+    "A": ".-", "B": "-...", "C": "-.-.", "D": "-..", "E": ".", "F": "..-.", 
+    "G": "--.", "H": "....", "I": "..", "J": "·---", "K": "-.-", "L": ".-..", 
+    "M": "--", "N": "-.", "O": "---", "P": ".--.", "Q": "--.-",
+    "R": ".-.", "S": "...", "T": "-", "U": "..-", "V": "...-", "W": ".--",
+    "X": "-..-", "Y": "-.--", "Z": "--.."}
+
+import sys
+
+texto = ""
+
+if len(sys.argv) == 2:
+    if len(sys.argv[1]) == 0:
+        print "usage: ./dalba-de.py <a-zA-Z string>"
+        sys.exit()
+    for c in sys.argv[1]:
+        if c in dictionary:
+            texto += dictionary[c]
+        elif c == " ":
+            texto += " "
+        else:
+            print "usage: ./dalba-de.py <a-zA-Z string>"
+            sys.exit()
+    print texto
+else:
+    print "usage: ./dalba-de.py <a-zA-Z string>"


chall02/dalba-de.py
```diff
@@ -0,0 +1,33 @@
...
+    "m": "--", "n": "-.", "o": "---", "p": ".--.", "q": "--.-",
+    "r": ".-.", "s": "...", "t": "-", "u": "..-", "v": "...-", "w": ".--",
+    "x": "-..-", "y": "-.--", "z": "--..",
+    "A": ".-", "B": "-...", "C": "-.-.", "D": "-..", "E": ".", "F": "..-.", 
+    "G": "--.", "H": "....", "I": "..", "J": "·---", "K": "-.-", "L": ".-..", 
+    "M": "--", "N": "-.", "O": "---", "P": ".--.", "Q": "--.-",
+    "R": ".-.", "S": "...", "T": "-", "U": "..-", "V": "...-", "W": ".--",
+    "X": "-..-", "Y": "-.--", "Z": "--.."}
+
+import sys
```
clafoutis42, line 15: Imports should always be at the top of the file

chall02/dalba-de.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+dictionary = {"a": ".-", "b": "-...", "c": "-.-.", "d": "-..", "e": ".", "f": "..-.",
```
clafoutis42, line 4: You could have made a dictionary with only lowercase or uppercase and later use `.lower()` or `.upper()` on your string

chall02/dalba-de.py
```diff
@@ -0,0 +1,33 @@
...
+    "R": ".-.", "S": "...", "T": "-", "U": "..-", "V": "...-", "W": ".--",
+    "X": "-..-", "Y": "-.--", "Z": "--.."}
+
+import sys
+
+texto = ""
+
+if len(sys.argv) == 2:
+    if len(sys.argv[1]) == 0:
+        print "usage: ./dalba-de.py <a-zA-Z string>"
```
clafoutis42, line 21: You should make this message more dynamic
```suggestion
        print("usage: %s <a-zA-Z string>" % sys.argv[0])
```
Note that I added parentheses, it is normal, you were supposed to turn-in your code in `python3.x` your turned it in in `python2.7` [which is deprecated](https://pythonclock.org/) since January the 1st of 2020.

chall02/dalba-de.py
```diff
@@ -0,0 +1,33 @@
...
+        sys.exit()
+    for c in sys.argv[1]:
+        if c in dictionary:
+            texto += dictionary[c]
+        elif c == " ":
+            texto += " "
+        else:
+            print "usage: ./dalba-de.py <a-zA-Z string>"
+            sys.exit()
+    print texto
```
clafoutis42, line 31: You could have simplified it with a more pythonic approach:
```suggestion
    if "".join(c for c in sys.argv[1] if c in dictionary.keys()) != sys.argv[1]:
        sys.exit("usage: %s <a-zA-Z string>" % sys.argv[0])
    print("".join(dictionary.get(c, " ") for c in sys.argv[1]))
```

chall02/dalba-de.py
```diff
@@ -0,0 +1,33 @@
...
+    "M": "--", "N": "-.", "O": "---", "P": ".--.", "Q": "--.-",
+    "R": ".-.", "S": "...", "T": "-", "U": "..-", "V": "...-", "W": ".--",
+    "X": "-..-", "Y": "-.--", "Z": "--.."}
+
+import sys
+
+texto = ""
+
+if len(sys.argv) == 2:
+    if len(sys.argv[1]) == 0:
```
clafoutis42, line 20: You could also have simplified this 2 conditions in one condition

clafoutis42 (commented):
You seem to have a working solution, although it might be more pythonic (I recommend [Fluent Python](https://www.amazon.com/Fluent-Python-Concise-Effective-Programming/dp/1491946008) to learn the good ways of Python).

clafoutis42:
Unfortunately you submitted your code in `Python2.7` when it was expected in `Python3.x`, I can't accept your PR.

PR #108, opened 2020-04-06
edramire - Constelacion Serpens

edramir18:
This script translate a string to morse code, using a python dictionary to store the morse code and a regular expression to validate that all characters on the input string are letters from A to Z and from a to z  plus the space character.

The script show an error message in case the number of arguments is diferent of 2, the input string is empty or has any not valid character.

After the validation is complete the string is convert to morse code letter to letter and output as a final string


chall02/edramire.py
```diff
@@ -0,0 +1,49 @@
...
+	" ": " "
+}
+msg = "usage: ./edramire.py <a-zA-Z string>"
+argn = len(sys.argv)
+if argn != 2:
+	print(msg)
+else:
+	string = sys.argv[1]
+	str_find = re.search(r'[^A-Za-z ]', string)
+	if (len(string) == 0 or str_find):
```
anna-dominguez, line 42: No need of parenthesis in Python, this looks a lot like C

chall02/edramire.py
```diff
@@ -0,0 +1,49 @@
...
+	"Z": "--..",
+	" ": " "
+}
+msg = "usage: ./edramire.py <a-zA-Z string>"
+argn = len(sys.argv)
+if argn != 2:
+	print(msg)
+else:
+	string = sys.argv[1]
+	str_find = re.search(r'[^A-Za-z ]', string)
```
anna-dominguez, line 41: You could use `isascii` or `isalpha` instead of a regexp, or if you're afraid to find a character that is not allowed, check if the characters received are in your dictionary

chall02/edramire.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python
```
anna-dominguez, line 1: Good 👍But it's always better to specified the version, here we want `python3` as a minimum version. I recommend you to read some documentation about this, there a lot of debates about shebang

chall02/edramire.py
```diff
@@ -0,0 +1,49 @@
...
+
+import sys
+import re
+
+morse = {
```
anna-dominguez, line 6: If you use globals in Python, it's better write them in upper case

chall02/edramire.py
```diff
@@ -0,0 +1,49 @@
...
+if argn != 2:
+	print(msg)
+else:
+	string = sys.argv[1]
+	str_find = re.search(r'[^A-Za-z ]', string)
+	if (len(string) == 0 or str_find):
+		print (msg)
+	else:
+		string = string.upper()
+		str_out = ""
```
anna-dominguez, line 46: Why use a new var instead of writing the result directly ? It use more memory, it's less efficient

anna-dominguez:
Very good job, it's working perfectly 👍 
I made you some reviews in your code directly. I'll add that it's a better habit to do a main in Python. Like @abguimba said:

- It doesn't matter how simple the script is, you may want to import it's contents without executing the file one day so it's good practice to implement a main() function and set the appropriate:

```python
if __name__ == "__main__":
    main()```

chall02/edramire.py
edramir18: Thanks for the input. I will take your advices for future projects.

PR #109, opened 2020-04-06
string to morse v1.0

Antogor:
I've make the all dictionary with upper and lowers letters, and i've tried to print the string passed to his value in the dictionary. For that i've created a function to compare if the character is alphabetic and if it is, with a "for" loop print his value in morse code.

For do all this i searched how work the dictionary in python and how to use it, and how this lenguage work whit the arguments.  Loops and conditionals are similar to C, so i did tests and search how syntax was


chall02/agarzon-.py
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env python3
```
anna-dominguez, line 1: Good 👍

chall02/agarzon-.py
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env python3
+
+import sys
+import argparse
+
+morse = {
```
anna-dominguez, line 6: If you use globals in Python, write them in upper case.
Here in your dictionary you could write the letter in upper or lower case, but not both and then change all what you receive in upper or lower case in order to retreive the character in the dictionary.

chall02/agarzon-.py
```diff
@@ -0,0 +1,87 @@
...
+	"W": ".--",
+	"x": "-..-",
+	"X": "-..-",
+	"y": "-.--",
+	"Y": "-.--",
+	"z": "--..",
+	"Z": "--.."
+}
+
+def is_alpha(s):
```
anna-dominguez, line 62: Good idea of function, with it you didn't used `isascii` or `isalpha` functions and could check the characters required

anna-dominguez (commented):
Too bad your file is not executable, there a lot of good things and also some code reviews to do.  I made some comments in your code. And i'll add that you it's a good practice to do a `main` instead of writing your code directly.  As @abguimba said: 

- It doesn't matter how simple the script is, you may want to import it's contents without executing the file one day so it's good practice to implement a main() function and set the appropriate:
```python
if __name__ == "__main__":
    main()```

PR #110, opened 2020-04-06
Chall02

GonzaloMartinCano:
To solve this problem we need:
-That the number of arguments is 2.
-That the string argument isn't a empty string.
-That the string argument that we want to convert contains only the characters defined in our dictionary in addition to the spaces.
If these conditions are correct we only have to go through our string changing each initial character for the corresponding one in our dictionary and put it in a new string that we will print.
Otherwise our program prints an error message.


aguimbao:

As stated in the README: `You have to submit a python3 executable script to encode a string to morse`., it should be executable as `./your_script.py`.
We specified we wanted python3.x, so it should have been `#!/usr/bin/env python3`.
- I don't think `# -*- coding: utf-8 -*-` is necessary in a single-encoding project scenario.
```python
if __name__ == "__main__":
    main()
```
- `sys.exit()` > `os.exit()`/`exit()`/`quit()`/`exit`.
- You store every character two times in the dictionary, one in lowercase and one in uppercase, you could have simply used the method `.lower()` on the string argument to convert the whole string to the same case and then only have to store the alphabet once, in lowercase.
-Your code looks like C, you should try to be more "pythonic".

You don't pass some of the tests, you don't print your usage correctly is wrong :( and I can't accept your PR cause your script wasn't executable by default, sorry !

PR #111, opened 2020-04-06
version 0.1 various args

bmk101:


bmk101:
A simple text_Morse converter. Python 3.x

chall02/rgallard.py
```diff
@@ -0,0 +1,48 @@
+import sys
```
anna-dominguez, line 1: A shebang is missing

chall02/rgallard.py
```diff
@@ -0,0 +1,48 @@
...
+
+morse = {
```
anna-dominguez, line 3: Better to write globals in upper case

chall02/rgallard.py
```diff
@@ -0,0 +1,48 @@
...
+	'U':'..-',
+	'V':'...-',
+	'W':'.--',
+	'X':'-..-',
+	'Y':'-.--',
+	'Z':'--..',
+	' ':' '
+	}
+
+i = 1
```
anna-dominguez, line 33: Why ?

chall02/rgallard.py
```diff
@@ -0,0 +1,48 @@
...
+j = 0
```
anna-dominguez, line 34: Why ?

chall02/rgallard.py
```diff
@@ -0,0 +1,48 @@
...
+talk = ""
+
+if (2 > len(sys.argv)):
+	talk = "<a-zA-Z string>"
+while (i < len(sys.argv)):
```
anna-dominguez, line 39: Why use index in Python 😱 it's not C, you don't need to iterate through your array or string with an index, you can use `for` as you did it below

chall02/rgallard.py
```diff
@@ -0,0 +1,48 @@
...
+	for j in sys.argv[i]:
```
anna-dominguez, line 40: You didn't need to declare `j` before

chall02/rgallard.py
```diff
@@ -0,0 +1,48 @@
...
+		j = j.upper()
+		if (j in morse.keys()):
```
anna-dominguez, line 42: Good 👍 it's a better way to check if it's a writable character than use the functions `isascii` or `isalpha` but it makes you declare a variable in order to store your result when you could write if you've done the checks before

anna-dominguez (commented):
Well done not forgetting to send an executable. But where is you comment explaining your code ??

PR #112, opened 2020-04-06
Chall02--Str2Morse

MiguelTolino:
At first, I save a dictionary with morse code values. Then I check the number of arguments passed and if the string is not empty, so I look for values ​​that are not in the dictionary. If there is none, the chain is wrong. Finally I print them.


aguimbao:
You should explain a bit more in the comment how you retrieve which morse code correspond to each character.

Your code is not bad, however you could improve in some places:

- Your file was executable (755 permissions) but can't be executed in the way we specified.
As stated in the README: `You have to submit a python3 executable script to encode a string to morse`., it should be executable as `./your_script.py`.
We specified we wanted python3.x, so it should have been `#!/usr/bin/env python3`.
```python
if __name__ == "__main__":
    main()
```
- Global variables like `morse` should be uppercase.
- `sys.exit()` is the recommended way to exit a program, however in Python it's not needed and not used like that at the end of the program, it looks like C. You don't need to "`return (0);`" in Python.
-Your code looks like C, you should try to be more "pythonic".
- You should have a function for usage to not have redundant code lines.

However I like how you convert all string to lowercase and how you don't store the whole string :) saves memory.
It's a shame cause you pass all the tests. I will accept this PR because your file is executable, even if it's lacking the shebang.

PR #113, opened 2020-04-06

abuadev:
#
Implemented shebang to indicate (in unix system) where the environment variable "python3" is.
#
Solved with a dictionary to list morse alphabet with a key, value pair based on modern english alphabet (including space character for a more friendly implementation).
#
Defined functions to obtain the command usage and the morse message, in order to separate the logic from the visual.
#
Implemented try/except to handling all possibles errors and exceptions we don't want in our program.
#


chall02/cvinolo-.py
```diff
@@ -0,0 +1,83 @@
...
+" ": " "
+}
+
+def get_usage():
+	return "usage: ./cvinolo-.py <a-zA-Z string>"
+
+def get_morse_message(message):
+	morse_message = ""
+	message = message.lower()
+	try:
```
anna-dominguez, line 69: It is better to use conditional check as you did line `77`
```python
if "blah" in B:
    A["blah"] = B["blah"]
```
 than KeyError exception for dictionary key 
```python
try:
    A["blah"] = B["blah"]
except KeyError:
    pass
```

You didn't specified the exceptions and it will catch all of them

chall02/cvinolo-.py
```diff
@@ -0,0 +1,83 @@
...
+"y": "-.--",
+"z": "--..",
+" ": " "
+}
+
+def get_usage():
+	return "usage: ./cvinolo-.py <a-zA-Z string>"
+
+def get_morse_message(message):
+	morse_message = ""
```
anna-dominguez, line 67: You could do this without declaring a new var, printing directly your result if it pass your checks.

chall02/cvinolo-.py
```diff
@@ -0,0 +1,83 @@
...
+#Defined functions to obtain the command usage and the
+#morse message, in order to separate the logic from the visual.
+#
+#Implemented try/except to handling all possibles errors and
+#exceptions we don't want in our program.
+#
+
+import sys
+
+morse_alphabet = {
```
anna-dominguez, line 33: If you use global it's better to write them in upper case

chall02/cvinolo-.py
```diff
@@ -0,0 +1,83 @@
...
+"u": "..-",
+"v": "...-",
+"w": ".--",
+"x": "-..-",
+"y": "-.--",
+"z": "--..",
+" ": " "
+}
+
+def get_usage():
```
anna-dominguez, line 63: Good 👍

chall02/cvinolo-.py
```diff
@@ -0,0 +1,83 @@
...
+	morse_message = ""
+	message = message.lower()
+	try:
+		for letter in message:
+			morse_message += morse_alphabet[letter]
+		return morse_message
+	except:
+		return get_usage()
+
+def main():
```
anna-dominguez, line 76: Good 😍

anna-dominguez (commented):
Sadly you didn't send an executable file take care to read the `README` perfectly next time

chall02/cvinolo-.py
abuadev: I'd like to know why it's better to write them in upper case. In [solution.py](https://github.com/42madrid/remote-challs/blob/master/chall02/solution.py#L5) you can see on line `5` a global variable in lower case. Also, [doing a search](https://www.google.com/search?sxsrf=ALeKk03YnGGBvNYqptHJ88yEYEW_X7NAcg%3A1586377506435&ei=IjOOXpWFGpLMgweIrYzADQ&q=python+naming+convention+global+variable&oq=Python+naming+convention+globa&gs_lcp=CgZwc3ktYWIQARgAMgUIABDLATIICAAQCBANEB4yCAgAEAgQDRAeOgQIABBHOgQIIxAnOgQIABBDOgUIABCDAToHCAAQFBCHAjoCCAA6BggAEBYQHkosCBcSKDBnMTAyZzk0Zzk5ZzEwMWc5MWc5MWc4N2cxMDBnODlnOTBnOTVnNjhKHggYEhowZzFnMWcxZzFnMWcxZzFnMWcxZzVnNWcyMFCH-RFYkNISYJHaEmgAcAh4AIABggGIAZsVkgEEMjUuNpgBAKABAqABAaoBB2d3cy13aXo&sclient=psy-ab) you're always going to find something related to:
- **Global variables** should be all lowercase.
- Words in a **global variable name** should be separated by an underscore.

aguimbao: According to PEP8s rule on constants:

`Constants are usually defined on a module level and written in all capital letters with underscores separating words. Examples include MAX_OVERFLOW and TOTAL.`

In this case, `morse_alphabet` is used as constant (constant keyword doesn't exist in Python but the concept does). Style guides are really subjective, but PEP8 is the most used and respected one.

As for `solution.py`, you will have to ask pistache 😅. Just kidding, we should have said that "we recommend" writing them in upper case, not that "it's better" to write them in upper case.

PR #114, opened 2020-04-06
estina : my version of Python problem

xplore93:
Hi there, it is my version of solution for the Python's problem.
First of all, the idea. To solve this, I use the Dictionary to define each MORSE code for letters from **a** to **z**.

- Check if the amount of arguments is 2 _and_ the second argument is not empty.
- Declare a variable for the coded **string**.
- Passing through all the characters of the argument, check if they are in the dictionary. _If not, exit with a message of usage._
- Put into our **string** code of the current character.
- When all the mess ends, print the coded string in the terminal.

I was thinking about to add the ' if __name__ == "main" ' block but I see that unnecessary here.

### DIFF
--- a/chall02/estina.py
+++ b/chall02/estina.py
@@ -11,10 +11,8 @@
 #                                                                              #
 # **************************************************************************** #
 
-# For arguments
 import sys
 
-# Declare our dictionaire for the MORSE
 CHAR_TO_MORSE = {
 	" ": " ",
 	"a": ".-",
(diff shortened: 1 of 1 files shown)


aguimbao:
I corrected your code even if there was a comment after a Deadline so it won't be accepted :( I'm sorry, think about it next time.

- File execution permissions was good, and the shebang was nice, maybe `#!/usr/bin/env python3` would have been better than `#!/usr/bin/python3` in case I don't have my python interpreter installed in the same place as you, I would recommend you looking into python shebangs.
```python
if __name__ == "__main__":
    main()
```
It's nice because you thought about this, I will always add the `main()` functionality even to my simplest scripts.
- You should have a function for usage to not have redundant code lines.
- `sys.exit()` > `os.exit()/exit()/quit()/exit`.

However I like how you use uppercase for your global variables and how you lower() your charactes.
There's a big fail though, you check for `c.lower()` in line 50, but when you try to add the morse code to the string in line 52, you don't repeat the process of lowering the case. (`c.lower()` won't transform the variable, it will just return that variable in lower case). So your dictionary lookup fails :( It's a shame cause without this error you would pass all tests !

aguimbao:
😅Oups little fail on my side accepting your PR. Had to revert it.

PR #115, opened 2020-04-06
cdorta-n Chall02

Top64:
Hello! First tome on Python~~ what a curious program language.

_So_, first, we prepare the functions to use. In order to avoid calling our morse "library" when the arguments are bad presented, this is the first check. If all is correct, we "change" the string to lower letter in order to use our "morse library". 

The last part is check each character, rewrite it if (is not space) and then, write the answer. If a "not expected character" is found, an error occur.

*This .py file was programmed on Windows 10* Could "but it wouldn't" fail on other system.

### DIFF
--- a/chall02/cdorta-n.py
+++ b/chall02/cdorta-n.py
@@ -1,0 +1,40 @@
+#/usr/bin/python3
+# -*- coding: utf-8 -*-
+
+import sys
+
+num = len(sys.argv)
+
+if num != 2:
+        print("usage: ./cdorta-n.py <a-zA-Z string>")
+        exit(0)
+else:
+    codigo_morse = {
+        "a": ".-", "b": "-...", "c": "-.-.", "d": "-..", "e": ".", "f": "..-.", 
+        "g": "--.", "h": "....", "i": "..", "j": "·---", "k": "-.-", "l": ".-..", 
+        "m": "--", "n": "-.", "ñ": "--.--", "o": "---", "p": ".__.", "q": "--.-",
+        "r": ".-.", "s": "...", "t": "-", "u": "..-", "v": "...-", "w": ".--",
+        "x": "-..-", "y": "-.--", "z": "--..",
+        
+        "0": "-----", "1": ".----", "2": "..---", "3": "...--", "4": "....-", 
+        "5": ".....", "6": "-....", "7": "--...", "8": "---..", "9": "----.",
+        
+        ".": ".-.-.-", ",": "-.-.--", "?": "..--..", "\"": ".-..-."
+    }
+    texto_codificado = ''
+
+    len = len(sys.argv[1])
+    sys.argv[1] = sys.argv[1].lower()
+    contador = 0
+
+    while (contador < len):
+        if not sys.argv[1][contador].isalpha():
+            if not sys.argv[1][contador].isspace():
+                print("usage: ./cdorta-n.py <a-zA-Z string>")
+                exit(0)
+        if sys.argv[1][contador] != " ":
+            texto_codificado += codigo_morse[sys.argv[1][contador]]
+        else:
+            texto_codificado += sys.argv[1][contador]
+        contador += 1
+    print(texto_codificado)


chall02/cdorta-n.py
```diff
@@ -0,0 +1,40 @@
...
+        exit(0)
+else:
+    codigo_morse = {
+        "a": ".-", "b": "-...", "c": "-.-.", "d": "-..", "e": ".", "f": "..-.", 
+        "g": "--.", "h": "....", "i": "..", "j": "·---", "k": "-.-", "l": ".-..", 
+        "m": "--", "n": "-.", "ñ": "--.--", "o": "---", "p": ".__.", "q": "--.-",
+        "r": ".-.", "s": "...", "t": "-", "u": "..-", "v": "...-", "w": ".--",
+        "x": "-..-", "y": "-.--", "z": "--..",
+        
+        "0": "-----", "1": ".----", "2": "..---", "3": "...--", "4": "....-",
```
anna-dominguez, line 19: Alpha characters shouldn't work in your program

chall02/cdorta-n.py
```diff
@@ -0,0 +1,40 @@
...
+        "a": ".-", "b": "-...", "c": "-.-.", "d": "-..", "e": ".", "f": "..-.", 
+        "g": "--.", "h": "....", "i": "..", "j": "·---", "k": "-.-", "l": ".-..", 
+        "m": "--", "n": "-.", "ñ": "--.--", "o": "---", "p": ".__.", "q": "--.-",
+        "r": ".-.", "s": "...", "t": "-", "u": "..-", "v": "...-", "w": ".--",
+        "x": "-..-", "y": "-.--", "z": "--..",
+        
+        "0": "-----", "1": ".----", "2": "..---", "3": "...--", "4": "....-", 
+        "5": ".....", "6": "-....", "7": "--...", "8": "---..", "9": "----.",
+        
+        ".": ".-.-.-", ",": "-.-.--", "?": "..--..", "\"": ".-..-."
```
anna-dominguez, line 22: Those characters shouldn't work too

chall02/cdorta-n.py
```diff
@@ -0,0 +1,40 @@
...
+
+import sys
+
+num = len(sys.argv)
+
+if num != 2:
+        print("usage: ./cdorta-n.py <a-zA-Z string>")
+        exit(0)
+else:
+    codigo_morse = {
```
anna-dominguez, line 12: Variable name in spanish 🤔

chall02/cdorta-n.py
```diff
@@ -0,0 +1,40 @@
...
+        "0": "-----", "1": ".----", "2": "..---", "3": "...--", "4": "....-", 
+        "5": ".....", "6": "-....", "7": "--...", "8": "---..", "9": "----.",
+        
+        ".": ".-.-.-", ",": "-.-.--", "?": "..--..", "\"": ".-..-."
+    }
+    texto_codificado = ''
+
+    len = len(sys.argv[1])
+    sys.argv[1] = sys.argv[1].lower()
+    contador = 0
```
anna-dominguez, line 28: Variable name in spanish2 😱

chall02/cdorta-n.py
```diff
@@ -0,0 +1,40 @@
+#/usr/bin/python3
+# -*- coding: utf-8 -*-
+
+import sys
+
+num = len(sys.argv)
+
+if num != 2:
+        print("usage: ./cdorta-n.py <a-zA-Z string>")
```
anna-dominguez, line 9: You should extract you error handling message because it's not the only time that you'll need it

chall02/cdorta-n.py
```diff
@@ -0,0 +1,40 @@
+#/usr/bin/python3
```
anna-dominguez, line 1: It's not a good way to write a shebang, I invite you to read documentation about it, it'll be fun: there is some debate about it

anna-dominguez (commented):
Too bad you file is not executable, I think the logic of your code is good, but your doing C in Python. You should read more about Python and it's good practice. For example you don't have a main in your file, or doing a global for your morses characters, etc

PR #116, opened 2020-04-06
chall02

misterjsr:
First see some first step of python, which I have no idea ... First, after importing sys, we declare the variable with the signs. Then with a loop, we return the characters as they appear. In case they are not in the array we return an error message.


aguimbao:
Hey, congrats on trying out Python, it's honestly not bad for a beginner.

Some things to note:

As stated in the README: You have to submit a python3 executable script to encode a string to morse., it should be executable as `./your_script.py`.
You have a shebang, which is good, however if my Python3 interpreter is not installed in the place your shebang specifies, it won't work. `#!/usr/bin/env python3` is much better.
I'll let you take a look into Python shebangs and ways to execute Python modules cause it's a lengthy topic.
- You should have implemented a usage function to not have redundant code lines.
- `print()` > `sys.stdout.write/sys.stderr.write`
- After line 26, you should have `sys.exit()` to not waste more time. (`sys.exit()` > `os.exit()/exit()/quit()/exit`.)
- It's nice that you print character by character instead of wasting memory storing the whole string, however, you don't sanitise the string before. So it can happen that you print half of the string, and then you find a forbidden character, which would be wrong. 
- Maybe it would have been better to store the morse dictionary in a global variable for readability purposes, but in this case it's a bit subjective as local variables are faster and waste less memory.
- You should sanitise output, `isalpha()` method for characters for example. Or just checking if the character in the string is present in the hardcoded dictionary. Your script crashes for invalid output like '#' or '\\'.

I really like how you implemented the `main()` functionality. I always do it, even for simple scripts, you never know when you are going to want to import it, even if it's a small simple module. 
For being your first time, it's really nice, very "pythonic" for someone who just started out, keep going !

I won't be able to accept your PR cause of execution rights, but don't worry, most of your colleagues had the same problem ;)

PR #117, opened 2020-04-06
Chall 02 gconde-m

gconde-m2:
first i  makes a library of the words in morse, I checked  that there are just 1 arguments and that there are something in it, then I just started a loop for check all the words in argv[1], using a flag to check it, its all is okay, I print argv.
my English sucks so hard.
thanks.


chall02/gconde-m.py
```diff
@@ -0,0 +1,57 @@
+from argparse import ArgumentParser
```
anna-dominguez, line 1: A shebang is missing, your should read good practice documentation about it, it's important and there's some fun debates about it 😄

chall02/gconde-m.py
```diff
@@ -0,0 +1,57 @@
...
+import sys
+import os
+
+morse = {
```
anna-dominguez, line 5: Your globals must be in uppercase

chall02/gconde-m.py
```diff
@@ -0,0 +1,57 @@
...
+'v':'...-', 
+'w':'.--', 
+'x':'-..-', 
+'y':'-.--', 
+'z':'--..',
+' ':' '
+}
+
+str = ''
+i = 0;
```
anna-dominguez, line 36: Why ?

chall02/gconde-m.py
```diff
@@ -0,0 +1,57 @@
...
+y = 0;
```
anna-dominguez, line 37: Why ?

chall02/gconde-m.py
```diff
@@ -0,0 +1,57 @@
...
+if len(sys.argv) != 2 or not sys.argv[1]:
+    print("<a-zA-Z string>")
```
anna-dominguez, line 39: You forget `./gconde-m`, it must be: `./gconde-m <a-zA-Z string>`

chall02/gconde-m.py
```diff
@@ -0,0 +1,57 @@
...
+    exit()
+else :
+    while i < len(sys.argv[1]):
```
anna-dominguez, line 42: In Python you should use 
```python
for elem in tab
```

more than use indexes. As you did your code, it seems like a C program. You should read more about Python and practice

anna-dominguez (commented):
Too bad your file is not executable. Don't forget to read my comments in your code. My best advice for you is to read documentation about Python, and train. Python is not like C, it's should readable, more simplistic

PR #118, opened 2020-04-06
chall02 ellaca-f

TECHedgehog:
First I declare an interpreter to run the program. Then I import sys to be able to pass arguments and also re to use compile function from this library.
Now I declare a dictionary with the translation of every word from a to z to morse.

In line 38, I test if there are more than 2 args or less, and return error string if that's true.
Line 40, error test of a empty string.
Line 42, error test of not alpha chars. I've used a regex function(re.compile) that create a string with only letters from a-z and A-Z to use it as a pattern to compare sys.argv[1] and be able to know if there is not alpha char.
Finally I print with a for loop every letter translated to morse using the dictionary and converted previously to lowercase. Also I add an end of line to the end.

### DIFF
--- a/chall02/ellaca-f.py
+++ b/chall02/ellaca-f.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python3
+import sys
+import re
+
+pattern = re.compile("[A-Za-z ]+")
+
+dic = {
+    "a":".-",
+    "b":"-...",
+    "c":"-.-.",
+    "d":"-..",
+    "e":".",
+    "f":"..-.", 
+    "g":"--.",
+    "h":"....", 
+    "i":"..",       
+    "j":".---", 
+    "k":"-.-",
+    "l":".-..", 
+    "m":"--",         
+    "n":"-.",         
+    "o":"---",
+    "p":".--.", 
+    "q":"--.-", 
+    "r":".-.",
+    "s":"...",
+    "t":"-",        
+    "u":"..-",
+    "v":"...-", 
+    "w":".--",
+    "x":"-..-", 
+    "y":"-.--", 
+    "z":"--..",
+    " ":" "
+
+}
+
+if len(sys.argv) != 2:
+    print("usage: ./xlogin.py <a-zA-Z string>")
+elif sys.argv[1] == "":
+    print("usage: ./xlogin.py <a-zA-Z string>")
+elif pattern.fullmatch(sys.argv[1]) is None:
+    print("usage: ./xlogin.py <a-zA-Z string>")
+else:
+    for c in sys.argv[1]:
+        print(dic[c.lower()], end="")
+    print()
+    


chall02/ellaca-f.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python3
```
anna-dominguez, line 1: Good 👍

chall02/ellaca-f.py
```diff
@@ -0,0 +1,48 @@
...
+import sys
+import re
+
+pattern = re.compile("[A-Za-z ]+")
+
+dic = {
```
anna-dominguez, line 7: The global name should be in upper case in Python

chall02/ellaca-f.py
```diff
@@ -0,0 +1,48 @@
...
+    "z":"--..",
+    " ":" "
+
+}
+
+if len(sys.argv) != 2:
+    print("usage: ./xlogin.py <a-zA-Z string>")
+elif sys.argv[1] == "":
+    print("usage: ./xlogin.py <a-zA-Z string>")
+elif pattern.fullmatch(sys.argv[1]) is None:
```
anna-dominguez, line 42: Why don't you do an only condition for those error handling ? -4 lines

chall02/ellaca-f.py
```diff
@@ -0,0 +1,48 @@
...
+    print("usage: ./xlogin.py <a-zA-Z string>")
```
anna-dominguez, line 43: You what you should return in case of error is `./xlogin` what is `./ellaca-f.py <a-zA-Z string>`

anna-dominguez (commented):
Too bad you didn't make an executable. You should read documentation about Python, I see a main is missing too

chall02/ellaca-f.py
TECHedgehog: It was giving me some problems under certain circumstances, like for example when sys.argv[1] did not exist... probably i was doing something bad... i was a bit lost. I never touched Py before... so i take the easy way kkkk

TECHedgehog:
So, what does this mean then?? You close my PR but does this count as if I have participated?? I spend a whole day alone learning python and solving problems, It's more than normal to commit mistakes i think... that's what participating means, just to try it... isn't it??
Thanks

anna-dominguez:
I saw your messages in the thread today, I'm sorry that it was not more specified than this. If you remember rushes in piscine it looks a lot like it, if you don't have in your PR what it was asked for the basis of the program we can't execute it, it's like we didn't receive anything. 
We made all of your reviews knowing that all was going to be hard to test. I hope that if you've done it, what you've learned is the most important. Now you know that in one day you can learn Python, or another programing language and do a little program.
Hope to see another of your  PR tomorrow !

TECHedgehog:
Thanks Anna (; dont worry I just wanted to clarify that... but it's ok, no problem... thanks for your help... i'll try to do something with the arrays. At least I will put a main... jjjj

PR #120, opened 2020-04-06
String to Morse

jtrancoso:
First I import sys to be able to work with arguments.

Then I create the morse dictionary for both lowercase and uppercase and I also get spaces inside it.

After that I make sure that it doesn't  work if only one argument is entered or if the string is empty.

And finally to translate the sentence to morse, I check if characters are alpha or spaces and translate them with my dictionary into a new string.


chall02/jtrancos.py
```diff
@@ -0,0 +1,38 @@
...
+#    Created: 2020/04/06 19:47:21 by jtrancos          #+#    #+#              #
+#    Updated: 2020/04/06 22:58:06 by jtrancos         ###   ########.fr        #
+#                                                                              #
+# **************************************************************************** #
+
+#!/usr/bin/env python
+
+import sys
+
+morse = {
```
anna-dominguez, line 17: Write your global names in upper case

chall02/jtrancos.py
```diff
@@ -0,0 +1,38 @@
...
+#    jtrancos.py                                        :+:      :+:    :+:    #
+#                                                     +:+ +:+         +:+      #
+#    By: jtrancos <marvin@42.fr>                    +#+  +:+       +#+         #
+#                                                 +#+#+#+#+#+   +#+            #
+#    Created: 2020/04/06 19:47:21 by jtrancos          #+#    #+#              #
+#    Updated: 2020/04/06 22:58:06 by jtrancos         ###   ########.fr        #
+#                                                                              #
+# **************************************************************************** #
+
+#!/usr/bin/env python
```
anna-dominguez, line 13: Good 👍 but you should read documentation about shebang, the Python version is missing you should write `python3`, not just `python`

anna-dominguez (commented):
Your file is not an executable as it is required in the `README` so it can't be executed by my tests. You should read documentation about good practice in Python, a main is missing

chall02/jtrancos.py
jtrancoso: So something like this would be correct?


#!/usr/bin/env python3

import sys

Morse = {
        'A': '.-', 'a': '.-', 'B': '-...', 'b': '-...','C': '-.-.', 'c': '-.-.',        
        'D': '-..', 'd': '-..', 'E': '.', 'e': '.', 'F': '..-.', 'f': '..-.',
        'G': '--.', 'g': '--.', 'H': '....', 'h': '....', 'I': '..', 'i': '..',
        'J': '.---', 'j': '.---', 'K': '-.-', 'k': '-.-', 'L': '.-..', 'l': '.-..',     
        'M': '--', 'm': '--', 'N': '-.', 'n': '-.', 'O': '---', 'o': '---', 'P': '.--.',
        'p': '.--.', 'Q': '--.-', 'q': '--.-', 'R': '.-.', 'r': '.-.', 'S': '...',      
        's': '...', 'T': '-', 't': '-', 'U': '..-', 'u': '..-', 'V': '...-',
        'v': '...-', 'W': '.--', 'w': '.--', 'X': '-..-', 'x': '-..-', 'Y': '-.--',     
        'y': '-.--', 'Z': '--..', 'z': '--..', ' ': ' '
}

if __name__ == "__main__"
        if len(sys.argv) != 2 or sys.argv[1] == "":
                print ("usage: ./jtrancos.py <a-zA-Z string>")
                sys.exit()
        if all(x.isalpha() or x.isspace() for x in sys.argv[1]):
                text = ''
                for letter in sys.argv[1]:
                        text = text + (morse[letter])
                print(text)
        else:
                print ("usage: ./jtrancos.py <a-zA-Z string>")

anna-dominguez: Well, that's a start 😄 
```python
#!/usr/bin/env python3

import sys

MORSE = {
'A': '.-',
'B': '-...', 
'C': '-.-.',
'D': '-..',
'E': '.',
'F': '..-.',
'G': '--.',
'H': '....', 
'I': '..',
'J': '.---',
'K': '-.-',
'L': '.-..',
'M': '--',
'N': '-.',
'O': '---', 
'P': '.--.',
'Q': '--.-',
'R': '.-.',
'S': '...',
'T': '-',
'U': '..-',
'V': '...-',
'W': '.--',
'X': '-..-',
'Y': '-.--',
'Z': '--..',
' ': ' '
}

def main():
  if len(sys.argv) != 2 or sys.argv[1] == "":
    print ("usage: ./jtrancos.py <a-zA-Z string>")
    sys.exit()
  if all(x.isalpha() or x.isspace() for x in sys.argv[1]):
    text = ''
    for letter in sys.argv[1]:
      text = text + (morse[letter])
    print(text)
  else:
    print ("usage: ./jtrancos.py  <a-zA-Z string>")

if __name__ == '__main__':
  main()
```

anna-dominguez: You don't need the characters in upper case and in lower case. If you want to simplify your code, you can do a function outside for your error handling. Now that you don't have upper and lower case, instead of doing check with `isalpha` `isspace` you can look if your characters are in your `MORSE` variable doing `uppercase`. Don't forget too that your error message wasn't right, you forgot a part of it

jtrancoso: Oh nice I see, yeah that would make it easier, thank you! :)

And for the error message would something like this be better?:
`def error():
print "usage: sys.argv[0] <a-zA-Z string>"`

anna-dominguez: Yes, for example. But with another function name because `error` could be use by your language, and be interpreted as other thing than your own function. 
Doing `print "usage: sys.argv[0] <a-zA-Z string>"` is a syntax of python2 in python3 you need to use `print("[...]")`
Don't forget your `exit` too.
It would be:
```python
def error():
  print(f"usage: {sys.argv[0]} <a-zA-Z string>")
  sys.exit()
```

jtrancoso: thank you!

PR #121, opened 2020-04-07
agutierr

obispowned:
Encriptador a Morse, con python


clafoutis42:
You were supposed to explain your code as mentioned in the root README. We won't be reviewing nor acception this pull request.

PR #122, opened 2020-04-07
string in morse

nsoria042:
First we control the number of arguments, two in this case, and that the string passed as an argument includes only letters. Otherwise we show the indicated message.
Then we pass the string to lowercase. 
Finally, by means of two tuples, one with the letters and the other with the morse symbols, we proceed to replace each other with each other. Finally we close the string, so that there is no newline at the end.

### DIFF
--- a/chall02/isoria-g.py
+++ b/chall02/isoria-g.py
@@ -1,0 +1,12 @@
+import sys
+letter=("a","b","c","d","e","f","g","h","i","j","k","l","m","n","o","p","q","r","s","t","u","v","w","x","y","z")
+morse=(".-","-...","-.-.","-..",".","..-.","--.","....","..",".---","-.-",".-..","--","-.","---",".--.","--.-",".-.","...","-","..-","...-",".--","-..-","-.--","--..")
+if len(sys.argv) == 2 and (all(c.isalpha() for c in sys.argv[1])):
+    string=(sys.argv[1]).lower()
+    i = 0
+    while i <= 25:
+        string = string.replace(letter[i], morse[i])
+        i=i+1
+    print(string, end='')
+else:
+    print ("usage: ./xlogin.py <a-zA-Z string>")


chall02/isoria-g.py
```diff
@@ -0,0 +1,12 @@
+import sys
+letter=("a","b","c","d","e","f","g","h","i","j","k","l","m","n","o","p","q","r","s","t","u","v","w","x","y","z")
+morse=(".-","-...","-.-.","-..",".","..-.","--.","....","..",".---","-.-",".-..","--","-.","---",".--.","--.-",".-.","...","-","..-","...-",".--","-..-","-.--","--..")
```
clafoutis42, line 3: Instead of tuple you could have used a dictionary as follows:
```suggestion
morse = code = {
    "a": ".-",
    "b": "-...",
    "c": "-.-.",
    "d": "-..",
    "e": ".",
    "f": "..-.",
    "g": "--.",
    "h": "....",
    "i": "..",
    "j": ".---",
    "k": "-.-",
    "l": ".-..",
    "m": "--",
    "n": "-.",
    "o": "---",
    "p": ".--.",
    "q": "--.-",
    "r": ".-.",
    "s": "...",
    "t": "-",
    "u": "..-",
    "v": "...-",
    "w": ".--",
    "x": "-..-",
    "y": "-.--",
    "z": "--..",
}
```

chall02/isoria-g.py
```diff
@@ -0,0 +1,12 @@
...
+if len(sys.argv) == 2 and (all(c.isalpha() for c in sys.argv[1])):
```
clafoutis42, line 4: No need to use comprehension here. `.isalpha()` is a string method and doing `sys.argv[1].isalpha()` would have had the same result.  
However you are forgetting to check for space character, here's how you could have done it:
```suggestion
if len(sys.argv) == 2 and sys.argv[1].replace(" ", "").isalpha():
```

chall02/isoria-g.py
```diff
@@ -0,0 +1,12 @@
...
+    string=(sys.argv[1]).lower()
```
clafoutis42, line 5: You are using redundant parentheses
```suggestion
    string = sys.argv[1].lower()
```

chall02/isoria-g.py
```diff
@@ -0,0 +1,12 @@
...
+    i = 0
+    while i <= 25:
+        string = string.replace(letter[i], morse[i])
+        i=i+1
```
clafoutis42, line 9: Using a for here would have been a more pythonic solution
```suggestion
    for i, code in enumerate(morse):
        string = string.replace(letter[i], code)
```
A further amelioration would be a comprehension using a dictionary as mentioned earlier
```suggestion
string = "".join(morse[c] for c in string)
```

chall02/isoria-g.py
```diff
@@ -0,0 +1,12 @@
...
+    print(string, end='')
```
clafoutis42, line 10: A new line is expected after the output
```suggestion
    print(string)
```

chall02/isoria-g.py
```diff
@@ -0,0 +1,12 @@
...
+else:
+    print ("usage: ./xlogin.py <a-zA-Z string>")
```
clafoutis42, line 12: Your file is named `isoria-g.py` therefore `./xlogin.py` in the login makes no sense. You have to keep in mind that the filename might change, which means that you should have done something dynamic:
```suggestion
    print ("usage: %s <a-zA-Z string>" % sys.argv[0])
```

clafoutis42 (commented):
Your code won't pass the test, and is missing the shebang to know which binary should parse it.  
Also your file is not an executable, therefore we won't merge the PR.  
But nice try! Here is [Fluent Python](https://www.amazon.com/Fluent-Python-Concise-Effective-Programming/dp/1491946008), a really good python book to help you get better with it!

clafoutis42:
As said, as we can't execute your file, we won't accept the PR.

PR #123, opened 2020-04-07
chall02

ccardozom:
This is my solution to use morse code with letters a to z 
- create a array with letters and your value on morse code
- verifycate in the standard input if only have 2 arguments, if is not true exit and show the message error
-when the input string have upcase letter, translate to lowercase and find in the array his morse code and print his value


chall02/ccardozo.py
```diff
@@ -0,0 +1,59 @@
+# **************************************************************************** #
```
anna-dominguez, line 1: A shebang is missing, it's important to use it, you can find a lot of documentation or debate about how to write it 😉

chall02/ccardozo.py
```diff
@@ -0,0 +1,59 @@
...
+#    ccardozo.py                                        :+:      :+:    :+:    #
+#                                                     +:+ +:+         +:+      #
+#    By: ccardozo <ccardozo@student.42madrid.com    +#+  +:+       +#+         #
+#                                                 +#+#+#+#+#+   +#+            #
+#    Created: 2020/04/07 02:35:33 by ccardozo          #+#    #+#              #
+#    Updated: 2020/04/07 02:35:33 by ccardozo         ###   ########.fr        #
+#                                                                              #
+# **************************************************************************** #
+
+morse_code = {  'a': '.-',
```
anna-dominguez, line 13: All you global should be written in upper case

anna-dominguez (commented):
Your file is not an executable as required in the `README` so I cannot execute it with my test. You have all you lasts challenge in you PR too...

For the logic of you program it seems good, the only thing that you could change is doing all you error handling at first, and then write your output. It would save you from declaring an extra variable

PR #124, opened 2020-04-07
...---... / .......--.-..-......-

esepunto:
The "salir" function (in Spanish, by mistake) is invoked when there is an error.
The program check that there are no errors in the data entry. 
The "MorseCode" function includes a Morse dictionary in capital letters by equivalence, and translates the message character by character after converting the characters to capital letters.
When an erroneous character is found, the "exit" function is invoked.
Finally, the translated message ("result") is printed.


aguimbao:

As stated in the README: You have to submit a python3 executable script to encode a string to morse., it should be executable as `./your_script.py` .
To achieve this, you would also need a shebang in the first line of the file like -> `#!/usr/bin/env python3`. You should document yourself about python modules, execution and shebang usage.
```python
if __name__ == "__main__":
    main()
```
- Morse dictionary should maybe be in global variable for improved readability but this is subjective as local variables save memory and are more performant.
- Also subjective but in Python it's prefered to use snake_case for variable and function/method names.
- Character method `isalpha()` is much better than checking ascii table for character sanitising.
- Your usage is badly formatted, forgot the `./`

I can't accept your PR cause of execution rights/shebang, best of luck for next challenge ;)

PR #125, opened 2020-04-07
Constelación Serpens (Chall02) by fjimenez

fjimenez81:
First I have created a dictionary with its key: value corresponding to the morse language, if the argument comes in upper case the change to lowercase, I go through the argument and if it is equal to the key, I passed the corresponding value to an aux and it was printed.

### DIFF
--- a/chall01/necklace.c
+++ b/chall01/necklace.c
@@ -1,0 +1,134 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   necklace.c                                         :+:      :+:    :+:   */
+/*                                                    +:+ +:+         +:+     */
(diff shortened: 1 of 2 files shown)


aguimbao:
First of all, you are commiting two files, for the chall01 and the chall02. For next challenge make sure you only commit to the current challenge.
Second of all, you just commited, which invalidated your PR as you can't commit after the Deadline.

- It would have been better to print char by char instead of wasting memory storing the whole string.
- Global variables should be uppercase.
- Your usage is formatted incorrectly.

It's a shame cause I like how you implemented your `main()` functionality and the rest of your functionality and your script had the correct execution rights and the right shebang, unlike most submissions in this challenge.

P.S.: Your commit messages should maybe be in english too :)

PR #132, opened 2020-04-07
Revert "estina : my version of Python problem"

aguimbao:
Reverts 42madrid/remote-challs#114


xplore93:
It is nice to read your comment. I thing that the fails are teaching us better than success. Thank you for the correction, it is important to me. GG

PR #134, opened 2020-04-09
Chall03 finished

migferna42:
**The snail function**

- BASE CASE: The first row is filled in order, and the last element is marked as filled (row_begin + 1), with the columns we do the same, fill the last column, and mark the last data as fill (col_end - 1)

- REST OF CASES: As long as they have rows available, they are completed in the same way as in the base case, but filled in by the last row. For the columns it is done in the same way, but moving from the beginning to the end.

The rest of the code, it is checked that all the input data is numerical, and that they have the same length

### DIFF
--- a/chall03/migferna.py
+++ b/chall03/migferna.py
@@ -33,7 +33,7 @@ def snail(matrix):
 
 def get_arg():
     matrix = []
-    if len(sys.argv) < 2 or (len(sys.argv) - 1 != len(sys.argv[1])):
+    if len(sys.argv) < 2 or (len(sys.argv) - 1 != len(sys.argv[1])) or len(sys.argv) > 21:wq:
         return None
     for i in sys.argv[1:]:
         if not i.isnumeric() or len(i) != len(sys.argv[1]):
--- a/chall03/migferna.py
+++ b/chall03/migferna.py
@@ -33,7 +33,7 @@ def snail(matrix):
 
 def get_arg():
     matrix = []
-    if len(sys.argv) < 2 or (len(sys.argv) - 1 != len(sys.argv[1])) or len(sys.argv) > 21:wq:
+    if len(sys.argv) < 2 or (len(sys.argv) - 1 != len(sys.argv[1])) or len(sys.argv) > 21:
         return None
     for i in sys.argv[1:]:
         if not i.isnumeric() or len(i) != len(sys.argv[1]):
--- a/chall03/migferna.py
+++ b/chall03/migferna.py
@@ -36,7 +36,7 @@ def get_arg():
     if len(sys.argv) < 2 or (len(sys.argv) - 1 != len(sys.argv[1])):
         return None
     for i in sys.argv[1:]:
-        if not i.isnumeric():
+        if not i.isnumeric() or len(i) != len(sys.argv[1]):
             return None
         matrix.append(list(i))
     return matrix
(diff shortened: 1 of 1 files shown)


chall03/migferna.py
```diff
@@ -0,0 +1,54 @@
...
+    if len(sys.argv) < 2 or (len(sys.argv) - 1 != len(sys.argv[1])) or len(sys.argv) > 21:
+        return None
+    for i in sys.argv[1:]:
+        if not i.isnumeric() or len(i) != len(sys.argv[1]):
+            return None
+        matrix.append(list(i))
+    return matrix
+
+def usage():
+    return print("usage: ./%s.py <1-9 squared_rows...>" % sys.argv[0])
```
clafoutis42, line 45: That's dynamic, but as `sys.argv[0]` is the name of the executable, you will be printing `././migferna.py.py` in your case.  
The correct way would have been this:
```suggestion
    return print("usage: %s. <1-9 squared_rows...>" % sys.argv[0])
```

chall03/migferna.py
```diff
@@ -0,0 +1,54 @@
...
+        col_begin += 1
+    return res
+
+def get_arg():
+    matrix = []
+    if len(sys.argv) < 2 or (len(sys.argv) - 1 != len(sys.argv[1])) or len(sys.argv) > 21:
+        return None
+    for i in sys.argv[1:]:
+        if not i.isnumeric() or len(i) != len(sys.argv[1]):
+            return None
```
clafoutis42, line 40: Nice use of string methods, but you still could have wrote the conditions in a better way:
```suggestion
    if len(sys.argv) < 2 and len(sys.argv) > 21:
        return None
    if not all(arg.isnumeric() and len(arg) == len(sys.argv[1:])):
        return None
```

chall03/migferna.py
```diff
@@ -0,0 +1,54 @@
...
+        if row_begin <= row_end:
+            for i in range(col_end, col_begin - 1, -1):
+                res.append(matrix[row_end][i])
+        row_end -= 1
+
+        if (col_begin <= col_end):
+            for i in range(row_end, row_begin - 1, -1):
+                res.append(matrix[i][col_begin])
+        col_begin += 1
+    return res
```
clafoutis42, line 32: This part works fine, but it looks like C code, there are a lot of python features you could have taken advantage of.

First of all you just want to wait that your arg list is empty, knowing that `not []` returns True, you could have written.
```diff
-    row_begin = 0
-    row_end = len(matrix) - 1
-    col_begin = 0
-    col_end = len(matrix[0]) - 1
-
-    while row_begin <= row_end and col_begin <= col_end:
+    while matrix:
```

You could have taken advantage of this on the whole thing, but to do so, as strings are not editable in python, you would have had to convert all the rows into arrays. A simple solution would be `array = [char for char in string]`.

Later you could have taken advantage of the `.pop` method that returns an element of the array at the given index by removing it from the said array.  
Also you could have used the following array operator `[::-1]` that returns a reversed copy of an array (or a string)...

I'll let you read the solution when it's uploaded for more details :)

clafoutis42 (commented):
Your solution resolves the problem.  
However it can be improved, also, the usage if unfortunately wrong :(

clafoutis42:
I am however really disappointed by the fact that your code is basically this:

As you are the first to submit this solution, we will discuss whether we keep it or not, however every other student having submitted this solution will see their PR refused.

PR #135, opened 2020-04-09
chall03

GabrielMartinezRodriguez:
I create a state machine with two states that alternate the direction of rotation. And at the same time it is traveling less distance.

### DIFF
--- a/chall03/gmartine.py
+++ b/chall03/gmartine.py
@@ -1,0 +1,40 @@
+import sys
+
+direction = [1, 0]
+position = [0, 0]
+repeat = 1
+flag  = 1
+
+if (len(sys.argv) == len(sys.argv[1]) + 1):
+    if (sys.argv[1].isalnum() and sys.argv[2].isalnum() and sys.argv[3].isalnum() and len(sys.argv[1]) == len(sys.argv[2]) and len(sys.argv[2]) == len(sys.argv[3])):
+        i = 0
+        j = 0
+        len = len(sys.argv[1])
+        recorrido = 0
+        while(len > 0):
+            print(sys.argv[position[1] + 1][position[0]], end = "")
+            recorrido = recorrido + 1
+            if (recorrido == len):
+                repeat = repeat + 1
+                if (repeat == 2):
+                    len = len - 1
+                    repeat = 0
+                else:
+                    flag = -flag
+                recorrido = 0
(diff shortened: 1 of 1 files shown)


chall03/gmartine.py
```diff
@@ -0,0 +1,40 @@
+import sys
```
anna-dominguez, line 1: You forgot the shebang, it's important so your OS know which parser to use to execute the file

chall03/gmartine.py
```diff
@@ -0,0 +1,40 @@
...
+
+direction = [1, 0]
+position = [0, 0]
+repeat = 1
+flag  = 1
+
+if (len(sys.argv) == len(sys.argv[1]) + 1):
+    if (sys.argv[1].isalnum() and sys.argv[2].isalnum() and sys.argv[3].isalnum() and len(sys.argv[1]) == len(sys.argv[2]) and len(sys.argv[2]) == len(sys.argv[3])):
```
anna-dominguez, line 9: You should think the other way around. and do the error conditions first with an extern function of error handling. It will simplify your code

chall03/gmartine.py
```diff
@@ -0,0 +1,40 @@
...
+        i = 0
```
anna-dominguez, line 10: Why ?

chall03/gmartine.py
```diff
@@ -0,0 +1,40 @@
...
+        j = 0
```
anna-dominguez, line 11: Why ?

chall03/gmartine.py
```diff
@@ -0,0 +1,40 @@
...
+        len = len(sys.argv[1])
+        recorrido = 0
```
anna-dominguez, line 13: Name in spanish forbidden

anna-dominguez (commented):
The shebang MUST be at the beginning of the file, or else it won't be parsed by the shell. You have to think of the rights of your file too, I didn't received an executable and cannot test it.
Not noticing this means you didn't test to execute your code using the execute syntax (`./gmartine.py`). Thus we won't merge the PR.

anna-dominguez:
You should read about Python a little in order to see the syntax of the language, your code is very similar to C. You should use more functions, and learn how to do a main in Python it's better do write one than to write everything in the root

PR #136, opened 2020-04-09
edramire - Galaxia de Bode

edramir18:
The script check all inputs are correct, before run the snail function, in case any error, the help message are displayed. 
The function snail work recursively, printing a row or a column according to the movement vector, on each iteration the boundaries of the square are reduced by 1. For example if the problem start with a square 3x3 after the first iteration will be 2x3, 2x2, 1x2 and so until become on a square 1x1.
The movement vector is rotated 90º after each iteration to generate the index of the cell to be printed.


chall03/edramire.py
```diff
@@ -0,0 +1,71 @@
+#! /usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+import sys
+
+ERROR_MSG = "usage: ./edramire.py <1-9 squared_rows...>"
```
aguimbao, line 6: A `usage()` function would have been better than storing the usage msg in a global const and just printing it whenever necessary in my opinion.

chall03/edramire.py
```diff
@@ -0,0 +1,71 @@
+#! /usr/bin/env python3
+# -*- coding: utf-8 -*-
```
aguimbao, line 2: `# -*- coding: utf-8 -*-` is a Python 2 thing. In Python 3+, the default encoding of source files is UTF-8 and that line is useless.

chall03/edramire.py
```diff
@@ -0,0 +1,71 @@
...
+            position["col"] += direction["x"]
+            snail(arr, position, limit, direction)
+            return
+
+
+def main():
+    size = len(sys.argv) - 1
+    if size == 0 or size != len(sys.argv[1]) or size > 20:
+        print(ERROR_MSG)
+        sys.exit()
```
aguimbao, line 42: You could have `sys.exit(ERROR_MSG)`, but as mentioned above, I think a usage function would have been a better approach.

chall03/edramire.py
```diff
@@ -0,0 +1,71 @@
...
+    line_size = len(sys.argv[1])
+    arr = []
+    for k in range(1, size + 1):
+        if line_size != len(sys.argv[k]) or not sys.argv[k].isdigit():
```
aguimbao, line 46: This check wouldn't take into account character '0' which wasn't allowed.

chall03/edramire.py
```diff
@@ -0,0 +1,71 @@
...
+        "top": 0,
+        "right": line_size,
+        "bottom": size
+    }
+    position = {
+        "row": 0,
+        "col": 0
+    }
+    snail(arr, position, limits, direction)
+    print("")
```
aguimbao, line 67: `print()` > `print("")`

aguimbao (commented):
Your approach was really nice to read. I hadn't thought about it when submitting the exercise.
However, python has a lot of function call overhead, which makes recursion not that much of a good approach. I think it's a really interesting approach to the exercise, but probably not the most understandable or performing, I don't even think this exercise needed an "inherently" recursive approach like other problems.

Your code looks a lot like C, you should try to use much more python builtin's and standard utilities rather than nesting if's inside while's, etc.

I liked reading your recursion though :) You can take a look at your colleagues code's for more "pythonic" approaches (The ones who didn't copy).

Execution permissions and shebang were on point, good job. Merged.

PR #137, opened 2020-04-09
chall03

GonzSanch:
First at all we check if args are valid, so we look if args are empty and if the number or args are the same length as each argument and isNumeric, if any argument fails this check function will return None else return a new list named matrix thats contains arg list[1:].

In main we use this handle error function to get args and check if fails, then return the error message.

otherwise to do the sanil sort function, I use a basic logarithm thats iterate using range(), range create a sequence of number given it three parameters start, stop and step. Then after each iteration will append the value in the snailed list. Finally using print i added a separator and called the list with *, this option remove brackets and quotes in list.


clafoutis42:

PR #138, opened 2020-04-09
chall03 - pcuadrad

pabloocg:
First I make a series of checks, that there is at least one argument, then that all the arguments are formed only by numbers and lastly that the length of each row of numbers is equal to the number of rows entered so that a square can be formed.
The algorithm that I have implemented to show the numbers of the spiral square consists of a series of steps that will be repeated for each turn of the spiral.

**First step:** save all the numbers on the first row, then the value of this row is increased so that on the next round all the numbers on the second round are taken, and so many turns of the spiral.

**Second Step:** save the number of the last column of all the rows. Then this value is decreased so that on the next round those from the previous column were taken following the same logic as in the first step

**Third step:** save all the numbers of the last row starting at the end and again the value will decrease so that the next row takes the previous row.

**Last step:** save the number of the first column of all the rows starting from the last row and reaching the first row (which will now be the second because it increased in the first step).

These steps will be repeated until all the numbers are saved and finally printed separated by commas.

### DIFF
--- a/chall03/pcuadrad.py
+++ b/chall03/pcuadrad.py
@@ -25,7 +25,7 @@ def snail(array):
         for row in range(rowEnd, rowBegin - 1, -1):
             snailArray.append(array[row][colBegin])
         colBegin += 1
-    print(', '.join(map(str, snailArray)))
+    print(', '.join(snailArray))
 
 def check(array):
     for elem in array:
@@ -37,7 +37,7 @@ def main():
     array = []
     for i in range(1, len(sys.argv)):
         array.append(sys.argv[i])
-    if len(array) == 0 or check(array) == False:
+    if len(array) == 0 or len(array) > 20 or check(array) == False:
         error()
     snail(array)
 
--- a/chall03/pcuadrad.py
+++ b/chall03/pcuadrad.py
@@ -1,0 +1,45 @@
+#!/usr/bin/env python
+
+import sys
+
+def error():
+    print ("usage: " + sys.argv[0] + " <1-9 squared_rows...>")
+    exit(1)
+
+def snail(array):
+    snailArray = []
+    rowBegin = 0
+    rowEnd = len(array) - 1
+    colBegin = 0
+    colEnd = len(array[0]) - 1
+    while rowBegin <= rowEnd and colBegin <= colEnd:
+        for col in range(colBegin, colEnd + 1):
+            snailArray.append(array[rowBegin][col])
+        rowBegin += 1
+        for row in range(rowBegin, rowEnd + 1):
+            snailArray.append(array[row][colEnd])
+        colEnd -= 1
+        for col in range(colEnd, colBegin - 1, -1):
+            snailArray.append(array[rowEnd][col])
+        rowEnd -= 1
+        for row in range(rowEnd, rowBegin - 1, -1):
+            snailArray.append(array[row][colBegin])
+        colBegin += 1
+    print(', '.join(map(str, snailArray)))
+
+def check(array):
+    for elem in array:
+        if len(elem) != len(array) or elem.isdigit() == False:
+            return False
+    return True
+
+def main():
+    array = []
+    for i in range(1, len(sys.argv)):
+        array.append(sys.argv[i])
+    if len(array) == 0 or check(array) == False:
+        error()
+    snail(array)
+
+if __name__ == "__main__":
+    main()


chall03/pcuadrad.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env python
```
aguimbao, line 1: `#!/usr/bin/env python3` > `#!/usr/bin/env python` for this challenge.

chall03/pcuadrad.py
```diff
@@ -0,0 +1,45 @@
...
+
+import sys
+
+def error():
```
aguimbao, line 5: I think "usage" would have been a better name than "error" for this function.

chall03/pcuadrad.py
```diff
@@ -0,0 +1,45 @@
...
+    print ("usage: " + sys.argv[0] + " <1-9 squared_rows...>")
+    exit(1)
```
aguimbao, line 7: `sys.exit()` > `exit()`.

And you could have done `sys.exit("usage: " + sys.argv[0] + " <1-9 squared_rows...>")`.

chall03/pcuadrad.py
```diff
@@ -0,0 +1,45 @@
...
+            snailArray.append(array[rowEnd][col])
+        rowEnd -= 1
+        for row in range(rowEnd, rowBegin - 1, -1):
+            snailArray.append(array[row][colBegin])
+        colBegin += 1
+    print(', '.join(snailArray))
+
+def check(array):
+    for elem in array:
+        if len(elem) != len(array) or elem.isdigit() == False:
```
aguimbao, line 32: Comparison to False should be 'if conf is False:' or 'if not cond:'

chall03/pcuadrad.py
```diff
@@ -0,0 +1,45 @@
...
+            return False
+    return True
+
+def main():
+    array = []
+    for i in range(1, len(sys.argv)):
+        array.append(sys.argv[i])
+    if len(array) == 0 or len(array) > 20 or check(array) == False:
```
aguimbao, line 40: Comparison to False should be 'if conf is False:' or 'if not cond:'

chall03/pcuadrad.py
```diff
@@ -0,0 +1,45 @@
...
+        for row in range(rowBegin, rowEnd + 1):
+            snailArray.append(array[row][colEnd])
+        colEnd -= 1
+        for col in range(colEnd, colBegin - 1, -1):
+            snailArray.append(array[rowEnd][col])
+        rowEnd -= 1
+        for row in range(rowEnd, rowBegin - 1, -1):
+            snailArray.append(array[row][colBegin])
+        colBegin += 1
+    print(', '.join(snailArray))
```
aguimbao, line 28: If you have already sanitised the input, why do you store all the array in memory instead of printing it while you traverse the whole array, saving memory in the process?

aguimbao (commented):
Nice approach :) Liked your explanations. Good job on the execution permissions and the shebang !

(But I prefer snake_case to camelCase 😁)

chall03/pcuadrad.py
pabloocg: For me that name makes more sense because it is like an exception to the normal flow of the program, but it is worth thanks for the note.

chall03/pcuadrad.py
pabloocg: Okay, I'll keep it in mind thanks!

chall03/pcuadrad.py
pabloocg: Ok thanks, I still don't know the correct python syntax

chall03/pcuadrad.py
pabloocg: For me it is cleaner to do a single print at the end of once, but it is true that it is less efficient, I will take it into account thanks.

chall03/pcuadrad.py
aguimbao: It definitely was a rather subjective observation from me.

chall03/pcuadrad.py
aguimbao: Don't worry. If you're interested in fixing this kind of errors there's really good python linters. `Black`, `Flake8` and `isort` (for imports).

chall03/pcuadrad.py
aguimbao: It's cleaner for me too, and in case there was an error external to the program, you wouldn't have printed anything yet and it would be better. But still, I think printing through the process is more performant.

PR #140, opened 2020-04-09
Chall03 cdorta-n

Top64:
Hello, in this activity, unlike the last one, I created a main() function and an error function to improve the code's quality. With the help of two other friends of 42, we figure out how to create a bidimensional variable and fill it with the arguments.

First check  
> A valid square can range from a size of 1*1 to a size of 20*20.

Then we create the Matrix, see if every line has the same size. Then we fill it if the character is a digit.

The spiral_traversal y a function that creates a string in the snail/spiral clockwise order. and finally we print it,


Top64:
**I'm working on Windows and I tried to create the Executable script but I couldn't find a way ** 
😢

clafoutis42:

PR #141, opened 2020-04-09
Callenge 3

victorsempere:
Main will receive the array of numbers and pass it on to a snail function processed as an array of arrays.
Inside if the arrays are ok. I build a list with all the arrays.
And then I put points where the direction will change and in that point an offset will be updated. The offset will move the pointer to the right character in the list to be printed.
We will be printing until all the characters has been printed.


chall03/vsempere.py
```diff
@@ -0,0 +1,91 @@
...
+    if direction == 0:
+        return 1;
+    if direction == 1:
+        return board_size;
+    if direction == 2:
+        return -1;
+    return -board_size;
+
+def buildArray(argv):
+    if (len(argv)<=1 or len(argv)>=10):
```
aguimbao, line 17: So you didn't understand correctly. We specified `NOTE: A valid square can range from a size of 1*1 to a size of 20*20.`.
The part in the usage that says `<1-9 squared_rows...>` means that the only valid characters are ascii 1 to 9. 

(Even if you understood that the biggest valid test was 9\*9, your script still prints usage in a 9\*9 instead of solving it).

chall03/vsempere.py
```diff
@@ -0,0 +1,91 @@
...
+        return None;
+
+    board = [];
+    i = 1;
+    row_len = len(argv[i]);
+    while (i <len(argv)):
```
aguimbao, line 23: This could have been better written as `for elem in argv[1:]` to be more "pythonic", try to use less indexes and more builtin methods.

chall03/vsempere.py
```diff
@@ -0,0 +1,91 @@
...
+        if (row_len != len(argv[i])):
+            return None;
+        for e in argv[i]:
+            if not e.isnumeric():
```
aguimbao, line 27: I find `isdigit()` more reliable than `.isnumeric()`. It can be done directly on the string rather than in each character btw.

chall03/vsempere.py
```diff
@@ -0,0 +1,91 @@
...
+    print('usage: ./vsempere.py <1-9 squared_rows...>');
+
+def getOffset(direction, board_size):
+    if direction == 0:
+        return 1;
+    if direction == 1:
+        return board_size;
+    if direction == 2:
+        return -1;
+    return -board_size;
```
aguimbao, line 14: This could use some `elif's` or a dictionary.

chall03/vsempere.py
```diff
@@ -0,0 +1,91 @@
...
+    if board is None or len(board)<1 or len(board)>9:
+        return 0;
+    row_len = len(board[0]);
+    rows = 0;
+    for row in board:
+        if (row_len != len(row)):
+            return 0;
+        for cell in row:
+            if (not cell.isnumeric()):
+                return 0;
```
aguimbao, line 54: I think you already made these checks in `buildArray()`

aguimbao (commented):
In python it's not recommended to use `;` for line termination, like in JS. 
It's possible to use it, but the only usefulness it can have is multi-line statements, which is not recommended by PEP8.

Also, I think snake_case is easier to read for function and variable naming, but that's a bit subjective.
It's a shame that you didn't understand the usage provided in the examples :(.

Good job on the execution permissions and the shebang though, merged.

PR #142, opened 2020-04-09
chall03 - fulldemo

fulldemo13:
Hi I'm trying to do my code more "phyntonic".
First I put "argv" in my string "array", with this I "delete" the program name,  and check if all values are numbers and if array is a square string. If it's not I return the error message.
When all is fine I start function "looper()". Here I set variables "low" and "high" that represents the top corners of our square string. I save the values square by square making more little this square incrementing "low" an decrementing "high". I know the number of squares with count, for example in 3x3 array i will have 2 squares.Array length + 1 divided 2.
I use four loops, first increments the positions in the top of the square, second decrements right side, third decrements the bottom and the fourth the left side of the square. In every loop I save the value in "res".
Finally I print res by inserting commas and spaces.

### DIFF
--- a/chall03/fulldemo.py
+++ b/chall03/fulldemo.py
@@ -1,0 +1,45 @@
+#! /usr/local/bin/python3
+
+import sys
+
+def error():
+    print ("usage: " + sys.argv[0] + " <1-9 squared_rows...>")
+    exit(1)
+
+def looper(array):
+	num = len(array)
+	res = []
+	low = 0
+	high = num - 1
+	count = int((num + 1) / 2) 
+	for i in range(count):
+		for j in range(low, high + 1):
+			res += array[i][j]
+		for j in range(low + 1, high + 1):
+			res += array[j][high]
+		for j in range(high - 1, low - 1, -1):
+			res += array[high][j]
+		for j in range(high - 1, low, -1):
+			res += array[j][low]
+		low = low + 1
+		high = high - 1
+	print(', '.join(res))
+
+def checker(array):
+	if len(sys.argv) < 2 or len(sys.argv) > 21:
+		error()
+	for i in array:
+		if not i.isdigit():
+			error()
+		if len(i) != len(array):
+			error()
+
+def main():
+	array = []
+	for i in range(1, len(sys.argv)):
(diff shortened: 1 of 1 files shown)


chall03/fulldemo.py
```diff
@@ -0,0 +1,45 @@
...
+def checker(array):
+	if len(sys.argv) < 2 or len(sys.argv) > 21:
+		error()
+	for i in array:
+		if not i.isdigit():
+			error()
+		if len(i) != len(array):
+			error()
+
+def main():
```
anna-dominguez, line 37: Good 😍

chall03/fulldemo.py
```diff
@@ -0,0 +1,45 @@
...
+
+import sys
+
+def error():
+    print ("usage: " + sys.argv[0] + " <1-9 squared_rows...>")
+    exit(1)
+
+def looper(array):
+	num = len(array)
+	res = []
```
anna-dominguez, line 11: If you've done the error handling before your solution function you don't need to initialize a variable and could print one by one the result found

chall03/fulldemo.py
```diff
@@ -0,0 +1,45 @@
+#! /usr/local/bin/python3
+
+import sys
+
+def error():
+    print ("usage: " + sys.argv[0] + " <1-9 squared_rows...>")
+    exit(1)
+
+def looper(array):
```
anna-dominguez, line 9: This is not a good function name, insufficiently explanatory

chall03/fulldemo.py
```diff
@@ -0,0 +1,45 @@
+#! /usr/local/bin/python3
```
anna-dominguez, line 1: Good 👍 !

chall03/fulldemo.py
```diff
@@ -0,0 +1,45 @@
...
+
+import sys
+
+def error():
```
anna-dominguez, line 5: Good 👍 !

anna-dominguez (commented):
A lot of good things ! Your code is working, you've done a main, didn't forgot the shebang and your file is executable. Good job !

PR #144, opened 2020-04-09
fgata-va's solution

FranciscoJesusGata:
# How do I solved it
First of all I check that the input is like the rules of the program dictates. Then I get the input I need and split it into single numbers. To sort the numbers create a loop with that iterates over the 4 movements of the snail sort in order (right, down, left, up). For every number I save, I delete it to delimit the matrix. And for every move I check if I sorted all the numbers (i < (side²))
# How my code works
As the previous program I created the function errorMessage, with a string formating with the file name, another function to validates the input and the snailSorting function. In the snailSorting function I declare the list to save the sorted data and the list containing the matrix to sort. Then I iterate over the 4 moves checking if the number saved are equal to lenght². Finally I print the list with the ' ,' separator and without the brackets.


chall03/fgata-va.py
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/python3
```
anna-dominguez, line 1: Good 👍

chall03/fgata-va.py
```diff
@@ -0,0 +1,59 @@
...
+            break
+        # Go left
+        for x in range(x, -1, -1):
+            snailSort.append(matrix[y][x])
+            i += 1
+        del matrix[y]
+        if(i >= maxlen):
+            break
+        # Go up
+        for y in range(len(matrix)-1, -1, -1):
```
anna-dominguez, line 46: I think you could simplify your code with all those almost duplicated `for`

anna-dominguez (commented):
🥳 your code is working, executable, with a shebang, good job !
The best advice I could give you is to do a main. It doesn't matter how simple the script is, you may want to import it's contents without executing the file one day so it's good practice to implement a `main()` function and set the appropriate:
```python
if __name__ == "__main__":
    main()```

FranciscoJesusGata:
Thanks for the advices, I'll try to use more functions and lists to avoid the repeats loop, and I'm taking note about the _main_ statement

PR #145, opened 2020-04-09
mpena-gu

mepg72:


mepg72:
quitando Numpy

mepg72 (commented):
del Numpy library

mepg72:
del Numpy library

chall03/mpena-gu.py
```diff
@@ -0,0 +1,84 @@
...
+            for col in reversed(range(col_init, col_end)):
+                snail_result.append(squared_rows[row][col])
+        else:
+             snail_result.append(squared_rows[row][col_end-1])
+    for row in reversed(range (row_init + 1, row_end - 1)):
+        snail_result.append(squared_rows[row][col_init])
+    snail_str = ", ".join(snail_result)
+    return snail_str
+
+if __name__ == "__main__":
```
anna-dominguez, line 61: It's a good behaviour to write a condition for your main, but you should do a function `main` called in this condition and not write your main directly in it. Like this :
```python
if __name__ == "__main__":
    main()
```

chall03/mpena-gu.py
```diff
@@ -0,0 +1,84 @@
...
+    if len(sys.argv) < 2 or not check_argv(sys.argv[1:]):
+        print ("usage: ./mpena-gu.py <1-9 squared_rows...>")
```
anna-dominguez, line 63: You could do an external function to avoid duplication of code

anna-dominguez (commented):
Your code is almost working, I can see that you're writing your error usage when I write a simple argument like "1", but it should work with it too.

chall03/mpena-gu.py
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/python
```
anna-dominguez, line 1: Your shebang must look like `#!/usr/bin/env python3`

mepg72:
Thanks vesta-nna for you correction, I'll solve all topics and deliver again.

PR #147, opened 2020-04-09
snails

jdiaz-co:
First I see if the arguments are valid, if they are not, I call the function error ().

If they are valid we look at the size of each argument (x) and the total size of argv (y), the latter we subtract 1 to remove the file from that size.

I create the array, then if x is different from y I call the function error () because the arguments are not correct.

Then I see if there is something in the arguments that is not a number, if there is, I go to the error () function

Then I save the arguments in the array.

I call the spiral_traversal function directly in the print function formatting it.

In this function it is where I go through the matrix in the order that the exercise asks me with various loops and conditions.


clafoutis42:

PR #148, opened 2020-04-09
0000

SanGilazana:
I add .txt whit my code comment


clafoutis42:

SanGilazana:
I worked with two other colleagues to get the exercise out. You can ask them, though I'll stick to what I've learned. Thanks for asking, but I didn't know about the e
xistence of that code. We've gotten information from other sites, but not from Github.
Besides, you have compared us with different results when the three of us have practically the same, since we have taken it out among the three of us. I think it's unfair that other colleagues have it too and you've taken it for granted. However, there is no indication in the rules. This way you only get us to do no more. HUGS!

PR #149, opened 2020-04-09
Snail Challenge

jtrancoso:
First I define `usage` with the error message and the exit. Then I create `fail_check` to test the lenght of the argvs and if they create a square array.

After that it's time for the snail sorting: 

1. Find each corner of the square and create a container array where we will append the numbers
2. Copy the first row
3. Append the last column
4. Copy the last row inversed
5. Copy the first column from bottom to top
6. Keep repeating making the spiral smaller each time until we've finished

And finally, I check if everything inside the array are numbers.

### DIFF
--- a/chall03/jtrancos.py
+++ b/chall03/jtrancos.py
@@ -1,0 +1,61 @@
+# **************************************************************************** #
+#                                                                              #
+#                                                         :::      ::::::::    #
+#    jtrancos.py                                        :+:      :+:    :+:    #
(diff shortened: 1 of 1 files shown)


jtrancoso:
Moved numeric check into `fail_check` function to improve order

clafoutis42:

jtrancoso:
Ok I understand and apologize, it seems we all got some heavy inspiration :( I'll look into the reviewed code and will do better next time, thank you!

PR #152, opened 2020-04-09
Chall03 solution fvega-tr

fvega-tr:


chall03/fvega-tr.py
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env python3
```
anna-dominguez, line 1: Good

chall03/fvega-tr.py
```diff
@@ -0,0 +1,87 @@
...
+					
+					else:
+						break
+
+		print(draw)
+
+	else:
+		print("usage: %s <1-9 squared_rows...>" % sys.argv[0])
+
+except:
```
anna-dominguez, line 86: Bad

chall03/fvega-tr.py
```diff
@@ -0,0 +1,87 @@
...
+	print("usage: %s <1-9 squared_rows...>" % sys.argv[0])
```
anna-dominguez, line 87: Worst, you have your error usage in your `try` and in your `exception`

chall03/fvega-tr.py
anna-dominguez: A `try`/`catch` must be done for one specific thing, not for all you program. It's a very bad habit to surround your code with those.

anna-dominguez (commented):
You didn't forgot the shebang but we asked for an executable this time too, you didn't set the rights of your file and we cannot test it. You should read the `README` of the challenge

anna-dominguez:
Not an executable and no comment of your work 😞

fvega-tr:
Ok, I will try to improve for next time, thank you very much for the
correction.
I thought about the permissions later, I did not remember that I was
working on windows.

PR #153, opened 2020-04-09
chall 03

miyomovil:
I know its wrong but i tried! :)
First of all i decided to filter the correct cases. Then I decided to make an iteration count and emulate
snail movements using rmax imax rmin and rmax but that was a clear bad strategy. It sorta works for 2x2 and 3x3 but not for the rest of the cases :(

### DIFF
--- a/chall03/antomart.py
+++ b/chall03/antomart.py
@@ -1,0 +1,85 @@
+#!/usr/bin/env python3
+
+import sys
+import re
+
+def main():
+    n_args = len(sys.argv)
+    args_val = sys.argv
+    args_val.remove(sys.argv[0])
+    pattern = re.compile("[1-9]+")
+    j = 0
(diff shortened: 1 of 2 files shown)


chall03/antomart.py
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env python3
```
anna-dominguez, line 1: Good 👏

anna-dominguez (commented):
Too bad that your file isn't an executable 😞 You've done your code as if it was C language. I strongly recommend you to read about Python and good habits. The fact that you've done a `main` is a good start !
I think that you have a lot of redundant things (error handling for example, you should start with it and avoid a  lots of conditions). It's not C and there is no norme but doing functions would help your to make your code more readable and understandable. You're on the right track, now all you have to do is practice.

PR #154, opened 2020-04-09
Snail the matrix - Chall03

JavierCervilla:
first of all, sorry for my english.
1. handling errors:
     no args, args <= 20 and >=1
     check the matrix to be square
     check that there are not any  '0'
2. create the matrix with two for iterations, the first one for the rows and in each row a new list for the columns.
3. when we have the matrix we do a snail function which do the next:
note: we know that there is only 4 posible movement left->right, up->down, right->left, down->up. We also know that we need 4 counters, col_start(+), col_end(-), row_start(+), row_end(-). the sign (-) is because it decreases and (+)because it increment when we print a new row or a new col. We have to adjust the ending depending if we need ", " or the "\n" that prints print() by default. 
     3.1.-first movement (left->right) we print the col_start->col_end and increases row_start
     3.2.-second movement(up->down) we print the row_start->row_end and decreases col_end 
     3.3.-third movement(right->left) we print the col_end->col->start and decreases row_end
     3.4.-fourth movement(down->up) we print the row_end->row_start and increases col_start
thanks Staff you are the best but this is the most difficult chall at the moment and is yours. :)
collaborator: @mmiguelx  by call.


clafoutis42:

JavierCervilla:
I took 6 tours doing that Code, i dont copy It you can see my commits and
verify It.  Imagen verte disapointed With your comentary

clafoutis42:
Please, you and 10 others have this code. Even though you added this "ending_function" logic to adapt it to the subject, [this commit](https://github.com/42madrid/remote-challs/pull/154/commits/384e118ec0f5a46bbe60d16a3c111bdf96069d8d) really gives it up, it corresponds exactly to the code I found on github.

JavierCervilla:
this was the reference, not from github: https://ourcodeworld.com/articles/read/827/how-to-format-a-given-array-matrix-in-spiral-form-snail-or-clockwise-spiral-sorting-in-python
and you can see that is not the same code, is more pythonic (than the first solution you send me to read the commentary which has copy paste and you accept it(same variables name, and first solve...) like the saying says.. if smells like shit and seems to be shit do not eat it, its shit... :) ### **but im very happy because we learn a lot yesterday doing it so 0 frustration**

PR #155, opened 2020-04-09
isoria-g chall03 snail

nsoria042:
First we check that the parameters passed meet the conditions. 
1. That there is more than one parameter. 
2. That's a square. 
3. That all parameters have the length of the first. 
4. That all characters are numbers. 
If everything is correct we proceed first to create the list composed of the different rows (each argument), and then call the function that creates a new list from the previous one, where we spiral through the matrix from (0.0) to (0.n), from (0.n) to (n,n), from (n,n) to (n.0), from (n.0) to (1.0) and so on until we reach the last coordinate. 
And finally we print it on the screen, separated by commas.


clafoutis42:

PR #156, opened 2020-04-09
agutierr

obispowned:
chall3


clafoutis42:
You uploaded the solution of monday's challenge, and you din't explain your code :/

PR #157, opened 2020-04-09
mpernia-

mmiguelx:
Hey, hey. Hello!

This one was a little harder than before.

1. Check if the arguments are between 1 and 20 and all of then have the same length than number of arguments so we get a perfect square, if not call error() (Function to close the program with a error message)
2. Create a matrix by adding all of the numbers one by one, if there's a 0 call error()
3. Start counters to row_start and col_start in 0 and row_end and col_end with the length of arguments - 1, because matrix index starts in 0 :o
4. While the start counter are equal or lesser than the end ones check the ending of the print() to allow you to use a more fluid print in every line (Thank you, @clafoutis42). Knowing this in every print -1 to end counters and +1 to start counters respectively.

It was a team effort this code: Thank you, @JavierCervilla.


clafoutis42:

You are also the second to add this `ending_def` function.

mmiguelx:
Hello, @clafoutis42. I didn't know it was a github with the solution and I'm not planning to look at it right now, I use https://ourcodeworld.com/articles/read/827/how-to-format-a-given-array-matrix-in-spiral-form-snail-or-clockwise-spiral-sorting-in-python as inspiration and making a twist with the previous reviews you made me. However, I'm not here for nothing more than learning and my pairs know that I'm not the copy code guy. I was the second one to have a function to define because as I said it was a team effort. Thanks for the feedback, looking forward for more challenges >:D.

I'll try to get something for the PR you mentioned before.

PR #158, opened 2020-04-09
chall03 - gbudau

gbudau:
**How i solved this problem:**
First manage all error cases.

Then i'm printing the argv array(ignoring program name), clockwise starting from outside, separating the digits by a comma.

**Short description of the code:**
Check for valid input: between 1 and 20 arguments, excluding program name, all arguments have the same length(the number of argument count) and all arguments contain only ASCII digits.

Then using `row` as top limit and `start` as right limit and `count` as left and bottom limits print the digits in the square in clockwise order.
Increment `row` and `start` and decrement `counter`
Keep printing until `row` is equal to `counter`

Print first `row` separated by comma or space(on last 1x1 square), increase `row` counter.
Print all digits on the left side, separated by comma.
Print all digits on the bottom side in reverse order separated by comma or space(on last 2x2 square)
and decrement `count`.
Print all digits on the right side separated by comma.
Keep incrementing `row` and `start`, and decrementing `count` until `row == count` meaning we reach the last inner square.
Using slicing  to manage limits and the order of printing.

Nice challenge!

### DIFF
--- a/chall03/gbudau.py
+++ b/chall03/gbudau.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python3
+import sys
+import curses.ascii
+
+def usage():
+    sys.exit(f"usage: {sys.argv[0]} <1-9 squared_rows...>")
+
+# Check if argument count is between 1 and 20
+def check_argc(count):
+    return count > 0 and count < 21
+
+# Check if string has the same length as count and is made of ascii digits
+def isalphanum_len(string, count):
+    length = 0
+    for letter in string:
+        if not curses.ascii.isdigit(letter) or length > count:
+            return False
+        length += 1
+    return length == count
+
+# Validate all arguments
+def check_each_arg(args, count):
+    for arg in args:
+        if not isalphanum_len(arg, count):
+            return False
+    return True
+
+# Error checking: valid number of command-line arguments, format and length of arguments
+def check_errors(args, count):
+    if not check_argc(count) or not check_each_arg(args, count):
+        usage()
+
+# Print digits in the matrix starting from outside, clockwise
+# Using slicing to manage length or order of printing
+# Increment rows and start of square and decrement count
+def print_snail(args, count):
+    start = 0
+    row = 0
+    while row < count:
+        for digit in args[row][start:count:]:
+            print(digit, end=', ') if row + 1 != count else print(digit)
+        row += 1
+        if row == count:
+            break
+        for arg in args[row:count - 1]:
+            print(arg[count - 1], end=', ')
+        for digit in args[count - 1][start + 1:count:][::-1]:
+            print(digit, end=', ')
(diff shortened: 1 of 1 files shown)


chall03/gbudau.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env python3
+import sys
+import curses.ascii
```
aguimbao, line 3: I like this. I think for this specific case `curses.ascii.isdigit()` > regular `.isdigit()`

chall03/gbudau.py
```diff
@@ -0,0 +1,61 @@
...
+
+def usage():
+    sys.exit(f"usage: {sys.argv[0]} <1-9 squared_rows...>")
```
aguimbao, line 6: I will tell you why this f-string formatting wasn't a good idea from my point of view. Your shebang specifies `python3`.
Sadly, on the system I develop (Debian 10), I have my `python3.7.7` in `/usr/local/bin/` which I regularly develop in, along with many other python versions. For one, I have a `python3.5.3` in `/usr/local/bin`. Which is the first python interpreter that is found in my environment when you use your shebang.

f-formatting was introduced in `python3.6`, so your script crashes on my system. In my opinion, you should have specified `python3.6` or `python3.7` in your shebang to make sure it gets executed correctly in "every system".

We specified we would use `python3.x`. So your program should work on any version of `python3`.

aguimbao (commented):
I don't have much to say about your program. I think it's pretty well written. I like snake_case, I like your comments, I like how it's organised and your usage of python.

For the shebang and the f-formatting, we'll see if it works in the environment where we run our tests ;)

chall03/gbudau.py
```diff
@@ -0,0 +1,61 @@
...
+    for letter in string:
+        if not curses.ascii.isdigit(letter) or length > count:
+            return False
+        length += 1
+    return length == count
+
+# Validate all arguments
+def check_each_arg(args, count):
+    for arg in args:
+        if not isalphanum_len(arg, count):
```
aguimbao, line 24: Instead of using the function `isalphanum_len()` and moving `count` around, you can:

```python
if not all (len(v) == len(args[0]) and v.isdigit() for v in args):
   return False
```

gbudau:
Thank you a lot for the feedback, very good points about the string formating and version of python,  i didn't realize it could cause backward compatibility problems and didn't knew it is posible to use python 3.x  to specify the version of python 3, learned a new thing.

chall03/gbudau.py
gbudau: Your solution looks really good and a lot more pythonic, i'm gonna try to implement this style in future assignments.

chall03/gbudau.py
gbudau: Really good points! Thank you!

PR #161, opened 2020-04-10
Space Oditty ssacrist

esepunto:
First, the program checks that the string of arguments complies with the request. 

Then a 2D-matrix is created where each row is saved as an array. 

An algorithm is applied to this array: it prints the elements one by one and in the correct order. I think this is more efficient.

Finally, the program prints ('') to close the string.

**Algorithm:**
First print([0],[0]. Then the rest of elements follow this sequence:
[0, +] * (total rows - 1)    => [0,1] [0,2] [0,3] [0,4] (in case of 5x5), plus [0,5] in case of 6x6, etc.
[+, 0] * (total rows - 1)    => [1,4] [2,4] [3,4] [4,4] 
[0, - ] * (total rows - 1)    => [4,3] [4,2] [4,1] [4,0]
[ -, 0] * (total rows - 2)    => [3,0] [2,0] [1,0]  
And the sequience starts again:
[0, +] * (total rows - 2)    => [1,1] [1,2] [1,3]
[+, 0] * (total rows - 3)    => ...
[0, - ] * (total rows - 3)
...Etc. It runs while (total rows - x) > 0.


esepunto:
First, the program checks that the chain of arguments complies with the request.

Then a matrix is created where each row is saved as an array.

An algorithm is applied to this array: it prints the elements one by one and in the correct order.

Finally, the program prints ('') to close the string.

Algorithm: 
First print([0],[0], and the rest of elements follow this sequence:
[0, +] * total rows => [0,1] [0,2] [0,3] [0,4] (in case of 5x5)
[+, 0] * total rows
[0, -] * total rows-1
[-, 0] * total rows-1
And again:
[0, +] * total rows-2
[+, 0] * total rows-2

esepunto:
I have a dude about coding. 

Which is preferable: that the code is better understood by a human or that the code is as efficient as possible for the compiler?

I came up with two possibilities for the check function, but I don't know which one is better. 

Possibility 1: in the code add. 

Possibility 2 (imho more difficult to understand to humans, but it's 3 lines short):
```
def check():
    if len(sys.argv) < 2 or len(sys.argv) > 21:
        byebye()
    args = len(sys.argv) - 1
    for i in range(1, len(sys.argv)):
        if args != len(sys.argv[i]) or (sys.argv[i]).isdigit() == False:
            byebye() 
```
What do you think about it?
Thanks!!

chall03/ssacrist.py
```diff
@@ -0,0 +1,49 @@
...
+
+def check():
+    if len(sys.argv) < 2 or len(sys.argv) > 21:
+        byebye()
+    sides = (len(sys.argv)) - 1
+    long = (len(sys.argv[1]))
+    if sides != long:
+        byebye()
+    for i in range(2, sides+1):
+        if len(sys.argv[i]) != long or (sys.argv[i]).isdigit() == False:
```
aguimbao, line 17: Comparison to False should be 'if cond is False:' or 'if not cond:'

chall03/ssacrist.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python3
+
+import sys
+
+def byebye():
```
aguimbao, line 5: You need to use more "serious" naming for functions and variables :)

chall03/ssacrist.py
```diff
@@ -0,0 +1,49 @@
...
+            print(", %s" %(matrix[i][j]), end='')
+        for contador in range (c, 0, -1): 
+            j -= 1
+            print(", %s" %(matrix[i][j]), end='')
+        c -= 1
+        for contador in range (c, 0, -1):
+            i -= 1
+            print(", %s" %(matrix[i][j]), end='')
+        h = 0
+    print('')
```
aguimbao, line 46: This looks too much like C. You should take a look at your colleague's solutions (the ones who didn't cheat) for a more "pythonic" approach 👍

chall03/ssacrist.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python3
+
+import sys
+
+def byebye():
+    print("usage: %s <1-9 squared_rows...>" % sys.argv[0])
+    sys.exit()
+
+def check():
+    if len(sys.argv) < 2 or len(sys.argv) > 21:
```
aguimbao, line 10: This approach is better:

```python
if (len(sys.argv) < 2 or len(sys.argv) > 21) or (not all (len(v) == len(sys.argv[1]) and v.isdigit() for v in sys.argv[1:])):
    byebye()
```
`check()` in 2 lines :)

aguimbao (commented):
I can't merge the PR.

You forgot to `chmod +x` your script (644 permissions), so even if your shebang is present, it couldn't be executed with `./ssacrist.py`.

aguimbao:
> I have a dude about coding.
> 
> Which is preferable: that the code is better understood by a human or that the code is as efficient as possible for the compiler?
> 
> I came up with two possibilities for the check function, but I don't know which one is better.
> 
> Possibility 1: in the code add.
> 
> Possibility 2 (imho more difficult to understand to humans, but it's 3 lines short):
> 
> ```
> def check():
>     if len(sys.argv) < 2 or len(sys.argv) > 21:
>         byebye()
>     args = len(sys.argv) - 1
>     for i in range(1, len(sys.argv)):
>         if args != len(sys.argv[i]) or (sys.argv[i]).isdigit() == False:
>             byebye() 
> ```
> 
> What do you think about it?
> Thanks!!

About this, usually it's better to find a short, but understandable way to do things. You should always look for the most performing way to do things, but it's also VERY important that other developers can understand your code easily, but you also need to take advantage of the capabilities of your language to make it as short and concise as possible in my opinion. There has to be a balance. 

The example of the `check()` function that I wrote for you is short, but it's not that easily understandable, it would need a comment. It's better when the code is `self-commented` (it doesn't need comments to be understood).

chall03/ssacrist.py
esepunto: I don't understand at all. Before push my file, I type  `-ls -la` to check the permissions and I see -rwxrwxrwx.  Is it necesary to do any other action??

chall03/ssacrist.py
esepunto: Well, I undestand you and maybe my level poor english is not a help to me, but... I code as I am. But I'll keep that in mind. Thanks!

chall03/ssacrist.py
aguimbao: Maybe you were in another branch?

You are submitting your PR from your fork:master-branch. And when I clone your repo and I check the master branch, the permissions in your file are `-rw-r--r--` (644) :(

chall03/ssacrist.py
aguimbao: I didn't mean this in a bad way, when I was a student I used to use funny variable and function names too all the time. But the day you'll be in a company, people won't like that, it's better to have descriptive and serious variable and function names and comments.

PR #162, opened 2020-04-10
chall03

bmk101:


clafoutis42:

Furthermore you didn't even try to explain the solution...

PR #163, opened 2020-04-10
not working

mcarrete42:
AAAhhh!! 
I didnt have time to finish the big while. I started doing everything with whiles, and then changed to for and not working,,,

### DIFF
--- a/chall03/mcarrete.py
+++ b/chall03/mcarrete.py
@@ -1,0 +1,42 @@
+#!/usr/bin/env python
+
+import sys
+
+def snailArray(argv):
+	i = 0
+	top_side = len(argv) - 1
+	side = len(argv) - 1
+	j = 1
+	while i <= top_side and j <= side:
+		for i in range(j - 1, side):
+			print(argv[j][i] + ",", end = " ")
+		i +=  1
+		for k in range(i, top_side):
+			print(argv[k + 1][side] + ",", end = " ")
+		j -= 1
+		if i <= top_side:
+			for k in range(side, j):
+				print(argv[top_side][k] + ",", end = " ")
+		top_side -= 1
+		if j <= side:
+			for k in range(top_side, i - 1, -1):
+				print(argv[k][j] + ",", end = " ")
+		j += 1
+
+
+def main(argv):
+	i = 1
+	error = "usage: ./mcarrete.py <1-9 squared_rows...>"
+	if len(argv) < 2:
+		print(error)
+		return(error)
+	while i < len(argv):
+		if len(argv[i]) != (len(argv) - 1) or argv[i].isdigit() == False:
+			print (error)
+			return error
+		i = i + 1
+	snailArray(argv)
+
+if __name__ == '__main__':
+	main(sys.argv)
+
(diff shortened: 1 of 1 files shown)


chall03/mcarrete.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python
```
aguimbao, line 1: `#!/usr/bin/env python3` > `#!/usr/bin/env python`

chall03/mcarrete.py
```diff
@@ -0,0 +1,42 @@
...
+		top_side -= 1
+		if j <= side:
+			for k in range(top_side, i - 1, -1):
+				print(argv[k][j] + ",", end = " ")
+		j += 1
+
+
+def main(argv):
+	i = 1
+	error = "usage: ./mcarrete.py <1-9 squared_rows...>"
```
aguimbao, line 37: It's better to define a `usage()` function and call it rather than printing "usage: ./mcarrete.py <1-9 squared_rows...>" like you do in lines 31 and 35.

chall03/mcarrete.py
```diff
@@ -0,0 +1,42 @@
...
+			print(argv[k + 1][side] + ",", end = " ")
+		j -= 1
+		if i <= top_side:
+			for k in range(side, j):
+				print(argv[top_side][k] + ",", end = " ")
+		top_side -= 1
+		if j <= side:
+			for k in range(top_side, i - 1, -1):
+				print(argv[k][j] + ",", end = " ")
+		j += 1
```
aguimbao, line 32: This looks a lot like C. Take a look at your colleague's answers for more "pythonic" ways to walk through the array.

chall03/mcarrete.py
```diff
@@ -0,0 +1,42 @@
...
+	i = 1
+	error = "usage: ./mcarrete.py <1-9 squared_rows...>"
+	if len(argv) < 2:
+		print(error)
+		return(error)
+	while i < len(argv):
+		if len(argv[i]) != (len(argv) - 1) or argv[i].isdigit() == False:
+			print (error)
+			return error
+		i = i + 1
```
aguimbao, line 45: `i += 1` > `i = i + 1`

chall03/mcarrete.py
```diff
@@ -0,0 +1,42 @@
...
+		j += 1
+
+
+def main(argv):
+	i = 1
+	error = "usage: ./mcarrete.py <1-9 squared_rows...>"
+	if len(argv) < 2:
+		print(error)
+		return(error)
+	while i < len(argv):
```
aguimbao, line 41:
```python
def usage():
    sys.exit("usage: ./mcarrete.py <1-9 squared_rows...>")

for arg in argv:
    if not arg.isdigit():
        usage()
```

This would be better to iterate over the argv.

aguimbao (commented):
I know your code didn't work, but be careful when you commit after the deadline. We won't be able to merge your PR :( You would still have got points for participating in this case if you hadn't commited again.

mcarrete42:
Thank you for all the comments!! Oh! too bad! I tohught the deadline was for the PR not for the commits! :( next time!

aguimbao:
Well. You can imagine that you could just submit a PR with a wrong solution, and then just fix it in a couple commits after the deadline. So we are not allowing that.

mcarrete42:
Makes sense! 👌

PR #164, opened 2020-04-10
vde-dios.c

vdedios:
**WARNING**
I know I am out of deadline but I've decided to submit a "decent" but out of time solution rather than a bad one in time.

**SOLUTION**
The snail pattern goes from outside to inside always repeating two patterns, forward and backwards. The forward patter always prints a full row plus the next rows last positions. The backwards, always a full reverse row and the first positions of the preceding rows.

Whenever a whole turn is completed (forward plus backward) boundary limits are reduced by one.


clafoutis42:
Alright, I'm closing the PR so we don't get confused but I'll review it once we're done with the rest ;)

vdedios:
Thank you so much!!

chall03/vde-dios.py
```diff
@@ -0,0 +1,49 @@
...
+import sys
+
+
+def snail(arg):
+    top = len(arg[1])
+    ini = 1
+    while (ini <= top):
+        i = ini
+        while (i <= top):
+            print(arg[ini][i - 1] if (ini is 1 and i is ini) else ", %s" % arg[ini][i - 1], end = '')
```
aguimbao, line 12: Use `==` or `!=` to compare str, bytes and int literals. Not `is`

chall03/vde-dios.py
```diff
@@ -0,0 +1,49 @@
...
+            while ( i > ini):
+                print(", %s" % arg[i][ini - 1], end = '')
+                i -= 1
+        ini += 1
+        top -= 1
+
+
+def check_args():
+    for c in range(len(sys.argv) - 1):
+        if len(sys.argv) - 1 != len(sys.argv[c + 1]) or len(sys.argv) > 21:
```
aguimbao, line 33: You check `len(sys.argv) > 21` in every loop, should just make that check one time.

I think this is a bad way to check if every argument has the same len.

Better way:

```python
if len(sys.argv) < 2 or not all (len(v) == len(sys.argv[1]) for v in sys.argv[1:]):
    usage()
```

chall03/vde-dios.py
```diff
@@ -0,0 +1,49 @@
...
+            return None
+        for c in sys.argv[c + 1]:
+            if not c.isnumeric():
```
aguimbao, line 36: I think in this exercise, `.isdigit()` method > `.isnumeric()` method.

You can directly perform `.isdigit()` on the string, rather than on each character.

So, like before:

```python
if not all (v.isdigit() for v in sys.argv[1:]):
    usage()
```

chall03/vde-dios.py
```diff
@@ -0,0 +1,49 @@
...
+                return None
+    return sys.argv
```
aguimbao, line 38: You don't have to return sys.argv, you can access it from anywhere.

chall03/vde-dios.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def main():
+    arg = check_args()
+    if arg is None:
+        return print("usage: %s <login> <a-A-Z string>" % sys.argv[0])
```
aguimbao, line 44: I think `sys.exit("usage: %s <login> <a-A-Z string>" % sys.argv[0])` would have been better.

You are using usage message from the last challenge.

chall03/vde-dios.py
```diff
@@ -0,0 +1,49 @@
...
+            i = top
+            while ( i >= ini):
+                print(", %s" % arg[top][i - 1], end = '')
+                i -= 1
+            i = top - 1
+            while ( i > ini):
+                print(", %s" % arg[i][ini - 1], end = '')
+                i -= 1
+        ini += 1
+        top -= 1
```
aguimbao, line 28: This looks a lot like C. Check out some of your colleague's code for more "pythonic" ways to solve it (The ones who didn't cheat).

chall03/vde-dios.py
```diff
@@ -0,0 +1,49 @@
...
+            print(", %s" % arg[i][top - 1], end = '')
+            i += 1
+        if (ini < top):
+            i = top
+            while ( i >= ini):
+                print(", %s" % arg[top][i - 1], end = '')
+                i -= 1
+            i = top - 1
+            while ( i > ini):
+                print(", %s" % arg[i][ini - 1], end = '')
```
aguimbao, line 25: `end=''` > `end = ' '`

aguimbao (commented):
Your approach is interesting, and it works in a decent time. Also execution permissions and shebang are on point.

However your program crashes with no arguments and your usage message is wrong. Also you commited after the deadline :(

vdedios:
Well, thanks for all the feedback, I will consider all your advice for next challenge!!

PR #166, opened 2020-04-12
Fixed Problems

mepg72:
Hi, it is my new version that fix the problems found in my program. Thanks for all.

### DIFF
--- a/chall03/mpena-gu.py
+++ b/chall03/mpena-gu.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python3
 import sys
 
 def get_argv_list(argv):
@@ -17,20 +17,24 @@ def create_squared_rows(list_argv):
     return matrix_list
 
 def check_argv(list_argv):
-    num_argv = 0
-    len_ref_argv = len(list_argv[0])
-    if len_ref_argv == 1:
-        if len(list_argv) != 2:
-            return False
-    elif len(list_argv) != len_ref_argv:
+    if len(list_argv) < 2:
         return False
     else:
-        for argv in list_argv:
-            num_argv += 1 
-            len_argv = len(argv)
-            if not argv.isdigit() or len_argv != len_ref_argv:
-                return False
-    return True
+        list_row = list_argv[1:] 
+        len_ref_argv = len(list_row[0])
+        num_row = len(list_row)
+        if num_row == 1 and len_ref_argv > 1:
+            return False
+        if num_row > 2 and num_row != len_ref_argv:
+            return False
+        else:
+            num_argv = 0 
+            for argv in list_row:
+                num_argv += 1 
+                len_argv = len(argv)
+                if not argv.isdigit() or len_argv != len_ref_argv:
+                    return False
+        return True
 
 def get_snail_str(snail_result):
     argv_row = []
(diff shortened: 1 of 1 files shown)


chall03/mpena-gu.py
```diff
@@ -17,20 +17,24 @@ def create_squared_rows(list_argv):
     return matrix_list
 
 def check_argv(list_argv):
-    num_argv = 0
-    len_ref_argv = len(list_argv[0])
-    if len_ref_argv == 1:
-        if len(list_argv) != 2:
-            return False
-    elif len(list_argv) != len_ref_argv:
+    if len(list_argv) < 2:
```
aguimbao, line 20: You should also check for `len(list_argv) > 21`

chall03/mpena-gu.py
```diff
@@ -17,20 +17,24 @@ def create_squared_rows(list_argv):
...
+        if num_row > 2 and num_row != len_ref_argv:
+            return False
+        else:
+            num_argv = 0 
+            for argv in list_row:
+                num_argv += 1 
+                len_argv = len(argv)
+                if not argv.isdigit() or len_argv != len_ref_argv:
+                    return False
+        return True
```
aguimbao, line 37: function `check_args()` in two line:

```python
if (len(list_args) < 2 or len(list_args) > 21) or (not all (len(v) == len(list_args[1]) and v.isdigit() for v in list_args[1:])):
    return False
```

chall03/mpena-gu.py
```diff
@@ -58,27 +62,35 @@ def snail (row_init, row_end, col_init, col_end, squared_rows):
...
+            row_init = i
+            row_end  = size_array - i
+            col_init = i
+            col_end  = size_array - i
+            snail_result += snail(row_init,row_end,col_init,col_end, squared)
+            i += 1
+        print(snail_result)
+
+def main():
+    if not check_argv(sys.argv):
```
aguimbao, line 90: You don't have to send `sys.argv` anywhere, you can access `sys.argv` from any part of the program.

chall03/mpena-gu.py
```diff
@@ -58,27 +62,35 @@ def snail (row_init, row_end, col_init, col_end, squared_rows):
...
+        print ("usage: %s <1-9 squared_rows...>" % sys.argv[0])
```
aguimbao, line 91: Would have been better: `sys.exit("usage: %s <1-9 squared_rows...>" % sys.argv[0])`

But still, you should have a `usage()` function better.

chall03/mpena-gu.py
```diff
@@ -17,20 +17,24 @@ def create_squared_rows(list_argv):
...
+        else:
+            num_argv = 0 
+            for argv in list_row:
+                num_argv += 1 
+                len_argv = len(argv)
+                if not argv.isdigit() or len_argv != len_ref_argv:
+                    return False
+        return True
 
 def get_snail_str(snail_result):
```
aguimbao, line 39: Should remove this function, not used.

chall03/mpena-gu.py
```diff
@@ -58,27 +62,35 @@ def snail (row_init, row_end, col_init, col_end, squared_rows):
...
+        while i< num_squared:
+            if i>0:
+                snail_result += ', '
+            row_init = i
+            row_end  = size_array - i
+            col_init = i
+            col_end  = size_array - i
+            snail_result += snail(row_init,row_end,col_init,col_end, squared)
+            i += 1
+        print(snail_result)
```
aguimbao, line 87: You had already sanitised the input so, you could have printed the solution character by character, instead of wasting memory storing everything inside a variable and printing it at the end.

aguimbao (commented):
You crash on `./mpena-gy.py "123" "456"`.

We won't correct any more PRs after the deadline, or accept them ! Good luck on next challenge :)

PR #169, opened 2020-04-13
edramire - Prometeo

edramir18:
The program based on the number of arguments supplied work with the standard input or the file names passed as arguments.
The program check if the first line contains only words formed by digits, as well the first word from the second line and the followind lines.
To decide the number of shelfs to give as answer the program sort the size of each shelf available from greater to lesser, based in this order the program decreases the number of books from the total, in case the result from the previous operation is less or equal to zero, the number of shelf processed is showed as the answer to the input file.


chall04/edramire.py
```diff
@@ -0,0 +1,52 @@
...
+                if len(shelfs_str) <= 1 or not shelfs_str[0].isdigit():
+                    print(f"./edramire.py: {filename or 'stdin'}: Can't read file")
+                    return
+                books += int(shelfs_str[0])
+        shelfs.sort(reverse=True)
+        for k in range(len(shelfs)):
+            books -= shelfs[k]
+            if books <= 0:
+                print(k + 1)
+                return
```
clafoutis42, line 32: Have you tried an input like:
```
10 10
20 book1
```
You can't split a book in to, this input is impossible, however your code returns that I need 2 shelves, however this case should return an error, as none of the shelves fits your book.

chall04/edramire.py
```diff
@@ -0,0 +1,52 @@
...
+        print(f'./edramire.py: {filename or "stdin"}: Not enough space in the given shelves')
+    except (PermissionError, FileNotFoundError, RuntimeError):
+        print(f"./edramire.py: {filename or 'stdin'}: Can't read file")
```
clafoutis42, line 35: You might want to be consistent with the quotes you use, I would prefer using double-quotes as main quotes:
```suggestion
        print(f"./edramire.py: {filename or 'stdin'}: Not enough space in the given shelves")
    except (PermissionError, FileNotFoundError, RuntimeError):
        print(f"./edramire.py: {filename or 'stdin'}: Can't read file")
```
That's a matter of taste of course, you could do the opposite, but keep in mind that for readability you should define a rule and stick with it

chall04/edramire.py
```diff
@@ -0,0 +1,52 @@
...
+    try:
+        for line in fileinput.input(filename):
+            shelfs_str = line.split(" ")
+            if len(shelfs) == 0:
+                for word in shelfs_str:
+                    word = word.rstrip('\n\r')
+                    if len(word) == 0:
+                        continue
+                    if not word.isdigit():
+                        print(f"./edramire.py: {filename or 'stdin'}: Can't read file")
```
clafoutis42, line 19: If `filename` is an empty string and is not `None`, you will try to read stdin because `"" or "stdin"` will always return `"stdin"`. However if `filename` isn't `None` it means that you somehow tried to pass `""` as a file, and you should not try to read stdin. In some cases, such a behavior could cause various problems.

It is not an error I would penalize here, but if you run `./edramire.py ""` it will in fact pass a parameter `""` (being an empty string), so you could have simulated this error case.

chall04/edramire.py
```diff
@@ -0,0 +1,52 @@
...
+    shelfs = []
+    books = 0
+    try:
+        for line in fileinput.input(filename):
+            shelfs_str = line.split(" ")
+            if len(shelfs) == 0:
+                for word in shelfs_str:
+                    word = word.rstrip('\n\r')
+                    if len(word) == 0:
+                        continue
```
clafoutis42, line 17: You're applying `rstrip` on all your words here, but you could have done it earlier on `line` which would have reduced the number of function calls.

I also note that you are jumping empty words, but getting empty words should never happen if you user `.split()` without parameters. Also, using split without parameters will exempt you from using `.rstrip`
```suggestion
        for line in fileinput.input(filename):
            shelfs_str = line.split()
            if len(shelfs) == 0:
                for word in shelfs_str:
```
A little example to illustrate this:
```python
In [1]: "aaaaaaaaaaaaaaaaaaa                 \n\t                  bbbbbbbbbbbbbbbbbbb\n\n\n\n\n\r".split()
Out[1]: ['aaaaaaaaaaaaaaaaaaa', 'bbbbbbbbbbbbbbbbbbb']
```

Then again, you could have replaced this whole loop with a comprehension. Even if that implies catching the `TypeError` exception.

chall04/edramire.py
```diff
@@ -0,0 +1,52 @@
...
+                    if not word.isdigit():
+                        print(f"./edramire.py: {filename or 'stdin'}: Can't read file")
+                        return
+                    shelfs.append(int(word))
+            else:
+                if len(shelfs_str) <= 1 or not shelfs_str[0].isdigit():
+                    print(f"./edramire.py: {filename or 'stdin'}: Can't read file")
+                    return
+                books += int(shelfs_str[0])
```
clafoutis42, line 26: Are you basically merging the books into a big one? :grin:

clafoutis42 (commented):
Your code works but does not solve the problem. I think you misunderstood the subject, it sounds like you are thinking you can either merge shelves or split books, but you can't, a book can't be on two shelves at the same time.  
It was a difficult problem though, nice try!

edramir18:
Thanks for the feedback, i working in the python bootcamp to improve the many flaws in my python code.
I knew something was wrong in my approach. That solution was very silly.

PR #170, opened 2020-04-13
Chal04-migferna

migferna42:
The data of the file, organized in a set of numbers for the shelves, and a structure (width, title) for the books.

The first thing is to order the data of the bookshelfs, from largest to smallest, so that we can put the books with more width, if the book enters, its value is updated, and if we do not go to the next

The index does not say the number of bookshelfs we need

To manage the data type of the shelves, I consider that if we pass it: 20 10 test 5 15, the shelf with test value is ignored, and in the list of estates there would only be [20, 10, 5, 15].

For error handling of books, I have considered accepting all books, until I find the first one in the wrong format, there we stop reading.

### DIFF
--- a/chall04/migferna.py
+++ b/chall04/migferna.py
@@ -74,7 +74,7 @@ def main():
             print(filename + ":")
             bookshelfs, books = open_files(filename)
             if bookshelfs != None and books != None:
-                print(order(bookshelfs, books))
+                print(order(bookshelfs, books), end='\n\n')
     else:
         bookshelfs, books = open_files(sys.argv[1])
         print(order(bookshelfs, books))
--- a/chall04/migferna.py
+++ b/chall04/migferna.py
@@ -23,12 +23,16 @@ def order(bookshelfs, books):
     return i if len(bookshelfs) == 0 else i+1
 
 def usage(message):
-    print("%s: %s: %s" % (sys.argv[0], sys.argv[1], message))
+    if (len(sys.argv) > 1):
+        print("%s: %s: %s" % (sys.argv[0], sys.argv[1], message))
+    else:
+        print("%s: stdin: %s" % (sys.argv[0], message))
 
 def open_files(filename):
     books = []
+    bookshelfs = []
     if (filename == os.path.basename(__file__)):
-        return None, None
+        return bookshelfs, books
     try:
         with open(filename, "r") as file:
             for i, line in enumerate(file):
@@ -44,7 +48,7 @@ def open_files(filename):
                     books.append(book)
         return bookshelfs, books
     except (FileNotFoundError, PermissionError):
-        return None, None
+        return bookshelfs, books
     
 
 def read():
@@ -52,7 +56,7 @@ def read():
     bookshelfs = []
     for i, line in enumerate(sys.stdin):
         if '' == line.rstrip():
-            break
+            return bookshelfs, books
         elif i == 0:
             bookshelfs = line.split()
             bookshelfs = [s for s in bookshelfs if s.isdigit()]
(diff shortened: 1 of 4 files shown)


clafoutis42:
I am currently reviewing your code, however as your PR embeds unwanted files, I have to close it. Sorry.

chall04/migferna.py
```diff
@@ -0,0 +1,112 @@
...
+        return bookshelfs, books
+    except (FileNotFoundError, PermissionError, ValueError, IndexError):
+        return bookshelfs, books
+    
+
+def read():
+    books = []
+    bookshelfs = []
+    for i, line in enumerate(sys.stdin):
+        if '' == line.rstrip():
```
clafoutis42, line 58: I see that you are using double quotes and single quotes in your code. To improve readability you should define a rule about this and stick with it.  

chall04/migferna.py
```diff
@@ -0,0 +1,112 @@
...
+            try:
+                line = line.split(" ", 1)
+                book = {
+                    "width": int(line[0]),
+                    "title": line[1].rstrip("\n")
+                }
+                books.append(book)
+            except (ValueError, IndexError):
+                return bookshelfs, books
+    return bookshelfs, books
```
clafoutis42, line 73: Why create 2 functions to do the same thing?
You could have created one single `read` function that takes a filedescriptor as a parameter.
e.g:
```python
if filename is None:
    return read(sys.stdin)
with open(filename, "r") as f:
    return read(f)
```

chall04/migferna.py
```diff
@@ -0,0 +1,112 @@
...
+                if i == 0:
+                    bookshelfs =  line.split()
+                    bookshelfs = [s for s in bookshelfs if s.isdigit()]
+                else:
+                    line = line.split(" ", 1)
+                    book = {
+                        "width": int(line[0]),
+                        "title": line[1].rstrip("\n")
+                    }
+                    books.append(book)
```
clafoutis42, line 48: I don't like this `if i == 0:`, why putting this in the loop if you know that you will always execute this on the first line?  
```suggestion
            bookshelfs = [int(s) for s in f.readline().split()]
            for i, line in enumerate(file):
                line = line.split(" ", 1)
                book = {
                    "width": int(line[0]),
                    "title": line[1].rstrip("\n")
                }
                books.append(book)
```
You could simplify it even more by using a comprehension
```suggestion
            bookshelfs = [int(s) for s in f.readline().split()]
            books = [{
                    "width": int(line.split(" ", 1)[0]),
                    "title": line.split(" ", 1)[1].rstrip("\n")
                } for line in file]
```

chall04/migferna.py
```diff
@@ -0,0 +1,112 @@
...
+    else:
+        bookshelfs, books = open_files(sys.argv[1])
+        if len(bookshelfs) > 0 and len(books) > 0:
+            output = order(bookshelfs, books)
+            if output != None:
+                print(output)
+            else:
+                usage("Not enough space in the given shelves")
+        else:
+            usage("Can't read file")
```
clafoutis42, line 109: Don't you think you're repeating the same code over and over?
You could basically use a function to englobe all the common part and then just do a simple main like:
```python3
    args = sys.argv[1:]
    if not args:
        return bookshelf(None)
    for arg in args:
        bookshelf(arg, printfile=len(args) > 1)
```
This is just a schema but that should give you the idea

chall04/migferna.py
```diff
@@ -0,0 +1,112 @@
...
+        for i, bookshelf in enumerate(bookshelfs):
+            if book['width'] <= bookshelf:
+                bookshelf -= book['width']
+                bookshelfs[i] = bookshelf
+                break
+            elif len(bookshelfs) == i + 1:
+                return
+            else:
+                bookshelf + 1
+    return i if len(bookshelfs) == 0 else i+1
```
clafoutis42, line 23: I had trouble understanding what you did here. There is a simpler and more readable way to reproduce this behavior.  
I understand you want to iterate on books and shelves, however the `for ... in ...` loop is not what you need here as you can't iterate according to a condition.

You might want to create an iterator for your shelves before entering the books loop, it would look like this
```python3
shelves_iter = iter(bookshelfs)
current_shelf = next(shelves_iter) # This is to initiate the variable and get the first shelf
for book in books:
    if shelf >= book["width"]:
        # Handle a book that fits
        continue # You stored the book, you may now continue to the next book
    try:
        shelf = next(shelf)
        # Validate your new shelf
    except StopIteration: # next(iterator) raises this exception when it can't iterate anymore
        break
```

Later I advise you to look at the `for ... else ...` syntax to handle your return values better

clafoutis42 (commented):
Your solution works for some cases, but fails on most of others (e.g: the big example case in the README)

chall04/migferna.py
migferna42: I thought of the same thing, but as a function I only read 1 time, and the other read for n-arguments ..., but if it's true, if I call from outside there is no problem. I didn't think about it enough ;)

chall04/migferna.py
migferna42: How simple it looks now! haha

chall04/migferna.py
migferna42: Thanks, from now on I will maintain a standard.

PR #171, opened 2020-04-13
vde-dios.py

vdedios:
**SOLUTION**

Given an input the program first checks if input comes from stdin or a file/s. When parsing information it checks wether shelf and books format is ok or not. If any error is encountered a corresponding error message is returned. Handled errors are:
- _Can't read file:_ Wrong input file
- _Not enough space in the given shelves:_ Not enough space for given shelves
- _Bad format_ (this one was added to better describe error if any input format mismatch is encountered) Valid input format is considered as:

   1. Shelves, only numbers separated by a (one) space
   2. Books: number + single space + any set of characters
   3. If no shelf or books lines are encountered it is also considered a bad format error since the problem wouldn't make sense

The outcome shelf decision is pretty straight forward. Shelves ares stored in a list. This list gets sorted from bigger to smaller. Then all books are added up so we get the total. Finally we start "filling" shelves with books . Since we have the shelves sorted from bigger to smaller, once we finish placing all books we will always get the minimum shelves required and no shelf will ever be empty.


chall04/vde-dios.py
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env python3
+
+import sys
+import re
+
+
+def error(e , i):
+    file_name =  "stdin" if i == 0 else sys.argv[i]
+    if (e == 1):
+        print("%s: %s: Bad format" %(sys.argv[0], file_name))
```
clafoutis42, line 10: We're going to accept that, because it is intelligent, however it was stated in the README that ALL the expected error messages were present in the files. You didn't have to create new ones.  

That said, `Can't read file` is a very general type of message, there are many reasons why a file can't be read:
- It doesn't exist
- You don't have permissions
- You don't recognize the format

As you might already have guessed, I'm trying to say that it is incoherent to be specific about the format error, and not the rest.

chall04/vde-dios.py
```diff
@@ -0,0 +1,82 @@
...
+    elif (e == 2):
+        print("%s: %s: Can't read file" %(sys.argv[0], file_name))
+    elif (e == 3):
+        print("%s: %s: Not enough space in the given shelves" %(sys.argv[0], file_name))
+
+
+# Check format plus sort shelves
+def shelf_parse(shelves):
+    l = shelves.replace("\n", "").split(" ")
```
clafoutis42, line 19: You do not need to replace the `\n`s if you use split without parameters:
```suggestion
    l = shelves.split()
```

chall04/vde-dios.py
```diff
@@ -0,0 +1,82 @@
...
+    l.sort(key=len, reverse=True)
+    for c in l:
+        if not c.isdigit():
+            return None
```
clafoutis42, line 23: Why would you sort by len? If you want to somehow give a priority to the largest numbers, this won't work if I pass `000001` as a shelf, the value would be `1` but it would be placed before `400`.  
Something more logical would have been to sort it by value:
```suggestion
    if not "".join(l).isdigit():
        return None
    l = sorted([int(s) for s in l], reverse=True)
```
Also, try using more explicit var names `l` is really short and might get confusing (as you can see in the comprehension I wrote)

chall04/vde-dios.py
```diff
@@ -0,0 +1,82 @@
...
+    return (l)
+
+
+# Check book format and add books
+def book_parse(books):
+    b = re.search('^\d+', books)
+    if re.search('^\d+\s\w+', books) is None:
+        return None
+    return int(b.group())
```
clafoutis42, line 32: That's a good and clean way to check for the format, some would have prefer to simply catch exceptions (like me).  

Although your regex is not really correct, what if I accidentally place to spaces between the size and the title?
And if you want to keep the title, your regex does not englobe the whole title.  
A correct regexp would have been: `^(\d+)\s+([^\s].+)$`
I chose to allow titles with other characters than the ones in `\w`, a title can have spaces, hyphens...

That said, you do not have to use 2 regexp here, as you noticed I used parentheses to define groups:
```suggestion
# Check book format and add books
def book_parse(books):
    result = re.search(r"^(\d+)\s+([^\s].+)$", books) # Always use rstrings to make regexp, it will be parsed differently to avoid conflicts (with `\` for example)
    if result is None:
        return None
    return int(result.group(1)) # result.group(1) will return the book's size, result.group(2) will return the book's title
```

chall04/vde-dios.py
```diff
@@ -0,0 +1,82 @@
...
+    l.sort(key=len, reverse=True)
+    for c in l:
+        if not c.isdigit():
+            return None
+    return (l)
+
+
+# Check book format and add books
+def book_parse(books):
+    b = re.search('^\d+', books)
```
clafoutis42, line 29: I see that you are using double quotes and single quotes in your code. To improve readability you should define a rule about this and stick with it.

chall04/vde-dios.py
```diff
@@ -0,0 +1,82 @@
...
+    capacity = 0
+    for c in shelves:
+        capacity += int(c)
+        n_shelves += 1
+        if capacity >= books:
+            break
+    if capacity < books:
+        error(3, i)
+    else:
+        print(n_shelves)
```
clafoutis42, line 47: I'm guessing you didn't understand the subject. You can't split books, and a book can't be on more than one shelf.  
Given this output your program will return `2` when it should return an error:
```
10 10
20 book
```

chall04/vde-dios.py
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env python3
+
+import sys
+import re
+
+
+def error(e , i):
+    file_name =  "stdin" if i == 0 else sys.argv[i]
```
clafoutis42, line 8: What does `i` mean? You should use more explicit variable names. Also you are in python, it sounds here like you should use the `bool` type

chall04/vde-dios.py
```diff
@@ -0,0 +1,82 @@
...
+    if (e == 1):
+        print("%s: %s: Bad format" %(sys.argv[0], file_name))
+    elif (e == 2):
+        print("%s: %s: Can't read file" %(sys.argv[0], file_name))
+    elif (e == 3):
+        print("%s: %s: Not enough space in the given shelves" %(sys.argv[0], file_name))
```
clafoutis42, line 14: Using more explicit identifiers for your errors would have been better, having to go back to the function each time I see a call to guess what will be the error message is a little annoying.  
You could have used different solution:
- A global variable
- Use the `enum.Enum` class
- Create your own types of exceptions to catch them later

chall04/vde-dios.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def main():
+    l = len(sys.argv)
+    if (l > 1):
+        for i in range(1, l):
+            try:
+                print("%s:\n" % sys.argv[i] if l > 2 else "", end = "")
+                f = open(sys.argv[i], "r")
+                number_of_shelves(f, i)
```
clafoutis42, line 72: I understand the function `number_of_shelves` closes the file descriptor, but this is a really bad practice to close it in another function like this.  
Use the `with ... as ...` syntax to avoid mistakes and be more readable:
```suggestion
                with open(sys.argv[i], "r") as f:
                    number_of_shelves(f, i)
```
Of course you would remove the `f.close()` from `number_of_shelves`

chall04/vde-dios.py
```diff
@@ -0,0 +1,82 @@
...
+            except:
```
clafoutis42, line 73: That's a really wide exception, you might want to be more precise on which exceptions to catch to avoir debugging problems

chall04/vde-dios.py
```diff
@@ -0,0 +1,82 @@
...
+                error(2, i)
+            print("\n" if i < l - 1 else "", end="")
```
clafoutis42, line 75: That's a little complex for something you could have written with a simple `if`
```suggestion
            if i < l - 1:
                print()
```

clafoutis42 (commented):
Your code works, but doesn't solve the problem. I think you did not understand the problem, you can't place a book on more than one shelf, and you can't split a book in half.

vdedios:
I just have realized about the subject misconception... What I understood was that the number in front of each book was the number copies you wanted to place and the number in shelves was the number of copies you could place...When more focussed on coding fast, more chances to make things wrong. Thank you!

chall04/vde-dios.py
vdedios: Totally agree with that. I realized later I was trying to sort strings rather than integers which does not work in the same way

chall04/vde-dios.py
vdedios: The reason for only using one space in **_re.search('^\d+\s\w+', books)_** was because in the subject was explicitly defined that book lines were have to be defined as a number + **a** space + tittle. 

About the usage of regexp is really smart, take note :)

chall04/vde-dios.py
vdedios: 📝

chall04/vde-dios.py
vdedios: I did not know about this syntax, take note!📝

PR #172, opened 2020-04-13
Chall04 - pcuadrad

pabloocg:
First in the main function check if parameters have been entered, if it has not been entered, I read the stdin and save it in a list.
If they have entered parameters, I try to open the files and in case there is an error when opening the file, the exception will be executed and the next file will continue.

I have created the function parse_file what it does is to parse the text, both the one that is saved in the file and the one that it receives from the stdin, the first line of the file I keep it apart and a split is made by the spaces and they will be converted to integers numbers, the rest of the lines I do the same but the split I do it only once to get the width of the book which is what mainly interests.

Once I have a list with the number of shelves and their size and another list with the width of all the books I call place_books, what I do in this function is go 'placing' (subtracting) each book by filling each shelf, which by the way is ordered from largest to smallest since I understand from the statement of the exercise that it is necessary to return the smallest number of shelves that are needed. 
When all the books are 'placed' and the number of each shelf is positive, I print the number of necessary shelves and return True, otherwise I return False and the error message will be executed.

### DIFF
--- a/chall04/pcuadrad.py
+++ b/chall04/pcuadrad.py
@@ -12,7 +12,7 @@ def place_books(books, shelfs):
         begin = shelfs[i]
         begin += rest
         rest = 0
-        if begin < 0:
+        if begin < 0 and i is len(shelfs) - 1:
             return False
         for j in range(last_book, len(books)):
             begin -= books[j]
@@ -23,6 +23,7 @@ def place_books(books, shelfs):
             if j is len(books) - 1:
                 print (i + 1)
                 return True
+    print (i + 1)
     return True
 
 def parse_file(txt):
@@ -59,7 +60,7 @@ def main():
             if len(sys.argv) is not 2:
                 print (sys.argv[i] + ":")
             try:
-                if sys.argv[i].find(".py"):
+                if sys.argv[i].find(".txt") is -1:
                     error_message(sys.argv[i], "Can't read file")
                     if len(sys.argv) is not 2 and i is not len(sys.argv) - 1:
                         print()
--- a/chall04/pcuadrad.py
+++ b/chall04/pcuadrad.py
@@ -1,0 +1,79 @@
+#!/usr/bin/env python3
+
+import sys
+
+def error_message(filename, message):
+    print (sys.argv[0] + ": " + filename + ": " + message)
+
+def place_books(books, shelfs):
+    rest = 0
+    last_book = 0
+    for i in range(0, len(shelfs)):
+        begin = shelfs[i]
+        begin += rest
+        rest = 0
+        if begin < 0:
+            return False
+        for j in range(last_book, len(books)):
+            begin -= books[j]
+            if begin < 0:
+                rest = begin
+                last_book = j + 1
+                break
+            if j is len(books) - 1:
+                print (i + 1)
+                return True
+    return True
+
+def parse_file(txt):
+    i = 0
+    books = []
+    shelfs_int = []
+    for line in txt:
+        if i is 0:
+            shelfs = line.split(' ')
+        else:
+            line2 = line.split(' ', 1)
+            books.append(int(line2[0]))
+        i += 1
+    [shelfs_int.append(int(i)) for i in shelfs]
+    shelfs_int = sorted(shelfs_int, reverse=True)
+    books = sorted(books, reverse=True)
+    if not place_books(books, shelfs_int):
+        return False
+    else:
+        return True
+
+def parse_stdin():
+    ret = []
+    for line in sys.stdin:
+        ret.append(line[:-1])
+    return ret
+
+def main():
+    if len(sys.argv) is 1:
+        if not parse_file(parse_stdin()):
+            error_message("stdin", "Not enough space in the given shelves")
+    else:
+        for i in range(1, len(sys.argv)):
+            if len(sys.argv) is not 2:
+                print (sys.argv[i] + ":")
+            try:
+                if sys.argv[i].find(".py"):
+                    error_message(sys.argv[i], "Can't read file")
+                    if len(sys.argv) is not 2 and i is not len(sys.argv) - 1:
+                        print()
+                    continue
+                file = open(sys.argv[i],'r')
+            except (OSError, PermissionError):
+                error_message(sys.argv[i], "Can't read file")
+            else:
+                content = file.read()
+                if not parse_file(content.split('\n')):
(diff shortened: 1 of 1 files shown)


chall04/pcuadrad.py
```diff
@@ -0,0 +1,80 @@
...
+    print (i + 1)
+    return True
+
+def parse_file(txt):
+    i = 0
+    books = []
+    shelfs_int = []
+    for line in txt:
+        if i is 0:
+            shelfs = line.split(' ')
```
clafoutis42, line 35: I see that you are using double quotes and single quotes in your code. To improve readability you should define a rule about this and stick with it.
I for one use double quotes as main quotes and single quotes when double quotes aren't usable (for example print(f"Value: {dict['key']}") where using double quotes in the dictionary reference would have conflicted with the quotes of the format string)

chall04/pcuadrad.py
```diff
@@ -0,0 +1,80 @@
...
+    if not place_books(books, shelfs_int):
+        return False
+    else:
+        return True
+
+def parse_stdin():
+    ret = []
+    for line in sys.stdin:
+        ret.append(line[:-1])
+    return ret
```
clafoutis42, line 52: Why could not this function read files ?
You just would have to pass it a file descriptor!
```suggestion
def parse_stdin(fd=sys.stdin):
    ret = []
    for line in fd:
        ret.append(line[:-1])
    return ret
```

chall04/pcuadrad.py
```diff
@@ -0,0 +1,80 @@
...
+
+def parse_file(txt):
+    i = 0
+    books = []
+    shelfs_int = []
+    for line in txt:
+        if i is 0:
+            shelfs = line.split(' ')
+        else:
+            line2 = line.split(' ', 1)
```
clafoutis42, line 37: I see that you are using double quotes and single quotes in your code. To improve readability you should define a rule about this and stick with it.

chall04/pcuadrad.py
clafoutis42, line 35: It is not specified that the output can have several spaces to separate shelves, but it might, you should use `.split()` without separators.
```suggestion
            shelfs = line.split()
```

chall04/pcuadrad.py
```diff
@@ -0,0 +1,80 @@
...
+            books.append(int(line2[0]))
+        i += 1
+    [shelfs_int.append(int(i)) for i in shelfs]
+    shelfs_int = sorted(shelfs_int, reverse=True)
```
clafoutis42, line 41: You didn't quite understand the user of comprehension, a comprehension will run and generate a list directly on memory, meaning that if you don't assign it, it is lost.
Here is a little [documentation about comprehensions](https://www.pythonforbeginners.com/basics/list-comprehensions-in-python). I could not explain it better.
```suggestion
    shelfs_int = sorted([int(i) for i in shelfs], reverse=True)
```

chall04/pcuadrad.py
```diff
@@ -0,0 +1,80 @@
...
+    books = sorted(books, reverse=True)
```
clafoutis42, line 42: If you already have the `books` list, doing assigning `sorted` will copy the list then sort it.  
It would have been more optimized to simply call the `sort` method on the list:
```suggestion
    books.sort(reverse=True)
```
You don't have to assign it, the elements will be moved inside the list, and no copy will be done.

chall04/pcuadrad.py
```diff
@@ -0,0 +1,80 @@
...
+    if not place_books(books, shelfs_int):
+        return False
+    else:
+        return True
```
clafoutis42, line 46: So basically you wrote `If not True return False else return True` ? Why don't you want to directly return... well... what is returned ?
```suggestion
    return place_books(books, shelfs_int)
```
😅

chall04/pcuadrad.py
```diff
@@ -0,0 +1,80 @@
...
+def main():
+    if len(sys.argv) is 1:
+        if not parse_file(parse_stdin()):
+            error_message("stdin", "Not enough space in the given shelves")
+    else:
+        for i in range(1, len(sys.argv)):
+            if len(sys.argv) is not 2:
+                print (sys.argv[i] + ":")
+            try:
+                if sys.argv[i].find(".txt") is -1:
```
clafoutis42, line 63: So I can't passe a file that does not have `.txt` as a parameter? You should remember that the extension is an indicator, but does not mean anything. Unfortunately, I have tests that pass files without `.txt`.

Also I advise you to read documentation on regular expression, with the regular expression you used, a valid file name would be anything as long as there is `txt` after any character in the filename (e.g: `readtxt.py` would be a valid filename according to your regex, `fsdkgflsd;gtxtdskgzjslk` would too)

To do it well, you would have needed to do this
```suggestion
                if sys.argv[i].find(r"\.txt$") is -1:
```

chall04/pcuadrad.py
```diff
@@ -0,0 +1,80 @@
...
+                    continue
+                file = open(sys.argv[i],'r')
+            except (OSError, PermissionError):
+                error_message(sys.argv[i], "Can't read file")
+            else:
+                content = file.read()
+                if not parse_file(content.split('\n')):
+                    error_message(sys.argv[i], "Not enough space in the given shelves")
+                file.close()
+            if len(sys.argv) is not 2 and i is not len(sys.argv) - 1:
```
clafoutis42, line 76: NEVER use `is not` in that case. [Read this](https://stackoverflow.com/questions/2209755/python-operation-vs-is-not).

chall04/pcuadrad.py
```diff
@@ -0,0 +1,80 @@
...
+            return False
+        for j in range(last_book, len(books)):
+            begin -= books[j]
+            if begin < 0:
+                rest = begin
+                last_book = j + 1
+                break
+            if j is len(books) - 1:
+                print (i + 1)
+                return True
```
clafoutis42, line 25: You can't split books. Did you try:
```
10 10
20 book
```
It should return an error but it returns that I need 2 shelves.

clafoutis42 (commented):
There is a lot of thing you can improve, would it be in python or in the logic itself. I think you didn't understand the subject, you can't split a book in half, and neither can you place a book on 2 shelves at a time.

pabloocg:
Thanks for the comments and the code improvements, I see that I did not understand the problem well.

PR #173, opened 2020-04-13
chall04-fulldemo

fulldemo13:
Hi, in this program I look if the input is a file or it's by stdin. Every input is save in a list using "readlines()" function. And I send it to "take_shelves()" with the number of the file is. I have discover "try/except" and I use it for check errors. When somethings goes wrong I use "error()" to display the error message.
First I calculate the length that I need to store the books, in function "shelves_to_use()". In this function I also check length is a positive number and that every length has a title.
Second I order the first line of the "file" from largest to smallest. With this I buy the less number of shelves. I add the length of all the shelves and check if the format of this line is correct.
In the end, I calculate the number of shelves that I need, and I check if it is available.
If it's available I return the number of the shelves that I need, if not I return error.
I only use two errors, if the format of "file" is wrong or if I don't have enough shelves.


fulldemo13:
Hi in chall05, I search the character 'i' and I check if te rule is true or false.
When I find a 'i' if after have a 'e' and before have a 'c' the rule isn't true (cie), and return 0.
When I find a 'i' and before have something different to "ce" , I return 0 because the rule is false.
In the other cases the rule always be true, and it returns 1.

chall04/fulldemo.py
```diff
@@ -0,0 +1,80 @@
...
+	if e == 0:
+		if n != 0:
+			print("%s: %s: Can't read file" % (sys.argv[0], sys.argv[n]))
+		else:
+			print("%s: stdin: Can't read file" % (sys.argv[0]))
+	else:
+		if n != 0:
+			print("%s: %s: Not enough space in the given shelves" % (sys.argv[0], sys.argv[n]))
+		else:
+			print("%s: stdin: Not enough space in the given shelves" % (sys.argv[0]))
```
clafoutis42, line 16: Using more explicit identifiers for your errors would have been better, having to go back to the function each time I see a call to guess what will be the error message is a little annoying.  
You could have used different solution:
- A global variable
- Use the `enum.Enum` class
- Create your own types of exceptions to catch them later

chall04/fulldemo.py
```diff
@@ -0,0 +1,80 @@
+#! /usr/local/bin/python3
+
+import sys
+import fileinput
+
+def error(e, n):
```
clafoutis42, line 6: Try to use more explicit var names, I shouldn't have to think about what means the variable, especially when it's the parameter of a function

chall04/fulldemo.py
```diff
@@ -0,0 +1,80 @@
...
+		line = all_file[i].split()
+		if len(line) < 2:
+			return (-1)
+		try:
+			shelves_use = int(line[0]) + shelves_use
+			if int(line[0]) < 0:
+				return (-1)
+		except:
+			return (-1)
+	return(shelves_use)
```
clafoutis42, line 30: Why are you merging the shelves? You won't be able to differentiate them later :/

chall04/fulldemo.py
```diff
@@ -0,0 +1,80 @@
...
+	else:
+		if n != 0:
+			print("%s: %s: Not enough space in the given shelves" % (sys.argv[0], sys.argv[n]))
+		else:
+			print("%s: stdin: Not enough space in the given shelves" % (sys.argv[0]))
+
+def shelves_to_use(all_file):
+	shelves_use = 0
+	for i in range(1, len(all_file)):
+		line = all_file[i].split()
```
clafoutis42, line 21: A more intuitive and pythonic usage of the for loop would be:
```suggestion
	for line in all_file[1:]):
		line = line.split()
```

chall04/fulldemo.py
```diff
@@ -0,0 +1,80 @@
...
+			return (-1)
+	return(shelves_use)
+
+def	check_to_buy(array):
+	for i in range(len(array)):
+		try:
+			if int(array[i]) < 0:
+				return (-1)
+		except:
+			return (-1)
```
clafoutis42, line 38: Why would you return `-1`? You should return either `True` or `False`.
Also there is a more pythonic way to do this
```suggestion
	try:
		return all(i >= 0 for i in array)
	except:
		return False
```

chall04/fulldemo.py
```diff
@@ -0,0 +1,80 @@
...
+		take_shelves(all_file, 0)
+	else:
+		for i in range(1, len(sys.argv)):
+			if len(sys.argv) > 2:
+				print(sys.argv[i] + ":")
+			try:
+				f = open(sys.argv[i], 'r')
+				all_file = f.readlines()
+				take_shelves(all_file, i)
+				f.close()
```
clafoutis42, line 75: You should use the `with` syntax
```suggestion
				with open(sys.argv[i], 'r') as f:
				    all_file = f.readlines()
				    take_shelves(all_file, i)
```

clafoutis42, line 75: You are repeating yourself, you could have made a function that takes a filedescriptor in parameter

clafoutis42 (commented):
Your python code still looks like C code. Also it seems that you did not understand the subject, you are either trying to split your books in half, putting them on more than one shelf, or merging the shelves.

clafoutis42:
Please submit your code for the chall05 on another branch and create a new pull request. Or else we won't accept your PR and you will lose both challenge participation.  
You have until tonight 20.42 to clean up the PR for chall04, if it is not cleaned, we will only accept a clean PR for the chall05.  
If at the time of turning in the chall05 the PR is not cleaned we won't accept your chall05 either.

fulldemo13:
Hi, I think I clean up the branch. Thanks for the corrections I'm discovering phyton with this. 
In this exercise I understood that I need enough shelves length for all books. Dividing the books of the collection or shelves if I need.

PR #174, opened 2020-04-14
Non-operative trial for Chall04 - ssacrist

esepunto:
I have looked for different ways to receive the data and treat them, but I have not been able to achieve what I wanted to do. 

After trying several options, I haven't figured out how I can extract the size of the books and shelves, and I don't know if I'm close to achieving itor I'm so far, so... I didn't work on the coding of the algorithm.  

I don't seem to be able to hit the right code...


chall04/ssacrist.py
```diff
@@ -0,0 +1,52 @@
...
+
+def Count_Lines(c):
+    num_lines = 0
+    with open(sys.argv[c]) as library:
+        for line in library:
+            num_lines += 1
+    return(num_lines)
+
+def main():
+    for c in range(1, len(sys.argv)):
```
aguimbao, line 44: I think it would have been better:

```python
 for c in sys.argv[1:]:
```

chall04/ssacrist.py
```diff
@@ -0,0 +1,52 @@
...
+    for line in books:
+        if line == "EOF":
+            break
+        else:
+            signatures.append(line)
+    quarry(signatures, num_lines)
+
+def Count_Lines(c):
+    num_lines = 0
+    with open(sys.argv[c]) as library:
```
aguimbao, line 38: You don't have to pass this `c` argument, you can access `sys.argv[c]` as:

```python
with open(sys.argv[-2]) as library:
```

chall04/ssacrist.py
```diff
@@ -0,0 +1,52 @@
...
+        for line in library:
+            num_lines += 1
+    return(num_lines)
```
aguimbao, line 41: I don't really understand what you were counting your "lines" here for...

chall04/ssacrist.py
```diff
@@ -0,0 +1,52 @@
...
+def quarry(signatures, num_lines):
+    for c in range (0, num_lines):
+        signatures[c].partition(" ")
+        print(signatures[c][0])
+        print(num_lines)
+
+def chop(books, num_lines):
+    signatures = []
+    for line in books:
+        if line == "EOF":
```
aguimbao, line 30: There's no `"EOF"`. The method you used previously, `.readlines()` reads all lines from the file and stores them in a list `[]`, so you can just `for` loop them until you reach the end.

aguimbao (commented):
Hey, this was a bit of a hard problem.

I think a good approach would have been to use `.readline()` on the file to read the first line, and then `.readlines()` to store the rest of the lines.
Then split the first line with `' '` as a separator, so you have all the shelves sizes, and then split the rest of the lines with `' '` as a separator too, and then just keep the fist `[0]` field of the list, which would tell you the size of the book.

Then it was a matter of brute-forcing to try and find the best combination of books in the shelves.

Sadly, your branch was deleted, so I couldn't even check execution permissions. I can't accept the PR :(

PR #175, opened 2020-04-14
chall04 - gbudau

gbudau:
  After checking that the file/buffer is valid i create two lists one with the bookshelves using first line values and another with books using first word from all the lines after that. 
I convert to integer and sort the lists from max to min.
I create another list to store used bookshelves, moving bookshelves used to this list and removing them from main bookshelves list and checking first here to see if theres enough space.
Each time a bookshelf is moved to the used_bookshelves list the shelf and book are removed from the lists.
Then keep looping until no books are left or the max value of bookshelves/used bookshelves is less than the max value of books.
If there are no `books` left in the list it will output the number of the `used_bookshelves` list else an error message

**Short description of your code**
Make sure input come from buffer or command line arguments.
If the input comes from buffer(heredoc), process and exit.
If it comes from stdin(keyboard) i display an error and exit.
If there are arguments try to open the files or print error if the file can't be open and continue with the next argument.

Open the file(argument), save first line in `bookshelves` list stripping newlines and spaces and making sure it contains only digits.
Save the first string from the  other lines in `books` list stripping newlines and spaces and making sure it contains only digits.
If any of these two operations fail print an error else use `fill_bookshelves` function with the lists created.

In fill_bookshelves function i convert to integer and sort the lists from max to min.
I create another list to store used bookshelves, moving bookshelves used to this list and removing them from main bookshelves list and checking first here to see if theres enough space.
Each time a bookshelf is moved to the used_bookshelves list the shelf and book are removed from the lists.
Then keep looping until no books  are left or the max value of bookshelves/used bookshelves is less than the max value of books.
If there are no `books` left in the list it will output the number of the `used_bookshelves` list else an error message

The algorithm currently needs some optimization, sometimes it needs more bookshelves than necessary and the speed can be improved.
I was not sure if the script should accept input from keyboard/redirect so im outputing an error in those cases.

### DIFF
--- a/chall04/gbudau.py
+++ b/chall04/gbudau.py
@@ -40,9 +40,10 @@ def verify_and_save_file(arg, source):
     # Save first line to bookshelves list from the file generator, striping spaces and newline
     # Splitting the line into a list of numeric strings
     # next() can return an error when the file is empty, without any newline so im using try/except
+    # Also checking the shelves to have non-zero values
     try:
         bookshelves = next(arg).rstrip().split()
-        if not bookshelves or any(not shelf.isdigit() for shelf in bookshelves):
+        if not bookshelves or any(not shelf.isdigit() or shelf[0] == '0' for shelf in bookshelves):
             return None, None
     except StopIteration:
         if source == STDIN:
--- a/chall04/gbudau.py
+++ b/chall04/gbudau.py
@@ -1,0 +1,93 @@
+#!/usr/bin/env python3
+
+import sys
+
+STDIN = 0
+FILE = 1
+
+def read_error(filename):
+    print("{}: {}: Can't read file".format(sys.argv[0], filename))
+
+def space_error(filename):
+    print("{}: {}: Not enough space in the given shelves".format(sys.argv[0], filename))
+
+def fill_bookshelves(filename, bookshelves, books):
+    # Convert books and bookshelves to integers
+    # Sort from max to min
+    books = [int(i) for i in books]
+    bookshelves = [int(i) for i in bookshelves]
+    books.sort(reverse = True)
+    bookshelves.sort(reverse = True)
+
+    used_bookshelves = []
+    while books and bookshelves and ((bookshelves[0] >= books[0]) or (used_bookshelves and used_bookshelves[0] >= books[0])):
+        if  used_bookshelves and used_bookshelves[0] >= books[0]:
+            used_bookshelves[0] -= books[0]
+            books.pop(0)
+            used_bookshelves.sort(reverse = True)
+        else:
+            used_bookshelves.append(bookshelves[0] - books[0])
+            used_bookshelves.sort(reverse = True)
+            books.pop(0)
+            bookshelves.pop(0)
+
+    space_error(filename) if books else print(len(used_bookshelves))
+
+def verify_and_save_file(arg, source):
+    bookshelves = []
+    books = []
+
+    # Save first line to bookshelves list from the file generator, striping spaces and newline
+    # Splitting the line into a list of numeric strings
+    # next() can return an error when the file is empty, without any newline so im using try/except
+    try:
+        bookshelves = next(arg).rstrip().split()
+        if not bookshelves or any(not shelf.isdigit() for shelf in bookshelves):
+            return None, None
+    except StopIteration:
+        if source == STDIN:
+            read_error("stdin")
+            sys.exit(0)
+        else: return None, None 
+
+    # Save first word from each line starting from second line into books list
+    # Checking if they are numeric and not empty
+    for line in arg:
+        books += line.rstrip().split()[:1]
+    if not books or any(not book.isdigit() for book in books):
+        return None, None
+    return bookshelves, books
+
+def main():
+    # Input come from buffer and not direct input, accepting heredoc only
+    if not sys.stdin.isatty():
+        bookshelves, books = verify_and_save_file(sys.stdin, STDIN)
+        if (not bookshelves or not books):
+            read_error("stdin")
+        else:
+            fill_bookshelves("stdin", bookshelves, books)
+        sys.exit(0)
+
+    # Running the script without input, treating this case as read error
+    if len(sys.argv) == 1:
+        read_error("stdin")
+
+    # Iterate over arguments, check for IO error, verify file and process
+    for i, arg in enumerate(sys.argv[1:], start=2):
+        try:
+            with open(arg) as f:
+                bookshelves, books = verify_and_save_file(f, FILE)
+                len(sys.argv) > 2 and print(arg, end=':\n')
+                if (not bookshelves or not books):
+                    read_error(arg)
+                    i < len(sys.argv) and print()
+                else:
+                    fill_bookshelves(f.name, bookshelves, books)
+                    i < len(sys.argv) and print()
+        except IOError:
+            len(sys.argv) > 2 and print(arg, end=':\n')
+            read_error(arg)
+            i < len(sys.argv) and print()
+
+if __name__ == '__main__':
+    main()


esepunto:
I love this code. Very well explained and very understandable. Thanks for sare! :)

chall04/gbudau.py
```diff
@@ -0,0 +1,94 @@
...
+def fill_bookshelves(filename, bookshelves, books):
+    # Convert books and bookshelves to integers
+    # Sort from max to min
+    books = [int(i) for i in books]
+    bookshelves = [int(i) for i in bookshelves]
+    books.sort(reverse = True)
+    bookshelves.sort(reverse = True)
+
+    used_bookshelves = []
+    while books and bookshelves and ((bookshelves[0] >= books[0]) or (used_bookshelves and used_bookshelves[0] >= books[0])):
```
gbudau, line 23: Placed parentheses wrong here, should be:
```suggestion
    while books and ((bookshelves and bookshelves[0] >= books[0]) or (used_bookshelves and used_bookshelves[0] >= books[0])):
```

clafoutis42:
You are embedding files from others in your pull request, please cleanup your branch if you want to get reviewed and merged. You have until tonight 20.42 pm.

gbudau:
Oops, sorry.
I think i managed to clean it up.

clafoutis42:
Perfect, I'll start reviewing it now 👌

chall04/gbudau.py
```diff
@@ -0,0 +1,94 @@
...
+
+def main():
+    # Input come from buffer and not direct input, accepting heredoc only
+    if not sys.stdin.isatty():
+        bookshelves, books = verify_and_save_file(sys.stdin, STDIN)
+        if (not bookshelves or not books):
+            read_error("stdin")
+        else:
+            fill_bookshelves("stdin", bookshelves, books)
+        sys.exit(0)
```
clafoutis42, line 70: Honestly, I won't consider this as an error, but I honestly would have prefered you let me passe files from stdin as a tty, as it was never asked to make such a protection.  
The heredoc approach was used in the example to make it more clear which part was the input, and which part was the output ;)

chall04/gbudau.py
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env python3
+
+import sys
+
+STDIN = 0
+FILE = 1
```
clafoutis42, line 6: This makes the code more readable, but you really could have used a boolean with an explicit names as parameters for your functions

chall04/gbudau.py
clafoutis42: It doesn't seem to cause any problem, or at least I'm not finding any error case.

chall04/gbudau.py
```diff
@@ -0,0 +1,94 @@
...
+    while books and bookshelves and ((bookshelves[0] >= books[0]) or (used_bookshelves and used_bookshelves[0] >= books[0])):
+        if  used_bookshelves and used_bookshelves[0] >= books[0]:
+            used_bookshelves[0] -= books[0]
+            books.pop(0)
+            used_bookshelves.sort(reverse = True)
+        else:
+            used_bookshelves.append(bookshelves[0] - books[0])
+            used_bookshelves.sort(reverse = True)
+            books.pop(0)
+            bookshelves.pop(0)
```
clafoutis42, line 32: That's an interesting algorithm. It helps you find a correct number that is not irrational.  
I'll let you look at the solution I'll push to understand better how you could have done the same without jungling between 2 lists.  
But I like that approach, however it could be simplified :) 

A little example of how you could have done it:
```python3
# Assuming your books and shelves are sorted in an decreasing order
def fill_bookshelves(books, shelves):
    used_shelves = 1
    shelves = iter(shelves)
    shelf = next(shelves)
    for book in books:
        if shelf >= book:
            shelf -= book
        else:
            try:
                shelf = next(shelves)
                used_shelves += 1
                if shelf < book:
                    break
            except StopIteration:
                break
    else:
        return used_shelves
```

chall04/gbudau.py
```diff
@@ -0,0 +1,94 @@
...
+def verify_and_save_file(arg, source):
+    bookshelves = []
+    books = []
+
+    # Save first line to bookshelves list from the file generator, striping spaces and newline
+    # Splitting the line into a list of numeric strings
+    # next() can return an error when the file is empty, without any newline so im using try/except
+    # Also checking the shelves to have non-zero values
+    try:
+        bookshelves = next(arg).rstrip().split()
```
clafoutis42, line 45: You do not need to use rstrip if you use split without parameters:
```suggestion
    bookshelves = next(arg).split()
```

chall04/gbudau.py
```diff
@@ -0,0 +1,94 @@
...
+        if source == STDIN:
+            read_error("stdin")
+            sys.exit(0)
+        else: return None, None 
+
+    # Save first word from each line starting from second line into books list
+    # Checking if they are numeric and not empty
+    for line in arg:
+        books += line.rstrip().split()[:1]
+    if not books or any(not book.isdigit() for book in books):
```
clafoutis42, line 58: Using `not all` would have been easier to read
```suggestion
    if not books or not all(book.isdigit() for book in books):
```

chall04/gbudau.py
```diff
@@ -0,0 +1,94 @@
...
+    bookshelves = []
+    books = []
+
+    # Save first line to bookshelves list from the file generator, striping spaces and newline
+    # Splitting the line into a list of numeric strings
+    # next() can return an error when the file is empty, without any newline so im using try/except
+    # Also checking the shelves to have non-zero values
+    try:
+        bookshelves = next(arg).rstrip().split()
+        if not bookshelves or any(not shelf.isdigit() or shelf[0] == '0' for shelf in bookshelves):
```
clafoutis42, line 46: You could have used `not all` instead of `any( not` here too.

chall04/gbudau.py
```diff
@@ -0,0 +1,94 @@
...
+def space_error(filename):
+    print("{}: {}: Not enough space in the given shelves".format(sys.argv[0], filename))
+
+def fill_bookshelves(filename, bookshelves, books):
+    # Convert books and bookshelves to integers
+    # Sort from max to min
+    books = [int(i) for i in books]
+    bookshelves = [int(i) for i in bookshelves]
+    books.sort(reverse = True)
+    bookshelves.sort(reverse = True)
```
clafoutis42, line 20: You can use the builtin `sorted` to do improve the readability:
```suggestion
    books = sorted([int(i) for i in books], reverse=True)
    bookshelves = sorted([int(i) for i in bookshelves], reverse=True)
```

clafoutis42 (commented):
Congratulations! You have the first, and the only working solution, and even if it can be improved, your code is good.

chall04/gbudau.py
gbudau: It outputs wrong answer in cases where the numbers of bookshelves is less than the number of books but there is enough space in those bookshelves.
```
1000
70 A Game of Thrones
76 A Clash of Kings
99 A Storm of Swords
75 A Feasts for Crows
```
With parentheses after `bookshelves` it will check if there are any `bookshelves` left but in this cases they have been moved to `used_bookshelves` so will return space error.

With parentheses before `bookshelves` it checks both `bookshelves` and `used_bookshelves` in case there is no "intact" `bookshelf` left, which is the way it should work.

chall04/gbudau.py
gbudau: My bad, took the examples too literal :D

chall04/gbudau.py
gbudau: Thank you, i didn't knew about this.

chall04/gbudau.py
gbudau: This looks so much better, thanks for the example.
I was thinking about a way to sort and convert at the same time or append and convert but didn't figure out how to do it.

chall04/gbudau.py
gbudau: Really interesting approach, thank you for sharing this. Looking forward to check out your solution.

gbudau:
Thank you very much for the feedback, appreciate it, i'm learning a lot!

PR #176, opened 2020-04-14
Solution for small cases by rarias-p

TH0RlN:
First of all I calculate all the possible combinations of shelves (this is the least efficient way and the reason why the program does not work for bigger cases). Then I got the length of the books and of each combination of shelves. I discard all the combinations with a shorter length than the books. Finally I get the combination with the smallest number of shelves.


aguimbao:
Please, revert your last 3 commits and make a new PR for challenge 05.

clafoutis42:
You are deleting files from others... You have until 20.42 to make a clean PR, or else we won't merge your code.

chall04/rarias-p.py
```diff
@@ -0,0 +1,91 @@
...
+						# object which has no lenth, that is the one causing 
+						# the problem here. This only means we are getting a
+						# non numeric value from the file.
+				
+	else:
+		try:
+			lines = []
+			for line in fileinput.input():
+				lines.append(line)
+				pass
```
aguimbao, line 84: Why `pass`?

chall04/rarias-p.py
```diff
@@ -0,0 +1,91 @@
...
+			lenths = checker(lines)
+			printer(argv, "stdin", lenths)
+		except TypeError:
+			printError("stdin", argv)
+
+if __name__ == '__main__':
+	main(sys.argv)
```
aguimbao, line 91: Why pass the `sys.argv` parameter when you can access it from anywhere?

chall04/rarias-p.py
```diff
@@ -0,0 +1,91 @@
...
+		possible = []
+		for bookshelf in power(shelves):
+			if listsum(bookshelf) > totalbooks:
+				possible.append(bookshelf)
+		lenths = []
+		for bookshelf in possible:
+			lenths.append(len(bookshelf))
+		return (lenths)
+	except ValueError:
+		pass			# This error is produces another error (line 79) which
```
aguimbao, line 55: Really bad practice to `pass` on an exception.

chall04/rarias-p.py
```diff
@@ -0,0 +1,91 @@
...
+						# come up with.
+
+def main(argv):
+	if len(argv) > 1:
+		fn = 1
+		for file in argv[1:]:
+			try:
+				f = open(argv[fn], "r")
+				lines = f.readlines()
+				lenths = checker(lines)
```
aguimbao, line 67: You mean `lengths`? 😜

chall04/rarias-p.py
```diff
@@ -0,0 +1,91 @@
...
+				
+	else:
+		try:
+			lines = []
+			for line in fileinput.input():
+				lines.append(line)
+				pass
+			lenths = checker(lines)
+			printer(argv, "stdin", lenths)
+		except TypeError:
```
aguimbao, line 87: You should try to catch `EOFError` too for safety incase I `Ctrl-D` your script.

aguimbao (commented):
I don't think it's a good idea to calculate all the possible combination of shelves. You could just calculate the "biggest" space with the smallest number of shelves, and if you don't have enough space, keep adding the next biggest shelf.

Anyway, you are deleting other people's files in your PR, you are also submitting your previous challenges, and worst of all, your file didn't have execution permissions (644) :( I will have to close it.

PR #178, opened 2020-04-14
Fit_books! 

bmk101:
Prometeo chall04. No merged! Checking if still available


clafoutis42:
You pushed your code too late, and didn't even try to explain it. I'll close this PR and we won't review it.

PR #179, opened 2020-04-16
vde-dios.c

vdedios:
**SOLUTION**

The solution is simple, following the rule **_"I before E except after C"_** the program walks through the input string. Whenever it encounters an "e" char, it compares if the preceding and the next characters are set up correctly following the rule.

### DIFF
--- a/chall05/vde-dios.c
+++ b/chall05/vde-dios.c
@@ -1,0 +1,19 @@
+int ft_ie_except_after_c(char* str)
+{
+    if (str)
+    {
+        while (*str)
+        {
+            if (*str == 'e' && *(str + 1))
+            {
+                if (*(str - 1) == 'i' && *(str + 1) == 'c')
+                    return (0);
+                if (*(str + 1) == 'i' && *(str + 2) != 'c')
+                    return (0);
+            }
+            str++;
+        }
+        return (1);
+    }
+    return (0);
+}
--- a/chall05/vde-dios.c
+++ b/chall05/vde-dios.c
@@ -1,15 +1,20 @@
 int ft_ie_except_after_c(char* str)
 {
+    char *p;
+
     if (str)
     {
+        p = str;
         while (*str)
         {
-            if (*str == 'e' && *(str + 1))
+            if (*str == 'e')
             {
-                if (*(str - 1) == 'i' && *(str - 2) == 'c')
-                    return (0);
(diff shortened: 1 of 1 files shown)


chall05/vde-dios.c
```diff
@@ -0,0 +1,31 @@
+char   ft_lower(char letter)
+{
+    if (letter >= 'A' && letter <= 'Z')
+        return (letter + 32);
+    return (letter);
+}
```
clafoutis42, line 6: It's nice that you thought about that! However, I believe it wasn't explicit in the subject but you could assume that the word passed was already a lowercased valid word.

chall05/vde-dios.c
```diff
@@ -0,0 +1,31 @@
...
+
+int ft_ie_except_after_c(char* str)
+{
+    char *p;
+
+    if (str)
```
clafoutis42, line 12: It is my opinion, but you don't have to protect from `NULL` when making such a function. A function is not like a program, a function CAN crash. Let me explain myself:  

chall05/vde-dios.c
```diff
@@ -0,0 +1,31 @@
...
+            if (ft_lower(*str) == 'e')
+            {
+                if (str != p && (str - 1)!= p)
+                    if (ft_lower(*(str - 1)) == 'i' && ft_lower(*(str - 2)) == 'c')
+                        return (0);
+                if (*(str + 1))
+                    if (ft_lower(*(str + 1)) == 'i' && (str == p || ft_lower(*(str - 1)) != 'c'))
+                        return (0);
+            }
+            str++;
```
clafoutis42, line 26: Having to use your `ft_lower` function prevents you from doing so, but you could have used `strstr` to look for the wanted occurrences. For example you know that you should never find `cie`, this could be solved in an `if`.  

clafoutis42 (commented):
Nice solution, it seems to work. It could still be improved, but it solves the problem.  
Sorry for the specifying that you could assume the words were already lowercased 😅

chall05/vde-dios.c
vdedios: That is true, however I tried to resolve the problem without using external functions since the problem seemed simple enough for making it with simple conditionals. I do not know if using strstr would lead to a performance improvement rather than having a more readable code (that is true)

PR #180, opened 2020-04-16
fgata-va's solution

FranciscoJesusGata:
# How do I solved it
I search for 'i' and 'e' inside of the string and if the next character is an 'i' or an 'e', then if I find "ei" and the previous char it's not a 'c' or the 'e' it's the first char return 0. If I find "ei" and the previous char it's 'c' and 'e' it's not the first char returns 0. If I find neither of them I simply return 1.


chall05/fgata-va.c
```diff
@@ -0,0 +1,18 @@
...
+	while (str[i] != '\0')
+	{
+		if (str[i] == 'e' || str[i] == 'i')
+		{
+			if (str[i] == 'i' && str[i + 1] == 'e' && (i > 0 && str[i - 1] == 'c'))
+				return (0);
+			else if (str[i] == 'e' && str[i + 1] == 'i' && (i == 0 || str[i - 1] != 'c'))
+				return (0);
+		}
+		i++;
```
clafoutis42, line 15: You could have used `strstr` to iterate and the different character combination you want to check!  
You would benefit from its optimizations, and would have gained time and readability in your code. I know that in 42's project you are encouraged to recode everything, but in this challenges you are allowed to take full advantage of the language imposed (as long as it remains in standard libraries).

clafoutis42 (commented):
Good solution, it works. It sill might be improved but good job!

chall05/fgata-va.c
FranciscoJesusGata: You're right I have to get used to use more C built-in functions

PR #181, opened 2020-04-16
Commit solution egarcia-

byrd92:
I check if there is a 'c' for the rule cie, if its true, return 0.
also i check if there is a 'e', followed by a 'i ' and check back if is not a 'c' the word before, if thats happen, i return 0.
otherwise I return 1


chall05/egarcia-.c
```diff
@@ -0,0 +1,30 @@
+#include <stdio.h>
+
+#define funtion(x)       printf(x" = %d\n", ft_ie_except_after_c(x));
```
clafoutis42, line 3: You could have removed your debugging lines

chall05/egarcia-.c
```diff
@@ -0,0 +1,30 @@
...
+
+int     ft_ie_except_after_c(char* str)
+{
+    while (*str)
+    {
+        if (*(str + 2) && *str == 'c' && *(str + 1) == 'i' && *(str + 2) == 'e')
+            return(0);
+        else if (*(str + 1) && *str == 'e' && *(str + 1) == 'i' && *(str - 1) != 'c')
+            return 0;
+        str++;
```
clafoutis42, line 13: You could have used `strstr` to iterate and the different character combination you want to check!  
You would benefit from its optimizations, and would have gained time and readability in your code. I know that in 42's project you are encouraged to recode everything, but in this challenges you are allowed to take full advantage of the language imposed (as long as it remains in standard libraries).

chall05/egarcia-.c
```diff
@@ -0,0 +1,30 @@
...
+#define funtion(x)       printf(x" = %d\n", ft_ie_except_after_c(x));
+
+int     ft_ie_except_after_c(char* str)
+{
+    while (*str)
+    {
+        if (*(str + 2) && *str == 'c' && *(str + 1) == 'i' && *(str + 2) == 'e')
+            return(0);
+        else if (*(str + 1) && *str == 'e' && *(str + 1) == 'i' && *(str - 1) != 'c')
+            return 0;
```
clafoutis42, line 12: I know there is no norm, but try to keep an integrity in your code ;)

chall05/egarcia-.c
```diff
@@ -0,0 +1,30 @@
+#include <stdio.h>
+
+#define funtion(x)       printf(x" = %d\n", ft_ie_except_after_c(x));
+
+int     ft_ie_except_after_c(char* str)
+{
+    while (*str)
+    {
+        if (*(str + 2) && *str == 'c' && *(str + 1) == 'i' && *(str + 2) == 'e')
```
clafoutis42, line 9: Watch out... Here you will be reading after the end of the string, it will crash, and it will be your fault :(

clafoutis42 (commented):
Your solution is working, but is crashing on sanitized memory as it tries to dereference a the `str + 2` which on the character before last will cause it to read outside of the passed string.

PR #182, opened 2020-04-16
MONGO

victorsempere:
If null I consider neither ei nor ie will be found so 1 will be returned
I look for ei and ie. If some founds I apply the rules and if not matches the exceptions are checked.
If not found neither ei nor ie 1 will be returned

### DIFF
--- a/chall05/vsempere.c
+++ b/chall05/vsempere.c
@@ -1,0 +1,72 @@
+# include <stdlib.h>
+
+static size_t		ft_strlen(const char *s)
+{
+	size_t	i;
+
+	i = 0;
+	while (s[i] != 0)
+		i++;
+	return (i);
+}
+
+static char		*ft_strstr(const char *big, const char *little)
+{
+	size_t		i;
+	size_t		j;
+
+	i = 0;
+	j = 0;
+	while (big[i])
+	{
+		while (big[i + j] == little[j] && big[i + j])
+			j++;
+		if (little[j] == '\0')
+			return ((char *)(big + i));
+		j = 0;
+		i++;
+	}
+	return (NULL);
+}
+
+static char ft_check_exception(char *str)
+{
+	char	*exc;
+	char	*f;
+	size_t	strl;
+
+	exc = "sleigh stein fahrenheit deifies either nuclei reimburse ancient juicier societies";
+	if (!(f = ft_strstr(exc, str)))
+		return (0);
+	strl = ft_strlen(str);
+	return (((f > exc && *(f - 1) == ' ') || f == str) && (f[strl] == '\0' || f[strl] == ' ') ? 1 : 0);
+}
+
+int			ft_ie_except_after_c(char* str)
+{
+	char	*f;
+	
+	if (!str)
+		return (1);
+	if ((f = ft_strstr(str, "ie")))
+	{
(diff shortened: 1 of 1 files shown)


chall05/vsempere.c
```diff
@@ -0,0 +1,61 @@
...
+	strl = ft_strlen(str);
+	return (((f > exc && *(f - 1) == ' ') || f == str) && (f[strl] == '\0' || f[strl] == ' ') ? 1 : 0);
+}
+
+int			ft_ie_except_after_c(char* str)
+{
+	char	*f;
+	
+	if (!str)
+		return (1);
```
clafoutis42, line 50: It is my opinion, but you don't have to protect from `NULL` when making such a function. A function is not like a program, a function CAN crash. Let me explain myself:  

chall05/vsempere.c
```diff
@@ -0,0 +1,61 @@
...
+	char	*exc;
+	char	*f;
+	size_t	strl;
+
+	exc = "sleigh stein fahrenheit deifies either nuclei reimburse ancient juicier societies";
+	if (!(f = ft_strstr(exc, str)))
+		return (0);
+	strl = ft_strlen(str);
+	return (((f > exc && *(f - 1) == ' ') || f == str) && (f[strl] == '\0' || f[strl] == ' ') ? 1 : 0);
+}
```
clafoutis42, line 43: Mmmh the exercice is loosing its point if you return true when an exception is present.  
You are also forgetting most of the exceptions... what about `eight`, for example?

By using this function, you are failing the chall 😅

chall05/vsempere.c
```diff
@@ -0,0 +1,61 @@
...
+
+int			ft_ie_except_after_c(char* str)
+{
+	char	*f;
+	
+	if (!str)
+		return (1);
+	if ((f = ft_strstr(str, "ie")))
```
clafoutis42, line 51: What if there is `ie` and then `cie` in a word?

chall05/vsempere.c
```diff
@@ -0,0 +1,61 @@
...
+}
+
+static char ft_check_exception(char *str)
+{
+	char	*exc;
+	char	*f;
+	size_t	strl;
+
+	exc = "sleigh stein fahrenheit deifies either nuclei reimburse ancient juicier societies";
+	if (!(f = ft_strstr(exc, str)))
```
clafoutis42, line 39: Well, `cier` would be considered an exception as it is part of the string 😅

chall05/vsempere.c
```diff
@@ -0,0 +1,61 @@
...
+	{
+		while (big[i + j] == little[j] && big[i + j])
+			j++;
+		if (little[j] == '\0')
+			return ((char *)(big + i));
+		j = 0;
+		i++;
+	}
+	return (NULL);
+}
```
clafoutis42, line 30: You could use the standard library, in those challs, you are encouraged to take advantage of the whole standard library, you already have the cursus' projects to learn about how they work 😉

clafoutis42 (commented):
The objective of this challenge would be to differentiate exceptions and words that follows the rule. Therefore returning `1` when you meet an exception is a little odd. Specifically as you are just checking for 8 exceptions, there are a lot more exceptions to this rule.
Also you are only checking the first occurrence, there might be more!

PR #183, opened 2020-04-16
chall05-jbaringo

baringo11:
First I check if the string exists, then I create a counter for going over the string and when I find the letter 'e' I check if there is a 'i' after and a 'c' before it like "cei", if there is no 'c' it return 0. The other exception checks if there is a 'e' after a 'i' and if there is a 'c' before the 'i' like "cie" returns 0.
If this exceptions don't happen it returns 1.


aguimbao (commented):
It's a nice simple approach, but for this simple exercise I would have liked to see some standard lib usage like `strstr()` or some pointer arithmetic. 

You go out of bounds for `str = "ie"` or `str = "i"`. Even if your code doesn't crash, it's still pretty dangerous. You should sanitise your addresses and debug your code more thoroughly ! Even for simple functions !

PR #184, opened 2020-04-16
Jtrancos - Chall05

jtrancoso:
To make it simpler we only look for the instances in which the string fails the norm, which are two cases:

1. If 'ei' is not preceded by 'c'.
2. If 'ie' is preceded by 'c'.

Using `i` we go through the string and if we find 'e' or 'i' we check the next position and the previous one, if they fail the rule we return '0'.

For every other case we return '1'.

### DIFF
--- a/chall05/jtrancos.c
+++ b/chall05/jtrancos.c
@@ -1,0 +1,36 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   jtrancos.c                                         :+:      :+:    :+:   */
+/*                                                    +:+ +:+         +:+     */
+/*   By: jtrancos <marvin@42.fr>                    +#+  +:+       +#+        */
+/*                                                +#+#+#+#+#+   +#+           */
+/*   Created: 2020/04/16 13:01:08 by jtrancos          #+#    #+#             */
+/*   Updated: 2020/04/16 13:31:06 by jtrancos         ###   ########.fr       */
+/*                                                                            */
+/* ************************************************************************** */
+
+#include <stdio.h>
+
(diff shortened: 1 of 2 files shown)


chall05/jtrancos.c
```diff
@@ -0,0 +1,29 @@
...
+#include <stdio.h>
+
+int	ft_ie_except_after_c(char* str)
+{
+	int i;
+
+	i = 0;
+	while (str[i])
+	{
+		if (str[i] == 'e' && str[i + 1] == 'i' && str[i - 1] != 'c')
```
clafoutis42, line 22: What if `e` is the first character? You'll just go out of the string and crash  
You should try to compile with the flag `-fsanitize=address` to be more aware of such problems.

chall05/jtrancos.c
```diff
@@ -0,0 +1,29 @@
...
+			return (0);
+		if (str[i] == 'i' && str[i + 1] == 'e' && str[i - 1] == 'c')
+			return (0);
+		i++;
+	}
```
clafoutis42, line 27: You could have used `strstr` to iterate on the character combination ;)

clafoutis42 (commented):
You seem to have understood the problem, and the solution in its logic is working. But you are reading outside of the string's reserved memory, therefore you're crashing.

chall05/jtrancos.c
jtrancoso: It could be solved by adding the condition `if (i != 0)` before both `if`, right?

chall05/jtrancos.c
jtrancoso: Thank you for the suggestion and the evaluation :)

PR #185, opened 2020-04-16
edramire - Mongo

edramir18:
To avoid confusion in case the string contains upper and lowercase character, first make a copy with all character in lowercase.
When i find a 'i' character in the string, i check before and after for an 'e' character and based if the preceding letter is 'c', return one or zero according to the rule i before e. In case i never find a letter 'i' i return 1 as result.

### DIFF
--- a/chall05/edramire.c
+++ b/chall05/edramire.c
@@ -1,0 +1,56 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   edramire.c                                         :+:      :+:    :+:   */
+/*                                                    +:+ +:+         +:+     */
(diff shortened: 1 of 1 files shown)


chall05/edramire.c
```diff
@@ -0,0 +1,56 @@
...
+{
+	size_t	i;
+	size_t	len;
+	int		resp;
+	char	*aux;
+
+	if (str == NULL)
+		return (1);
+	len = strlen(str);
+	aux = ft_to_lower(str, len);
```
anna-dominguez, line 39: It's a good idea but you could do a comparison of characters instead of doing a `malloc`. I don't think it's necessary to do one.

anna-dominguez (approved):
I have an error with this one `iecie` or `ebeforeiexceptafterc`. It should return `0` and you return `1`. You're close to the solution!

PR #186, opened 2020-04-16
first try

bazuara:
My solution goes through the string, provided that is three or more characters long, searching for an 'i' char. When found, it checks the i before e rule, except after c, and if it does, returns 1.

Base case is 1 because if there are no 'i' the rule is correct


migferna42:
Please comment in English ;)

chall05/bazuara.c
```diff
@@ -0,0 +1,50 @@
...
+
+	i = 0;
+	if (s == NULL)
+		return (i);
+	while (s[i] != '\0')
+	{
+		i++;
+	}
+	return (i);
+}
```
clafoutis42, line 27: You are allowed to use the standard library in the challenges

clafoutis42, line 27: You are allowed to use the standard library in the challenges

chall05/bazuara.c
```diff
@@ -0,0 +1,50 @@
...
+	if (len < 3)
+		return (1);
+	while (i < len)
+	{
+		if (str[i] == 'i')
+		{
+			if ((str[i + 1] == 'e') && (str[i - 1] != 'c'))
+				return (1);
+		}
+		if (str[i - 2] == 'c' && str[i - 1] == 'e')
```
clafoutis42, line 45: You will be reading outside of the string's reserved memory. Therefore you are most likely to cause a segfault here (and it will in my tests).

chall05/bazuara.c
```diff
@@ -0,0 +1,50 @@
...
+			return (1);
+		i++;
+	}
```
clafoutis42, line 48: A cleaner solution would have been to use `strstr` to iterate on the character combinations present

clafoutis42 (commented):
The logic is correct, but the implementation will cause a segfault as you will be reading outside of the string's reserved memory :(

PR #187, opened 2020-04-16
Create szavarce.c

szavarce42:
I created a function to check for a string inside another string which is called to check for specific rules.

### DIFF
--- a/chall05/szavarce.c
+++ b/chall05/szavarce.c
@@ -1,0 +1,42 @@
+int ft_found(char *find, char *in_str)
+{
+    int pos;
+    int posb;
+    int ret;
+
+    pos = 0;
+    posb = 0;
+    ret = 1;
+    while (in_str[pos] != '\0')
+    {
+        posb = 0;
+        ret = 1;
+        while (find[posb])
+        {
+            if (find[posb] != in_str[pos + posb])
+                ret = 0;
+            posb++;
+        }
+        if (ret == 1)
+            return (1);
+        pos++;
+    }
+    return (0);
+}
+
+int ft_ie_except_after_c(char* str)
+{
+    if (!str[0])
+        return (0);
+    if (ft_found("ei",str) && ft_found("cei",str))
+        return (1);
+    if (ft_found("ei",str) && !ft_found("cei",str))
+        return (0);        
+    if (ft_found("ie",str) && !ft_found("cie",str))
+        return (1);
+    if (ft_found("ie",str) && ft_found("cie",str))
+        return (0);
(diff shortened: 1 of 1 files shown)


chall05/szavarce.c
```diff
@@ -0,0 +1,42 @@
...
+        return (1);
+    if (ft_found("ei",str) && !ft_found("cei",str))
+        return (0);        
+    if (ft_found("ie",str) && !ft_found("cie",str))
+        return (1);
+    if (ft_found("ie",str) && ft_found("cie",str))
+        return (0);
+    if (!ft_found("ie",str) && !ft_found("ei",str))
+        return (1);
+    return (0);
```
clafoutis42, line 41: Given that all preceding conditions weren't fulfilled, this one will always be true here
```suggestion
    return (1);
```

chall05/szavarce.c
```diff
@@ -0,0 +1,42 @@
...
+    if (ft_found("ei",str) && ft_found("cei",str))
+        return (1);
+    if (ft_found("ei",str) && !ft_found("cei",str))
+        return (0);        
+    if (ft_found("ie",str) && !ft_found("cie",str))
+        return (1);
+    if (ft_found("ie",str) && ft_found("cie",str))
+        return (0);
+    if (!ft_found("ie",str) && !ft_found("ei",str))
+        return (1);
```
clafoutis42, line 40: You are only checking for the first occurence of each one of those, meaning that if I pass `eicei` you will return `1` when you should return `0`.

chall05/szavarce.c
```diff
@@ -0,0 +1,42 @@
...
+            if (find[posb] != in_str[pos + posb])
+                ret = 0;
+            posb++;
+        }
+        if (ret == 1)
+            return (1);
+        pos++;
+    }
+    return (0);
+}
```
clafoutis42, line 25: You could have used the standard library's `strstr` ;)

chall05/szavarce.c
```diff
@@ -0,0 +1,42 @@
...
+    while (in_str[pos] != '\0')
+    {
+        posb = 0;
+        ret = 1;
+        while (find[posb])
+        {
+            if (find[posb] != in_str[pos + posb])
+                ret = 0;
+            posb++;
+        }
```
clafoutis42, line 19: Here you will end up reading outside of `in_str` reserved memory. You should check for `in_str`'s end too! You will crash in my tests.

clafoutis42 (commented):
The solution only checks for the first occurrence of each combination. It should check more.  
Also there is a segfault in your solution, too bad!

PR #189, opened 2020-04-16
antgonza

ToniGP:
I have solved the problem using the strnstr function from my library

### DIFF
--- a/chall05/antgonza.c
+++ b/chall05/antgonza.c
@@ -1,0 +1,67 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   antgonza.c                                         :+:      :+:    :+:   */
+/*                                                    +:+ +:+         +:+     */
+/*   By: antgonza <marvin@42.fr>                    +#+  +:+       +#+        */
+/*                                                +#+#+#+#+#+   +#+           */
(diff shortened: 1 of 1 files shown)


chall05/antgonza.c
```diff
@@ -0,0 +1,67 @@
...
+					&& needle[j] != '\0' && (i + j) < len)
+			j++;
+		if (needle[j] == '\0')
+			return ((char *)&haystack[i]);
+		else
+			j = 0;
+		i++;
+	}
+	return (NULL);
+}
```
clafoutis42, line 51: Why not use the one's from the standard library?

chall05/antgonza.c
```diff
@@ -0,0 +1,67 @@
...
+
+	i = ft_strlen(str);
+	if ((!(ft_strnstr(str, "ei", i)) && (!(ft_strnstr(str, "ie", i)))))
+		return (1);
+	if (ft_strnstr(str, "cei", i) || (ft_strnstr(str, "ie", i)))
+	{
+		if (ft_strnstr(str, "cie", i))
+			return (0);
+		return (1);
+	}
```
clafoutis42, line 65: Too bad, you're only checking for the first occurence of each combination :(

clafoutis42 (commented):
Too bad, you're only checking for the first occurrence of each combination. Giving `eicei` your code will return `1` when it should return `0`.  
Better luck next time!

PR #190, opened 2020-04-16
Update Permission

szavarce42:


chall05/szavarce.c
```diff
@@ -0,0 +1,42 @@
...
+        return (1);
+    if (ft_found("ei",str) && !ft_found("cei",str))
+        return (0);        
+    if (ft_found("ie",str) && !ft_found("cie",str))
+        return (1);
+    if (ft_found("ie",str) && ft_found("cie",str))
+        return (0);
+    if (!ft_found("ie",str) && !ft_found("ei",str))
+        return (1);
+    return (0);
```
aguimbao, line 41: You could have grouped this conditions, but you could have also used `strstr()` instead of making your own function.

chall05/szavarce.c
```diff
@@ -0,0 +1,42 @@
...
+    pos = 0;
+    posb = 0;
+    ret = 1;
+    while (in_str[pos] != '\0')
+    {
+        posb = 0;
+        ret = 1;
+        while (find[posb])
+        {
+            if (find[posb] != in_str[pos + posb])
```
aguimbao, line 16: You go out of bounds here for `str = "ie"`. This could cause a crash. You need to check your bounds.

chall05/szavarce.c
```diff
@@ -0,0 +1,42 @@
...
+                ret = 0;
```
aguimbao, line 17: Why don't you `return (1);` here ?

aguimbao (commented):
I think you could have thought of a simpler approach !

You also fail for `str = "eicei"`.

Also, there was no comment in your PR explaining your code, so I'm not gonna merge it. Please explain your PR in english next time, it's good practise for you and it helps me understand too 😄

PR #191, opened 2020-04-16
estina : chall05

xplore93:
It is my solution for the Chall05.
The idea is:

- Check if the string exists and is not null.
- For each character, check if it's an *i* and the next is *e* or it's an *e* and the next is *i*.
-- Then check if it's not the first character of the word.
- Call a function that returns the lower character for the anterior one and then call a function and check if the anterior was a *c*.
- Then return the value of *rule* variable:
-- If there are not any *ie* nor *ei*, then it will return 1.
-- In other case it will return 1 or 0


xplore93:
Fixed that if at some moment the rule is 0, stop the loop.

chall05/estina.c
```diff
@@ -0,0 +1,47 @@
...
+/*                                                                            */
+/* ************************************************************************** */
+
+static char	ft_lower(char c)
+{
+	if (c >= 'A' && c <= 'Z')
+		return (c + 32);
+	else
+		return (c);
+}
```
clafoutis42, line 19: It's nice that you thought about that! However, I believe it wasn't explicit in the subject but you could assume that the word passed was already a lowercased valid word.

chall05/estina.c
```diff
@@ -0,0 +1,47 @@
...
+
+}
+
+int			ft_ie_except_after_c(char* str)
+{
+	int		i;
+	int		rule;
+
+	if (!str || !*str)
+		return (0);
```
clafoutis42, line 35: It is my opinion, but you don't have to protect from `NULL` when making such a function. A function is not like a program, a function CAN crash. Let me explain myself:  

chall05/estina.c
```diff
@@ -0,0 +1,47 @@
...
+	i = 0;
+	rule = 2;
+	while (str[i] && rule)
+	{
+		if (ft_lower(str[i]) == 'e' && ft_lower(str[i + 1]) == 'i' && i)
+			rule = ft_check_c(ft_lower(str[i - 1]), 1);
+		if (ft_lower(str[i]) == 'i' && ft_lower(str[i + 1]) == 'e')
+			rule = i ? ft_check_c(ft_lower(str[i - 1]), 0) : 1;
+		i++;
```
clafoutis42, line 44: Having to use your `ft_lower` function prevents you from doing so, but you could have used `strstr` to look for the wanted occurrences. For example you know that you should never find `cie`, this could be solved in an `if`.  

chall05/estina.c
```diff
@@ -0,0 +1,47 @@
...
+	}
+	return (rule == 2 ? 1 : rule);
```
clafoutis42, line 46: It's a little diffcult to understand the meaning of rule here, a more readable approach would have been to return `0` in your loop on meeting a bad combination, and returning `1` if you go through the whole loop.

chall05/estina.c
```diff
@@ -0,0 +1,47 @@
...
+	int		i;
+	int		rule;
+
+	if (!str || !*str)
+		return (0);
+	i = 0;
+	rule = 2;
+	while (str[i] && rule)
+	{
+		if (ft_lower(str[i]) == 'e' && ft_lower(str[i + 1]) == 'i' && i)
```
clafoutis42, line 40: This condition protects you from a crash but keeps you from verificating if a word that starts with `ei` is good or bad.  
In fact you should return `0` if a word starts with `ei` as it would not be prefixed by `c`.  
Test this case on `ei` (or `either`, `eight`... If you want to test it with real words.

clafoutis42 (commented):
Nice code. However your solution fails if the word starts with `ei`, that's unfortunate :(  
Better luck next time!

chall05/estina.c
xplore93: I was too preventive ;)

chall05/estina.c
xplore93: 42 teached me that I should care about this in functions also ;)

chall05/estina.c
xplore93: I thing it escaped from me as for ie I check that if it's beginning

xplore93: And the use of strstr would make the things easier, smart solution!

chall05/estina.c
xplore93: Surely will be much easier once found the fail of the rule, return 0

xplore93:
Thank you Charles, as always, you showed me the way to better solutions. Appreciate it.

chall05/estina.c
clafoutis42: If I remember well, the only functions that need to care about this, are the chained list functions, and there's a reason behind that, it's because `NULL` is considered as an empty list.  
Otherwise, if you take the standard library for example, try passing `NULL` to `strstr`, it will crash ;)

PR #192, opened 2020-04-16
chall05 Antomart

miyomovil:
To solve this challenge I just create a loop where i go along the string finding if the string has the rule requirements.  
If the string has 'i' 'e' and not a 'c'  returns 1. it  also returns 1 for all the cases where not the rule and the excepcion is present. When the rule is not acomplished it returns 0. The function controls uppercases at the start of the word.


chall05/antomart.c
```diff
@@ -0,0 +1,35 @@
...
+#include <stdio.h>
+ 
+int ft_ie_except_after_c(char* str)
+{
+    int i;
+
+    i = 0;
+    while (str[i])
+    {
+        if (((str[i] == 'i' || str[i] == 'I') && (str[i + 1] == 'e' )) || ((str[i] == 'c' || str[i] == 'C') && (str[i + 1] == 'e') && (str[i + 2] == 'i')))
```
anna-dominguez, line 23: Why doing some checks with upper case and other without ? Your conditions are very complex for what it is asked you could simplify a lot

anna-dominguez (commented):
You have an error with this test `iecie` that returns 1 and should be returning 0 for example. You can simplify a lot your code

PR #193, opened 2020-04-16
Spell by the spelling - ssacrist

esepunto:
****IMPORTANT:** _This comment is for the fisrt version. To read the comment of the last version, please, see the last comment. Sorry for the inconvenience._**

The program assumes that the word follows the rule (== 1), unless it finds the following cases:

- If "ei" appears in a word, it must be preceded by "c".
- If "ie" appears in a word, it must not be preceded by "c".

To do this, the program makes two checks:

Check 1.- It runs through the string and if it finds an "e", it checks:
- [a] If the next character is an "i".  
- [b] If the above character is not a "c" or the "e" is the beginning of the word.
If both conditions are met (a. and b.), the word does NOT follow the rule (== 0).

Check 2.- Go through the string and if it finds an "i", check it:
- [a] If the next character is an "e".
- [b] If the previous character is a "c".
If both conditions are met (a. and b.), the word does NOT follow the rule (== 0).

### DIFF
--- a/chall05/ssacrist.c
+++ b/chall05/ssacrist.c
@@ -1,28 +1,26 @@
-#include <string.h>
+#include<ctype.h>
 
-int ft_ie_except_after_c(char* str)
+int     ft_ie_except_after_c(char* str)
 {
-    int  len;
-    int  result;
-    int  i;
-    
-    len = strlen(str);
+    int     result;
+    int     i;
+
     i = 0;
     result = 1;
-    while (i < len)
+    while (str[i])
     {
-        while (str[i] == 'e')
+        while (tolower(str[i]) == 'e')
         {
-            while (str[i + 1] == 'i' && (str[i - 1] != 'c' || !str[i - 1]))
+            while (tolower(str[i + 1]) == 'i' && (tolower(str[i - 1]) != 'c' || !str[i - 1]))
             {    
                 result = 0;
                 i++;
             }
             i++;
         }
-        while (str[i] == 'i')
+        while (tolower(str[i]) == 'i')
         {
-            while (str[i + 1] == 'e' && str[i - 1] == 'c')
+            while (tolower(str[i + 1]) == 'e' && tolower(str[i - 1]) == 'c')
             {    
                 result = 0;
                 i++;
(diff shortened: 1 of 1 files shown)


esepunto:
I found many _errors_ (approachs for improvement) in my code for this challenge: fisrt my code does not consider capital letters, then the use of 'while', and the variable called 'result' is not necessary... 

Well, I corrected it but... _Looks like I picked the wrong week to code..._

esepunto:
**Updated comment.** _For the last version of the code._ 

**Rules:** 
If "ei" appears in a word, it must be inmediately follow "c".
If "ie" appears in a word, it must not immediately follow "c".
A word also follows the rule if neither "ei" nor "ie" appears anywhere in the word.

**Chall:**
Write a function that tells you whether or not a given word follows the "I before E except after C" rule. 
Return 1 if the word follows the rule.
Return 0 if the word does not follow the rule.

**Solution**
The function runs through the chain and makes consecutive checks to find exceptions to the rule, and considers upper and lower case.

Check 1.- If it finds an "e", checks:
[1.a] If the next character is an "i".
[1.b] If the previous character is not a "c" or the "e" is the beginning of the word.
If conditions are met, the word does NOT follow the rule, and returns 0.

Check 2.- If it finds an "i", checks:
[2.a] If the next character is an "e".
[2.b] If the previous character is a "c".
If conditions are met, the word does NOT follow the rule, and returns 0.

If the function reaches the end of the string without any of the exceptions being met, it returns 1.

anna-dominguez (commented):
I have a fail on the first test I have 😞You should use `*(str - 1)` very carefully. When my string is `ie` or starts with `ie` you're going out of it. But this is a very good start of code, you should practice having this in mind with those examples!

esepunto:
Many thanks, Anna! I've a question. I would appreciate if you could help me again.

In the tests I've performed, the solution of my function is indeed wrong (= 0) when the word is "ie" or begins with "ie". But the result is correct when the word is "ei" or begins with "ei" (and I've identificated my error).

I compile with `gcc -Wall -Wextra -Werror` and work with ubuntu installed following your instructions: _For those who can't, or don't want to install that VM, you can use `xubuntu 18.04 LTS` and install the packages listed here: https://pastebin.com/4HnnSUpe_

Do you know why the word "ei" might be wrong when you run it from your computer?

Thanks!!

PR #194, opened 2020-04-16
Challenge 05 - cdorta-n

Top64:
Hello!

For this challenge, we need to check a "string", where the problem will be resolved. This time, the solution require read 2 letters, "e" and "i", check their positions, and the letter behind them, in this case, the letter "c".

Solution Base:

Look for the letter "e" -->  Look for the"i" next to it   --> if it's true, check if there is a "c" behind.
Look for the letter "i" --> Look for the "e" next o it --> if it's true, check if there isn't a "c" behind.

The default return is 1 ->> Because, is a word have 1 letter, we don't need to check it. 

An a alternative version only check for "direct errors" where those "solutions" aren't true.


chall05/cdorta-n.c
```diff
@@ -0,0 +1,33 @@
...
+int		ft_ie_except_after_c(char *str)
+{
+	int c;
+
+	c = 0;
+	while (str[c + 1] != '\0')
+	{
+		if (str[c] == 'e' && str[c + 1] == 'i')
+		{
+			if (str[c - 1] != 'c')
```
anna-dominguez, line 22: 😱

chall05/cdorta-n.c
```diff
@@ -0,0 +1,33 @@
...
+				return (0);
+		}
+		if (str[c] == 'i' && str[c + 1] == 'e')
+		{
+			if (str[c - 1] == 'c')
```
anna-dominguez, line 27: 😱

anna-dominguez:
Oupsy @abguimba 😆 
You have a start that you could simplify. But you're doing the same error as other of your pairs. Here when I'm testing with the string `ie` you're trying to access something before the start of the string with `str[c - 1]` that a very bad behavior if you are at the beginning of your string. You should practice having this in mind 😄

PR #195, opened 2020-04-16
chall05: dperez.

dperez42:
Check every character in the given string.
If the string is null return(1), the rule is ok?
If a "e" is found, check if the next character is a "i" if so, look for the character before the "e" is a "c" is so the rule is ok, and return(1).

if a "i" is found, check if the next character is a "e" if so look for the character before a "i" is not a "c" is so the rule is ok, and return(1), if it is equal to "c" return(0).

### DIFF
--- a/chall05/dperez-z.c
+++ b/chall05/dperez-z.c
@@ -1,15 +1,6 @@
 #include <unistd.h>
 #include <stdio.h>
 
-int ft_strlen(char *str)
-{
-	int cont;
-	cont = 0;
-	while(str[cont]!='\0')
-		cont++;
-	return(cont);
-}
-
 int ft_ie_except_after_c(char *str)
 {
 	int i;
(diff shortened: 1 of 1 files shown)


chall05/dperez-z.c
```diff
@@ -0,0 +1,36 @@
...
+					return(1);
+				else 
+					return(0);
+			}
+		}
+		if (str[i]=='i')
+		{
+			if (str[i+1]=='e' && str[i+1]!='\0')
+			{
+				if (str[i-1]!='c' && (i-1)>=0)
```
aguimbao, line 27: This is really dangerous. You go out of bounds for `str = "ie"` for example. You should always check your bounds.

aguimbao (commented):
This is really subjective but for me it's clearer when you write `while (str[i] != '\0')` than when you write `while (str[i]!='\0')`.

I think it's a nice approach, but you should always be careful with your str bounds.

dperez42:
Thanks for your comments, I will be careful next time.

PR #196, opened 2020-04-16
Chall05-fulldemo

fulldemo13:
Hi in chall05, I search the character 'i' and I check if te rule is true or false.
When I find a 'i' if after have a 'e' and before have a 'c' the rule isn't true (cie), and return 0.
When I find a 'i' and before have something different to "ce" , I return 0 because the rule is false.
In the other cases the rule always be true, and it returns 1.

### DIFF
--- a/chall04/fulldemo.py
+++ b/chall04/fulldemo.py
@@ -1,0 +1,80 @@
+#! /usr/local/bin/python3
+
+import sys
+import fileinput
+
+def error(e, n):
+	if e == 0:
+		if n != 0:
+			print("%s: %s: Can't read file" % (sys.argv[0], sys.argv[n]))
+		else:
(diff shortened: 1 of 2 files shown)


esepunto:
Bravo!! Your approachs always impress me. I'm really looking forward to read your solutions for the challengues. Just one comment: upper and lower case is not consider.

fulldemo13:
Really thanks I forgot it!!!

anna-dominguez:
I have some fails with your code because you're checking before the string, for example when I test it with the string `"ie"`. You're not checking if you're at `str[0]` before trying to look before, so you're doing `str[0 - 1]` 😞 
I cannot accept your PR, like pistache commented you with your last challenge, there's a conflict with your branch.

PR #197, opened 2020-04-16
chall05

nsoria042:
We solve it through a control of word length and the appearance of prohibited strings. 
If the length of the word is less than or equal to 1 we return a 1. 
If the length of the word is 2 and other than 'ei' we return 1. 
And for word lengths greater than 2 we check using the strstr function that neither 'cie' or any combination of letter (except c) + 'ei' appears and return 1, in all other cases we return 0.

### DIFF
--- a/chall05/isoria-g.c
+++ b/chall05/isoria-g.c
@@ -1,0 +1,58 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   isoria-g.c                                         :+:      :+:    :+:   */
+/*                                                    +:+ +:+         +:+     */
+/*   By: marvin <marvin@student.42.fr>              +#+  +:+       +#+        */
+/*                                                +#+#+#+#+#+   +#+           */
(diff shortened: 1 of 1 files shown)


chall05/isoria-g.c
```diff
@@ -0,0 +1,58 @@
...
+        strstr(str, "wei") != NULL || strstr(str, "xei") != NULL ||
+        strstr(str, "yei") != NULL || strstr(str, "zei") != NULL ||
+        strstr(str, "cie") != NULL)
+        return (0);
+    else
+        return (1);
+}
+
+int ft_ie_except_after_c(char* str)
+{
```
aguimbao, line 49: You should check for `str == NULL` when working with `char*`.

aguimbao (commented):
I don't like this approach at all. It's not a good idea to hardcode all those possibilities like that. It's hard to read, you lose modularity, you have a higher risk in edge cases, it's usually slower, etc.

Also, you forgot to return 0 with `str = "ei"`.

PR #198, opened 2020-04-16
Chall05_fvega-tr

fvega-tr:


chall05/fvega-tr.c
```diff
@@ -0,0 +1,68 @@
...
+{
+	int i;
+	int len;
+	int boolean;
+	char *str;
+	
+	boolean = 1;
+	i = 0;
+	len = 0;
+	if (!str2 || !*str2)
```
aguimbao, line 50: Nice check but you forgot to check if `str == ""`.

chall05/fvega-tr.c
```diff
@@ -0,0 +1,68 @@
...
+		return (0);
+	
+	
+	while (str2[i])
+		i ++;
+	len = i;
+	i = 0;
+	str = malloc(sizeof(char) * (len + 1));
```
aguimbao, line 58: Malloc was unnecessary, we were only checking for lowercase strings.

chall05/fvega-tr.c
```diff
@@ -0,0 +1,68 @@
...
+		c += 32;
+	return (c);
+}
+
+int ft_bucle(char *str, int len, int boolean)
+{
+	int i;
+	
+	i = 0;
+		while (str[i])
```
aguimbao, line 17: You always go out of bounds here. Very dangerous.

aguimbao (commented):
You fail for `str = ""`, also you go out of bounds constantly. And there was no PR explanation for your code, so I won't be able to merge it.

fvega-tr:
I don't understand Where I go out bounds

aguimbao:
In line 17, I commented it. You can debug your code and you will understand. You go out of bounds for `str = "ie"`, you can debug that case.

PR #199, opened 2020-04-16
Chall05 migferna

migferna42:
I look for the first reference of the 'i' character, and check if the rules apply, any case that does not comply with the rule returns 0, if it is valid, the input word is updated (making the word smaller in each iteration). If we get out of the loop, it can only mean that it meets the standards.

*All inputs to function ft_ie_except_after_c (char * str), lowercase conversion applied


anna-dominguez:
I think you could simplify a lot you code. It's working for all my tests with a simple `gcc`, but you have a fail on my first test on`"ie"` you forgot the case when your string starts with the characters you're looking at, and you're basically doing `str[0 - 1]` writing `prev = it - 1`, you're out of bound.

PR #200, opened 2020-04-16
chall05 - gbudau

gbudau:
**How i solved this problem / Short description of the code**

:x: <kbd>CIE</kbd>
While iterating over the string if the current character is <kbd>C</kbd> and if the next two characters are <kbd>I</kbd> and <kbd>E</kbd> will `return 0`.

:heavy_check_mark: <kbd>CE</kbd>
If the current character is <kbd>E</kbd> and the previous is <kbd>C</kbd> (using pointer arithmetic to make sure the comparison is not out of bounds) will continue the loop and not compare the next character.

:x: <kbd>EI</kbd>
Else will check if the next character is <kbd>I</kbd> and `return 0` if it is.

Keep looping and comparing until the end of the string.
:heavy_check_mark: If all cases are valid it will `return 1`.

Using `tolower()` function so the comparisons are case-insensitive.

### DIFF
--- a/chall05/gbudau.c
+++ b/chall05/gbudau.c
@@ -9,7 +9,7 @@ int		ft_ie_except_after_c(char* str)
 	while (*str)
 	{
 		if (tolower(*str) == 'c')
-			if (*(str + 1) && tolower(*(str + 1)) == 'i')
+			if (tolower(*(str + 1)) == 'i')
 				return (0);
 		if (tolower(*str) == 'e')
 		{
@@ -18,7 +18,7 @@ int		ft_ie_except_after_c(char* str)
 				str++;
 				continue;
 			}
-			if (*(str + 1) && tolower(*(str + 1)) == 'i')
+			if (tolower(*(str + 1)) == 'i')
 				return (0);
 		}
 		str++;
--- a/chall05/gbudau.c
+++ b/chall05/gbudau.c
@@ -9,8 +9,10 @@ int		ft_ie_except_after_c(char* str)
 	while (*str)
 	{
 		if (tolower(*str) == 'c')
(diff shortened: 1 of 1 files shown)


chall05/gbudau.c
```diff
@@ -0,0 +1,29 @@
...
+		}
+		else if (tolower(*str) == 'e')
+		{
+			if (str - start > 0 && tolower(*(str - 1)) == 'c')
+			{
+				str++;
+				continue;
+			}
+			if (tolower(*(str + 1)) == 'i')
+				return (0);
```
clafoutis42, line 24: Having to use `ft_lower` function prevents you from doing so, but you could have used `strstr` to look for the wanted occurrences. For example you know that you should never find `cie`, this could be solved in an `if`.  

You also could have organized your conditions differently, I must say that this is really hard to read. Maybe putting some of the checks in an additional fonction would have eased the reading too.

clafoutis42 (commented):
Good job. You have a working solution!

chall05/gbudau.c
gbudau: Sorry about the readability, i made another version based on your sugestions and
it looks a lot better :slightly_smiling_face:

```
#include <string.h>

int	ft_ie_except_after_c(char *str)
{
	size_t	i;

	if (strstr(str, "cie"))
			return (0);
	i = 0;
	while (str[i])
	{
		if (str[i] == 'e')
		{
			if (i && str[i - 1] == 'c')
			{
				i++;
				continue;
			}
			if (str[i + 1] == 'i')
				return (0);
		}
		i++;
	}
	return (1);
}
```

Also thank you for the feedback :heart:

PR #201, opened 2020-04-17
chall05 - mmartin-

Mariomm-marti:
**Challenge approach:**
I simply thought on iterating through the whole string, checking whether the rule pass or not.

**Code implementation:**
Iterate the string, checking if the following character is not NUL (as the check wouldn't make sense). If previous char isn't 'c' but I'm at 'ei', just return false.
Happens the same way if I'm on 'c' and the following characters are 'ie'.

Finally, returns true if everything went well.

### DIFF
--- a/chall05/mmartin-.c
+++ b/chall05/mmartin-.c
@@ -1,0 +1,32 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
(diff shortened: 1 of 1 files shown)


aguimbao:
Nice standard and short implementation, pointer arithmetic > indexes ❤️ .

Even if you don't crash, you go out of bounds for `str = "ie"` for example.

PR #202, opened 2020-04-17
Chall05

veleita:
I am aware that the time limit is over, I am sorry for that. My function returns false only in the cases where the rule does not apply, for everything else it returns true. I wanted to make the first letter of the word lowercase in case it was uppercase, but I haven't managed to make it compile so the function will not recognise invalid cases where the first letter is uppercase (f.ex.: "Cieling" ). Please don't be harsh on me or else I will cry.

### DIFF
--- a/chall05/mzomeno-.c
+++ b/chall05/mzomeno-.c
@@ -1,0 +1,51 @@
+#include <stdio.h>
+
+short ft_ie_except_after_c(char *str)
+{
+  /*
+  * First, we make the first letter lowercase by applying the following mask:
+  * 32 = 00100000
+  * Thus, if bit 5 was 0 (uppercase letters always have their bit five put to 0),
+  * we would set it to 1 (so the resultant byte would be the correspondant
+  * lowercase letter). Else, the byte would remain the same.
+  *
+  * Note that uppercase letters in the middle of the word are not taken into account
+  */
+  /**
+   ** I wanted to put the following line of code into the function:
+   ** *str |= 32;
+   ** But it doesn’t compile. It throws a segfault error, I haven’t figured out why
+   **/
+  while (*str)
+    { 
+      /*
+       * Now, we check out the "cie" cases
+       */
+      if (*str == 'c' && *(str + 1) == 'i' && *(str + 2) == 'e')
+      return (0);
+      /*
+       * Then, we check out the "ei" cases which aren’t preceeded by ’c’
+       */
+      if (*str == 'e' && *(str + 1) == 'i' && *(str - 1) != 'c')
+      return (0);
+      /*
+       * I am aware that these two conditions could have been put together
+       * with an || separator, but for the sake of legibility I will leave
+       * the code like that
+       */
+      str++;
+    }
+  /*
+   * For every possible combination of letters that hasn’t been checked out,
+   * the rule is considered to be followed and the program returns true (1).
+   */
+  return (1);
+}
+
+int main()
+{
+  char *word = "juicier";
+
+  printf ("%s == %d", word, ft_ie_except_after_c(word));
+  return (0);
+}


chall05/mzomeno-.c
```diff
@@ -0,0 +1,51 @@
...
+      str++;
+    }
+  /*
+   * For every possible combination of letters that hasn’t been checked out,
+   * the rule is considered to be followed and the program returns true (1).
+   */
+  return (1);
+}
+
+int main()
```
aguimbao, line 45: We didn't ask for a `main()`, just a function.

I know it was just for testing 😛.

chall05/mzomeno-.c
```diff
@@ -0,0 +1,51 @@
...
+    { 
+      /*
+       * Now, we check out the "cie" cases
+       */
+      if (*str == 'c' && *(str + 1) == 'i' && *(str + 2) == 'e')
+      return (0);
+      /*
+       * Then, we check out the "ei" cases which aren’t preceeded by ’c’
+       */
+      if (*str == 'e' && *(str + 1) == 'i' && *(str - 1) != 'c')
```
aguimbao, line 29: Checking for `index - 1` is usually very dangerous, you always need to check your bounds.

In this case, you go out of bounds for `str = "ei"`, the program doesn't crash, but it's still very very dangerous.

aguimbao (commented):
About the uppercase to lowercase. I think you were looking for the `XOR` operator -> `^=`. Thing is, we were only testing full lowercase strings. So your program passed all the tests, even if you go out of bounds at times.

It's a shame that you submitted after the deadline, cause it was very well written and explained and I liked the simple approach. I'm sure you will be in time for the next challenge on Monday 🤓

chall05/mzomeno-.c
veleita: Okay. I thought it would be easier for you to test the function if I left the main there. I will submit the function clean next time.

chall05/mzomeno-.c
veleita: I heard about it but I still don't understand why is it so problematic to read memory outside the string. I can research it on my own, anyways.

veleita:
But the toupper function wouldn't work with a XOR operator, would it? That would turn the letter to uppercase if it was lowercase. Instead, an OR operator sets the bit 5 to one if it is 0, and leaves it untouched if it is already 1 - that is, if it is already a lowercase letter.

vdedios:
Nice solution!! The reason why that `*str |= 32;` gave you (probably) bus error is because you are trying to modify the value of a string literal (string that is on static memory) and that is not allowed in C. Strings allocated on stack are considered read-only.

The only scenario where that could work would be having a string allocated on the heap. In your main, you can try for example doing something like:

`_printf ("%s == %d", word, ft_ie_except_after_c(**_strdup_**("hello world"));_
`

But there is a problem with that, you don't know beforehand if an input string comes from stack or heap... So the best way to approximate a solution would be just returning` (*str | 32) `

I leave you a couple of references: 
https://stackoverflow.com/questions/1011455/is-it-possible-to-modify-a-string-of-char-in-c
https://softwareengineering.stackexchange.com/questions/294748/why-are-c-string-literals-read-only

chall05/mzomeno-.c
aguimbao: If you read out of bounds you will most likely read adjacent variables, or garbage memory. It causes undefined behavior and if you were to interact with that memory it could cause all kind of problems.

aguimbao:
String literals are more often than not either in stack, nor heap, but hardcoded on data segments. They definitely shouldn't be modified. 

I think `XOR`ing the characters will just inverse the case, so if you apply it only to the uppercase characters it will make them lowercase, but maybe it isn't the best approach. It'll let you guys read more about all of this ! Good job both of you in your challenges anyway !

PR #204, opened 2020-04-20
vde-dios.c

vdedios:
# SOLUTION

## Error management
There is a separated function that checks if king is duplicated, there any more characters beyond the allowed set `.KPBQR\n` and also checks if the board is a valid square (constant side plus length under `81 chars + 8 separators`

## Calculations
First of all, king position is calculated in order to reference all calculations from there. We will start moving from the beginning of the string. If we encounter a valid enemy (explained bellow) we then calculate if there is any collision on its way to the king (diagonal, vertical or horizontal depending on the enemy). If no obstacle is encountered will have success and will `return 0.` Otherwise we will keep stepping through the string and looking for more enemies.

### How are positions calculated?
In order to make all calculations, string positions are transformed from 1D to 2D space. This is achieve by:
`pos.x = string_position / (board_side + 1)`
`pos.y = string_position - (board_side + 1) * pos.x`

### What is a valid enemy?
There are two types of enemies, horizontal and diagonal. They are checked as follows:
- **Horizontal:** They have to be the correct enemy `RQ`  and its x or y position must be the same as the king
- **Diagonal:** They have to be the correct enemy `PBQ` and the absolute value of its slope (calculated as `Δy/Δx`) must be 1. **NOTE:** if we have a `P` item, Δx must be 1 and no more. Also important, if `Δx = 0` we exit the function to avoid float point errors.

All Δx and Δy are calculated from the king to the element.

### DIFF
--- a/chall06/vde-dios.c
+++ b/chall06/vde-dios.c
@@ -12,18 +12,15 @@ typedef struct s_element{
 
 int ft_check_format(char *board)
 {
-    int i;
+    int i = -1;
+    int len = strlen(board);;
     int side;
-    int len;
 
-    //Valid charset
-    i = -1;
-    len = strlen(board);
-
+    //Valid charset?
     while(++i < len)
         if (!strchr(".KPBRQ\n", board[i]))
             return (0);
-    //Valid square
+    //Valid square?
     if (len > 89)
         return (0);
     side = sqrt(len);
@@ -34,7 +31,7 @@ int ft_check_format(char *board)
             return (0);
         i += side + 1;
     }
-    //Only one king
+    //Only one king?
     if (strchr(strchr(board, 'K') + 1, 'K'))
         return (0);
     return (1);
@@ -54,11 +51,9 @@ t_element ft_element_coord(int pos, int side, char *board)
 
 int     ft_check_colission(t_element king, t_element enemy, char *board)
 {
-    int x;
-    int y;
+    int x = enemy.x - king.x;;
+    int y = enemy.y - king.y;;
     
-    x = enemy.x - king.x;
-    y = enemy.y - king.y;
     while (abs(x) > 1 ||abs(y) > 1)
     {
         if (x)
@@ -73,21 +68,24 @@ int     ft_check_colission(t_element king, t_element enemy, char *board)
 
 int     ft_check_enemy(t_element king, t_element enemy, char *board)
 {
+    int delta_x = (enemy.x - king.x);
+    int delta_y = (enemy.y - king.y);
+
     if (strchr("RQ", enemy.id))
          if ((enemy.x == king.x ||enemy.y == king.y))
              return(!ft_check_colission(king, enemy, board));
     if (strchr("QB", enemy.id))
-        if (abs((enemy.x - king.x) / (enemy.y - king.y)) == 1)
+        if (delta_x && delta_y && abs(delta_y/ delta_x) == 1)
              return(!ft_check_colission(king, enemy, board));
     if (enemy.id == 'P')
-        if ((enemy.y - king.y) == 1 && abs((enemy.x - king.x)) == 1)
+        if (delta_y == 1 && abs(delta_x) == 1)
              return(!ft_check_colission(king, enemy, board));
     return (0);
 }
 
 int ft_check_mate(char *board)
 {
-    int pos;
+    int pos = 0;
     int side;
     t_element king;
     t_element enemy;
@@ -95,7 +93,6 @@ int ft_check_mate(char *board)
     if (!ft_check_format(board))
         return (1);
     side = sqrt(strlen(board));
-    pos = 0;
     while (board[pos] != 'K')
         pos++;
     king = ft_element_coord(pos, side, board);
--- a/chall06/vde-dios.c
+++ b/chall06/vde-dios.c
@@ -1,0 +1,116 @@
+#include <math.h>
+#include <string.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+typedef struct s_element{
+    int x;
+    int y;
+}       t_element;
+
+int ft_check_format(char *board)
+{
+    char *set = ".KPBRQ\n";
+    int i;
+    int size;
+    int len;
+
+    //Valid charset
+    i = -1;
+    len = strlen(board);
+
+    while(++i < len)
+        if (!strchr(set, board[i]))
+            return (0);
+    //Valid square
+    if (len > 89)
+        return (0);
+    size = sqrt(len);
+    i = size;
+    while (board[i] && i < len)
+    {
+        if (board[i] != '\n')
+            return (0);
+        i += size + 1;
+    }
+    //Only one king
+    if (strchr(strchr(board, 'K') + 1, 'K'))
+        return (0);
+    return (1);
+}
+
+int ft_get_king_pos(char *board)
+{
+    int pos;
+
+    while (board[pos] != 'K')
+        pos++;
+    return (pos);
+}
+
+t_element ft_element_coord(int pos, int size)
+{
+    t_element el;
+
+    el.x = pos / (size + 1);
+    el.y = pos - (size + 1) * el.x;
+    return (el);
+}
+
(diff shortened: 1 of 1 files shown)


chall06/vde-dios.c
```diff
@@ -0,0 +1,111 @@
...
+    side = sqrt(len);
+    i = side;
+    while (board[i] && i < len)
+    {
+        if (board[i] != '\n')
+            return (0);
+        i += side + 1;
+    }
+    //Only one king?
+    if (strchr(strchr(board, 'K') + 1, 'K'))
```
clafoutis42, line 35: What if there's no K in the input? Well, you'll have a segfault :/

chall06/vde-dios.c
```diff
@@ -0,0 +1,111 @@
...
+    el.side = side;
+    return (el);
+}
+
+int     ft_check_colission(t_element king, t_element enemy, char *board)
+{
+    int x = enemy.x - king.x;;
+    int y = enemy.y - king.y;;
+    
+    while (abs(x) > 1 ||abs(y) > 1)
```
clafoutis42, line 57: This condition means that at one point the direction will change.  _(Disclaimer: I noticed the formulas for your coordinates are wrong, this too might be the cause of the problem I noticed)_
For example here:
```
........
........
........
....B...
........
.......K
........
........
```
Your code returns that the king in the bishop's range, although it is not.

chall06/vde-dios.c
```diff
@@ -0,0 +1,111 @@
...
+    {
+        if (x)
+            x = x - (x > 0 ? 1 : -1);
+        if (y)
+            y = y - (y > 0 ? 1 : -1);
+        if (board[(king.side + 1) * (king.x + x) + (king.y + y)] != '.')
+            return (1);
+    }
```
clafoutis42, line 65: You have a good idea with this function, the problem here is that you alter the movement meant by the piece. In fact, it seems you will always find your way to the king when it comes to pieces that move in diagonal at least.

chall06/vde-dios.c
```diff
@@ -0,0 +1,111 @@
...
+
+    if (strchr("RQ", enemy.id))
+         if ((enemy.x == king.x ||enemy.y == king.y))
+             return(!ft_check_colission(king, enemy, board));
+    if (strchr("QB", enemy.id))
+        if (delta_x && delta_y && abs(delta_y/ delta_x) == 1)
+             return(!ft_check_colission(king, enemy, board));
+    if (enemy.id == 'P')
+        if (delta_y == 1 && abs(delta_x) == 1)
+             return(!ft_check_colission(king, enemy, board));
```
clafoutis42, line 82: This is hard to read. You could have used a function per piece, and to avoid repetitive ifs, you could have used a function pointer array.

Here's a snippet of my code to give you a better idea of what I mean:
```c
static int		chess_bishop(char *board, int x, int y, int size) {
 	if (piece_move(board, x, y, 1, 1, size) == CHESS_MATE)
 		return (CHESS_MATE);
 	if (piece_move(board, x, y, -1, -1, size) == CHESS_MATE)
 		return (CHESS_MATE);
 	if (piece_move(board, x, y, 1, -1, size) == CHESS_MATE)
 		return (CHESS_MATE);
	if (piece_move(board, x, y, -1, 1, size) == CHESS_MATE)
		return (CHESS_MATE);
	return (CHESS_SAFE);
}
```

```c
static void		init_moves(f_chess *moves) {
	int i = 0;

	while (i < 255)
		moves[i++] = &chess_unknwn;
	moves['.']	= &chess_case;
	moves['\n']	= &chess_case;
	moves['\0']	= &chess_case;
	moves['K']	= &chess_case;
	moves['P']	= &chess_pawn;
	moves['B']	= &chess_bishop;
	moves['R']	= &chess_rook;
	moves['Q']	= &chess_queen;
}
```

```c
static int		piece_move(char *board, int x, int y, int xd, int yd, int size) {
	while (IS_INSIDE(x, y, size) && board[compute_pos(x+xd, y+yd, size)] == '.') {
		x += xd;
		y += yd;
	}
	if (!IS_INSIDE(x+xd, y+yd, size) || board[compute_pos(x+xd, y+yd, size)] != 'K')
		return (CHESS_SAFE);
	return (CHESS_MATE);
}
```

chall06/vde-dios.c
```diff
@@ -0,0 +1,111 @@
...
+    return (1);
+}
+
+t_element ft_element_coord(int pos, int side, char *board)
+{
+    t_element el;
+
+    el.x = pos / (side + 1);
+    el.y = pos - (side + 1) * el.x;
+    el.pos = pos;
```
clafoutis42, line 46: Is this element of the structure really useful? It seems you never use it

chall06/vde-dios.c
```diff
@@ -0,0 +1,111 @@
...
+        return (0);
+    return (1);
+}
+
+t_element ft_element_coord(int pos, int side, char *board)
+{
+    t_element el;
+
+    el.x = pos / (side + 1);
+    el.y = pos - (side + 1) * el.x;
```
clafoutis42, line 45: Those calculation are wrong:
```
........
........
........
....B...
........
.......K
........
........
```
In this board the king is situated at `X: 7`,`Y: 5` (`ordinal: 52` in the string).  
But, given that `x = pos / (side + 1)`, this would mean that `7 = 52 / (8 + 1)` -> `7 = 52 / 9` -> `7 = 5`. Which as you already know, is wrong.
It is the same with `y = pos - (side + 1) * x` (assuming that we got `x` right this time):
`5 = 52 - (8 + 1) * 7` -> `5 = 52 - 9 * 7` -> `5 = -11`, which is also wrong.

Given this formula (which is correct): `pos = x + y * (size + 1)`, we have the following formulas:
- `x = pos - y * (size + 1)`
- `y = (pos - x) / (size +1)`

As `x` and `y` are co-dependent there are 2 solutions:
- If you want to keep going with formulas, I think there is one, but to be honest I didn't want to do more maths than necessary
- You could have incremented `x` and `y` earlier in the loop

I chose the second option, and used the first formula to compute the position in the string.

clafoutis42 (commented):
You have good ideas, but a lot of errors. Keep going, you were almost there !  
Maybe try to run more tests next time :)

chall06/vde-dios.c
vdedios: mmm... I think this is due to my mistake writting '/' instead of '%'

chall06/vde-dios.c
vdedios: Despite I have first said that I made a mistake writting `/` instead of `%` (bellow I explain why) I just  have realized what is happening here. 

I think you have misunderstood how I have made these calculations. It may be due to not applying a convencional standard in computer science. I have assumed X as rows and Y as columns as you would do on Algebra. 

So when in the example you point, king is really in `X: 5, Y: 7`. If we follow my formulas:

In **X** we would have: `x = pos / (side + 1)` would lead to  `x = 52 / (8 + 1)` and `x = 52 / 9` so `x = 5` which is totally right. 

If we move to **Y**, we would have: `y = pos - (side + 1) * x` that would lead to `7 = 52 - (8 + 1) * 5` and `y = 52 - 45 ` so `y = 7` which is also totally right.

I admit that this formulas may not be the most optimized ones, but they work. If we wanted to have a **leaner solution** we could have done something different and here comes the modulus operator I was referring to. Assuming now a conventional standard (X as  columns and Y as rows)

- **X:** `x = pos % (side + 1)` leads to `x = 52 % 9` so `x = 7`
- **Y:** `y = pos / (side + 1)` leads to `y = 52 / 9` so `y = 5`

And now you have a pair of lean formulas that work in any of both conventions.

I assume then that the mistakes I have come from other part of the code.

chall06/vde-dios.c
vdedios: 😐

chall06/vde-dios.c
vdedios: 📝

chall06/vde-dios.c
vdedios: The tests I made (although it has been demonstrated that I should run more) I did not have issues finding blocking pieces in diagonal

PR #205, opened 2020-04-20
edramire - Asgard

edramir18:
The purpose of this function is determine based in a chess board position passed as parameter, if the king is on check by the others pieces.
I tried to make a simple solution with few iterations and without additional memory allocations. I check the size of the string to validate if is valid or not, zero or more of 89 characters is a string invalid. After that check all lines are the same size and the number of  lines and characters by line are the same.
The string is checked for multiples characters `K` or any character not valid, and the king position is stored for posterior usage.
Taking the parametric form of a line, i check for every direction starting from the king position, for vertical and horizonal directions check for `Q` and `R` characthers, and on the diagonal directions check for `Q` and `B` characters. The case for peons is checked as a special case.
The advantage of using the parametric form, is that possible to use one function for any direction without worry about how to build the index to use in each verification.


chall06/edramire.c
```diff
@@ -0,0 +1,116 @@
...
+	len = strlen(board);
+	size = get_chess_size(board, len);
+	king = size > 0 ? get_king_index(board) : -1;
+	if (king == -1)
+		return (1);
+	if (king + size + 2 < len && board[king + size + 2] == 'P')
+		return (0);
+	if (king + size < len && board[king + size] == 'P')
+		return (0);
+	check = check_mate(king, size + 1, board, 1, 0, 'R');
```
aguimbao, line 107: Honestly, I found your parametric function very interesting, I looked it up and I learned some maths 😄.

Maybe it wouldn't have been more legible to loop over the `check_mate()` function with an enum holding all the posible `dx`, `dy`, and `c` possibilities than having all those ternaries laying around? There's a lot of extra checks if you find the check mate condition in the first function call.

chall06/edramire.c
```diff
@@ -0,0 +1,116 @@
...
+			++board;
+		}
+		if (row == 0)
+			size = col;
+		else if (col != size)
+			return (0);
+		++row;
+		board += *board == '\n' ? 1 : 0;
+	}
+	return (row == col ? size : 0);
```
aguimbao, line 41: You successfully return `0` for a `\n` at the beginning of the board, but not at the end. (But this is a bit of an edge case and it wasn't completely specified)

aguimbao (commented):
I liked your parametric function approach, but I think this exercise was asking for a couple `typedef`'s and a structure maybe. I just don't like seeing that many parameters being passed around each function, but maybe that's my norminette-mind.

Good job though !

chall06/edramire.c
edramir18: Thanks a lot, i tried to make something different. I dislike that function too but only had an extra parameter, on other aspects is still on norm. Because the code is for the challenge i try to be clear on the code, without expressions too complex. I'm happy i can show you something new.
I will take all your advices in mind on  for future challenges.

chall06/edramire.c
edramir18: I checked the error commented by you, but that function is only to get the size of the line. In case the line has a new line character at the beginning the funcion check_mate will return 1, as specified in the subject of the challenge.

aguimbao: Yes you check for "\n...K\n...R....\n...." ==  invalid cause `\n` at the beginning, but what about "...K\n...R....\n....\n" == invalid cause `\n` at the end ? 

Anyway I'm not sure this tests will be included in the end.

edramir18: In that case, the function will fail on a 9x9 map with new line character at the end, i take only as valid a string with 81 character and 8 new line character; this validation are incomplete 
`if (len == 0 || len > 89)`. On smaller maps has will not problem

aguimbao: If the map is already 9x9 it will detect it yeah but for example here -> `board = ".R.....K\n........\n........\n........\n........\n........\n........\n........\n"` it's an 8x8 with an extra `\n` at the end it should return `1` cause of it, but you return `0`. Anyway, again, I don't think we will add this test at the end. :)

PR #206, opened 2020-04-20
chall06 - pcuadrad

pabloocg:
First I do a split for the line breaks to be able to handle the board better.
I do checks to see if the length of each line correspond to the line number of the board and also to see if no different character than Q, R, P, B, '.', K has been set.
I also check if there is more than one king (K) and if there is at least one figure that can attack the king.
After all the checks have been made, I scan the board looking for the figures that attack the king.
If I find a P (Pawn) I do a function that checks if the king is in a diagonal position to the left and right.
For the other figures, I have made the function move:
**int     move(char **board, int x_pos, int y_pos, int size, int x_counter, int y_counter)**
This function subtracts or adds the position of x and y depending on x_counter and y_counter. In this way I am able to traverse diagonals to the right and left (Bishop) and also in a straight line (Rook).
If it finds a different figure than the king(K), I stop the loop and return 0. And if I find the king, I return 1.
Lastly, for the Queen (Q), since she can make all the movements, I first make a search as if she were a bishop and then as a rook, doing all her moves.

### DIFF
--- a/chall06/pcuadrad.c
+++ b/chall06/pcuadrad.c
@@ -1,0 +1,260 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+static int		isinset(char c, char charset)
+{
+	if (charset == c)
+		return (1);
+	return (0);
+}
+
+static int		count_str(char const *str, char c)
+{
+	int		count;
+
+	count = 0;
+	while (*str)
+	{
+		while (*str && isinset(*str, c))
+			str++;
+		if (*str && !isinset(*str, c))
+		{
+			count++;
+			while (*str && !isinset(*str, c))
+				str++;
+		}
+	}
+	return (count);
+}
+
+static char		*malloc_string(char *str, char c)
+{
+	char	*word;
+	int		i;
+
+	i = 0;
+	while (str[i] && !isinset(str[i], c))
+		i++;
+	if (!(word = (char *)malloc(sizeof(char) * (i + 1))))
+		return (NULL);
+	i = 0;
+	while (str[i] && !isinset(str[i], c))
+	{
+		word[i] = str[i];
+		i++;
+	}
+	word[i] = '\0';
+	return (word);
+}
+
+static char		**free_split(char ***tabstr)
+{
+	int i;
+
+	i = 0;
+	while (*tabstr[i])
+	{
+		free(*tabstr[i]);
+		i++;
+	}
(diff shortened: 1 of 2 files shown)


chall06/pcuadrad.c
```diff
@@ -0,0 +1,260 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+static int		isinset(char c, char charset)
```
aguimbao, line 5: I think this function would be better formatted as a `#DEFINE` condition, but I think it's pretty subjective.

aguimbao (commented):
I don't really know what to say, your function is pretty well written and explanatory.
Maybe grouping some of the conditions in the `bishop()`/`rook()` functions but I think you would lose readability.
Or using size_t for your indexes, instead of starting them at `-1`, I think it's less dangerous to use unsigned variables, as you can't go out of bounds as often.
Maybe using `static` keyword in front of your functions would be a good practice.
Also this challenge was asking for a good structure in my opinion and a couple `typedef's and `#DEFINE`s, it would improve readability I think.

Also, this wasn't really clarified, but your program might fail for tests with `\n` at the beginning or at the end of the board, or with multiple `\n` separating the rows of the board.

I saw some more optimised approaches, not using malloc and reorganising the rows, and comparing the rows on the fly with some good arithmetic. You also probably didn't need to scan every single case of the board, knowing that the king can only be attacked from the front, back, sides and diagonally.

I saw there was a mistake in the README and corrected, but please don't modify it yourself, ask for it manual change in the slack thread, we won't accept modifications to other files next time.

Your solution works flawlessly for the rest of the tests I performed though.

pabloocg:
Thank you very much for the correction, now that you say it, I see that I spend a lot of time checking each figure, when I could have checked everything from the king's position if one attacked him.
In reference to the line breaks you said, I've tried this: ft_check_mate("\n\n\n\n..K\n\n\n.P.\n\n...\n" and it works, the split function does the job. Or do you mean that it should give an error because the parameter was not well formatted?

aguimbao:
Yes I wasn't clear, `\n` at the beginning or at the end of the board, like  `ft_check_mate("..K\n.P.\n...\n")` should be invalid cause bad formatted, and also  `ft_check_mate("..K\n.P.\n\n...")` same, bad formatted. But I'm not sure this tests will be included in the end.

PR #207, opened 2020-04-20
chall06-fulldemo

fulldemo13:
Hi, first I check the letters of the board and I allocate memory to the string to be able to use the "strtok" function for divide the string. While I'm dividing and saving the board I also check if the board is square.
Now, I start to looking for letters in the map and when I find it I call the function that "moves" the corresponding chess piece.
With rooks and bishops I start looking for up, down, left  and down. If in one of this moves I find the king or I 'm out of string I try with other move.
For the queen I use both functions rook and bishop.
In the end, when I return the value, I also call a function that cleans the memory I used.

### DIFF
--- a/chall06/fulldemo.c
+++ b/chall06/fulldemo.c
@@ -1,0 +1,259 @@
+#include <string.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+int board_check(char *board)
+{
+	int len;
+	int i;
+	int king;
+	int count;
+
+	i  = 0;
+	king = 0;
+	count = 0;
+	while (board[i] != '\0')
+	{
+		if (board[i] != '\n' && board[i] != 'K' && board[i] != 'P'
+		&& board[i] != 'B' && board[i] != 'R' && board[i] != 'Q' 
+		&& board[i] != '.')
+			return (0);
+		if (board[i] == 'K')
+			king++;
+		if (board[i] == 'P' || board[i] == 'B' || board[i] == 'R' || board[i] == 'Q')
+			count++;
+		i++;
+	}
+	if (king > 1 || king == 0 || count == 0)
+		return (0);
+	len = 0;
+	while (board[len] != '\n' && board[len] != '\0')
+		len++;
+	return (len);
+}
+
+int clear_board(char *str, char **board, int r)
+{
+	int len;
+	int i;
+
+	i = 0;
+	if (board)
+	{
+		len = strlen(board[0]);
+		while (i < len)
+		{
+			free(board[i]);
+			i++;
+		}
+		free(board);
+	}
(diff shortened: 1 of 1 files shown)


chall06/fulldemo.c
```diff
@@ -0,0 +1,259 @@
...
+		}
+		i++;
+	}
+	return (1);
+}
+
+int ft_check_mate(char *board)
+{
+	int len;
+	char **map;
```
clafoutis42, line 243: Although it seems easier to do this, you don't have to. You already have the board, every lines is separated by a `\n`, and you are supposed to have a perfect square. Which means you can use coordinates in a formula to get the position of your pieces: `pos = x + y * (size + 1)`  

Using such a method will help you avoid useless memory allocations.

chall06/fulldemo.c
```diff
@@ -0,0 +1,259 @@
...
+		if (map[n][j] == 'K')
+			return (1);
+	}
+	if ((n = j - 1) >= 0)
+	{
+		while (n > 0 && map[n][j] == '.')
+			n--;
+		if (map[n][j] == 'K')
+			return (1);
+	}
```
clafoutis42, line 114: You could made a single function for all of this, to which you would have passed the different directions on `x` and `y`.
This same function could have been used by all of your pieces !

chall06/fulldemo.c
```diff
@@ -0,0 +1,259 @@
...
+		i++;
+	}
+	return (map);
+}
+int rook(char **map, int i, int j)
+{
+	int n;
+	int len;
+
+	len = strlen(map[0]) - 1;
```
clafoutis42, line 86: Because of this `- 1` you are missing all the pieces that are on the right and bottom extremities of the board, e.g:
```
RK
..
```

I found the real source of the error: [see this comment](https://github.com/42madrid/remote-challs/pull/207/files#r413107996)

clafoutis42 (commented):
You have an almost correct solution. You were mistaken in your algorithm and you are missing the pieces at the right and bottom extremities of the board.  
Also you are using malloc, it was not necessary in this exercices.
Keep going!

chall06/fulldemo.c
```diff
@@ -0,0 +1,259 @@
...
+			return (1);
+	}
+	if ((n = j - 1) >= 0)
+	{
+		while (n > 0 && map[i][n] == '.')
+			n--;
+		if (map[i][n] == 'K')
+			return (1);	
+	}
+	if ((n = i + 1) < len)
```
clafoutis42, line 101: In fact I just noticed the error I mentioned earlier is here, the condition is wrong, you should have used inferior or equal here!  
This error is only present for the rook, the bishop seems to be fine.

PR #208, opened 2020-04-20
Chall06 szavarce

szavarce42:
1) call a fuction to check for basic validations: is it square? are all lines the same width?
2) I check for the number of lines, and if greater than 9 it will return 1
3) function ft_get_board will store the board in a matrix of ints
4) Then I check if the pieces I have are valid. There should be only one king
5) I then call the function that will check the moves.
6) This function will call a specific function depending on the piece
7) If the piece is a Queen, it will check for both the Rook and Bishops functions.


chall06/szavarce.c
```diff
@@ -0,0 +1,313 @@
...
+        pos++;
+    }
+    if (xcount != ycount)
+        return (0);
+    return (1);
+}
+
+int ft_check_mate(char *board)
+{
+    int **newboard;
```
clafoutis42, line 298: Why would you use an `int` matrix to store characters?  
Also you do not need to convert the string into a matrix, you could use a formula to calculate the position in the string with `x` and `y`.

chall06/szavarce.c
```diff
@@ -0,0 +1,313 @@
...
+        }
+        else
+        {
+            xcount++;
+        }
+        pos++;
+    }
+    if (xcount != ycount)
+        return (0);
+    return (1);
```
clafoutis42, line 293: You could have used strchr to get the different delimiters, and later subtract pointers to get the lenght

chall06/szavarce.c
```diff
@@ -0,0 +1,313 @@
...
+        while (posx <= endx)
+        {
+            if (newboard[posy][posx] != '.' && newboard[posy][posx] != 'K')
+                return (0);
+            posx++;
+        }
+        posy++;
+    }
+    return (1);
+}
```
clafoutis42, line 58: I don't understand what you are trying to do here, it seems you are going from row to row to see if the king is here.  
I don't recognize the patter of any piece :/

I mean, why are you iterating over all the `x`s?

Really it's hard to understand what you are trying to do :/

clafoutis42 (commented):
Nice try, however your code doesn't seem to work. I don't really understand how you tried to implement the pieces functions.  
It seems you are trying to do something really complicated, when the solution was simpler.  
You could have had a function that simply takes the direction (whether it should add or subtract to the `x` and/or `y`) ?

PR #209, opened 2020-04-20
chall06: dperez-z

dperez42:
First i create three funtions to checks:
- check_is_king: If there is only  one king.
- check_is_pieces: check is all de characters in the string are permitted.
- check_dimension: check if the board is square and less than 9 per side.
In the main function, after make the before checkings. I read the string and copy it in an array for easy use, at the same time store the position of the king, kr y kc. 
With the king´s position check if there is a piece in  the ways of movements  of tower, peon and bishop. 
Also checks is the string is null or "".


chall06/dperez-z.c
```diff
@@ -0,0 +1,246 @@
...
+				kc = c;
+				kr = r;
+			}
+			i++;
+			c++;
+		}
+		i++;
+		r++;
+	}
+	/*
```
aguimbao, line 166: Please remove unused code before final submission

chall06/dperez-z.c
```diff
@@ -0,0 +1,246 @@
...
+        }
+
+
+	return(0);
+}
+
+				
+int ft_check_mate(char *board)
+{
+	int col;
```
aguimbao, line 110: This is... not pretty. This exercise was asking for a coord based struct, I know there's no norminette, but this is much harder to read.

chall06/dperez-z.c
```diff
@@ -0,0 +1,246 @@
...
+                        c++;
+                }
+                printf("\n");
+		r++;
+        }
+	*/
+	i = 0;
+	while (kr-i>-1)
+	{
+		if (table[kr-i][kc]=='Q' || table[kr-i][kc]=='R')
```
aguimbao, line 185: This will segfault for `board = "K"`.

chall06/dperez-z.c
```diff
@@ -0,0 +1,246 @@
...
+int ft_strlen(char *board)
+{
+	int i;
+	i = 0;
+	while(board[i]!='\0')
+		i++;
+	return(i);
+}
+
+int check_rook(char *board, int row, int king)
```
aguimbao, line 79: You never use this function.

aguimbao (commented):
Your function is pretty hard to read. You should have separated it in different functions.

Your bishop and rook checks fail a lot of times, for example `board="..B.....\n........\n........\n........\n........\n.......K\n........\n........"` should return `0`. , same as `board="..R....K\n........\n........\n........\n........\n........\n........\n........"`, but you return `1` in both.

I think malloc wasn't needed, with some smart maths, you could do your checks much better. Also, the trick for this exercise was using a smart approach, starting your checks from the king position, given he can only be attacked from the front, back, sides, or diagonally, you didn't have to check every position in the board.

I think this exercise needed some `typedef`/`#DEFINE`s and a struct would have been much better than all your variables.

Btw you forgot to free your malloc, if we call your function consecutively your program can crash after a while.

PR #210, opened 2020-04-20
CheckMate Problem - cdorta-n

Top64:
### ¡Hello!

In this project, at the beginning, we check the string for the correct characters that we will use (if we found an error, a message appear and the function end. The next step is to create a bidimensional variable, and then, fill it with our data. With all these steps done, we can look for the different pieces, and watch their area around their coordinates.

**BISHOP PART:**
Check the diagonals in the positive and negative direction looking for the King Coords. If a coordinate is found filled, the search stop. If this was the first direction, the second direction start searching.

**ROOK PART**
Just check if the X coordinate or Y coordinate is the same as the King coordinates. Depending on the equality, we search in the positive and negative of that direction.

*TABLE DISTRIBUTION*

Left Top Side start at [0, 0], if we have a 3x3 square, the bottom right coordinate will be [2, 2].

 _0 1 2  ->> to 8 Max (Being the 9 size)
0 XXX 0
1 XXX 1
2 XXX 2
 _0 1 2


chall06/cdorta-n.c
```diff
@@ -0,0 +1,225 @@
...
+				e = -1;
+			if (board[c] != '.' && board[c] != 'P' && board[c] != 'B'
+			&& board[c] != 'R' && board[c] != 'Q' && board[c] != 'K')
+				e = -1;
+			c++;
+		}
+	}
+	if (e < 0 || k == 0 || lines > 9)
+	{
+		printf("\n--->An error occur with the chessboard, check the input, check the characters and the number of KIng pieces presented <---\n");
```
aguimbao, line 202: We didn't ask for usage printing.

chall06/cdorta-n.c
```diff
@@ -0,0 +1,225 @@
...
+	j = 0;
+	c = 0;
+	while (j < size && board[c] != '\0')
+	{
+		i = 0;
+		while (i < size && board[c] != '\0')
+		{
+			if (board[c] == '\n')
+				c++;
+			printf("[%c]", board[c]);
```
aguimbao, line 39: We didn't ask for board printing.

chall06/cdorta-n.c
```diff
@@ -0,0 +1,225 @@
...
+			}
+			i++;
+		}
+		j++;
+	}
+	
+	//CHECK PAWN POSITION RANGE WITH THE KING
+	if (b > 0)
+	{
+		if (!((table[b + 1][a + 1] == '.' || table[b + 1][a + 1] == 'R')
```
aguimbao, line 71: I think you had some mistakes here, because you don't check for pawns at all. You also `return (0)` here for `board = "..\n.K"`

Due to this mistake, you also fail for the basic pawn test: `board = "K..\n.P.\n..."`

chall06/cdorta-n.c
```diff
@@ -0,0 +1,225 @@
...
+	** THE COMBINATION OF THE OTHER TWO, WE CAN CHECK THE QUEEN
+	*/
+
+	j = 0;
+	while (j < size)
+	{
+		i = 0;
+		while (i < size)
+		{
+			if (table[j][i] == 'B' || table[j][i] == 'Q')
```
aguimbao, line 89: There's also some problems with your Bishop check, you return `0` for `board = "........\n........\n........\n....B...\n........\n.......K\n........\n........"` when you should return `1` cause the King is misaligned with the Bishop

aguimbao (commented):
Hey, good job. Some things to note: 

- You should use `static` keyword for functions that don't need to be used outside of their file-scope, it's good practice.
- We didn't ask for printing of any kind, nor board printing, nor usage printing, this would have invalidated your function if we were in exam :(.
- I think you forgot to check for collisions between pieces, for example `board = "..B.....\n........\n........\n........\n......P.\n.......K\n........\n........"` returns `0`, but it should return `1`, because even if the Bishop and the King are aligned, there's a Pawn in the middle
- You should have probably used structs for all those variables. And divide your function `ft_chesstable()` into smaller functions, I know there's no norminette in these challenges, but it would just take you 15 minutes to organise your code, and it would be much easier to understand and look at.
- You could have also used some `typedef`s or `#DEFINE`s for common situations and conditions, I'll let you look into that.
- I think your rook check is a bit off too, you return `0` for `board = ".R......\n........\n........\n........\n........\n.......K\n........\n........"` which doesn't make any sense 😅

 It honestly was a good approach, I prefer this than using malloc and reorganising the whole board, too bad there was some errors :(

chall06/cdorta-n.c
Top64: Yes, just saw the If statement was wrong and didn't check the first line.

chall06/cdorta-n.c
Top64: I'm trying to solve this error... but in my case, it returns 1, the other problem of the if statement does not affect this one, so I don't know.

aguimbao: This was my fault, it returns `1` correctly, sorry :(

Top64:
1441342441unknown
unknown
u1223nknown
Anotación 2020-04-23 125534

Hello again!  I'm really grateful for your correction. I take note of everything to improve my code! But the wrongs returns you tell me, It doesn't happen to me. Well, nothing important, thank you very much!

aguimbao:
Yes, my tests were faulty, I excuse myself for this. We will upload the tests in ~30 mins or so so you can try them yourself !

PR #211, opened 2020-04-20
fgata-va's solution

FranciscoJesusGata:
# How do I solved it
I analyze the board searching for any problem:
- There is no king
- There is no piece apart from the king
 - The board is not square and is greater than 9
- There is a invalid char

Then I allocate a write a matrix with the board. Once the board is allocated I search for every enemy and with their coordinates (instead of do a loop for every direction, I use a variable for the distance of the cell from the piece to check 4 cells at once), I see the possible movements and if the king is in one of those cases. Before the program finishes I free the matrix and return the state of the king and the board.


chall06/fgata-va.c
```diff
@@ -0,0 +1,162 @@
...
+	j = 0;
+	while (board[j] != '\0')
+	{
+		if(board[j] == '\n')
+			len++;
+		j++;
+	}
+	if(!(matrix = (char**)malloc((len + 1)* sizeof(char*))))
+		return (NULL);
+	matrix[len + 1] = "\0";
```
aguimbao, line 82: I think it's better to `matrix[len + 1] = NULL`.

You have a heap-buffer-overflow here on line 82. You can check it with `-fsanitize=address` option for `gcc` for `board = "...\nR.K\n..."` for example.

aguimbao (commented):
Your code was really hard to test and debug.
Mainly because you messed up with your memory allocations, there was also a couple of double-free's here and there.
You can test it yourself, I will give you some test cases.
Your allocations fail for `board = ".R.....K\n........\n........\n........\n........\n........\n........\n........"`. 

There's something wrong with your allocation and it ends up corrupting the next allocations (we do consecutive tests). It's nice that you free your variables, but given that they are corrupted, they are crashing your whole program.
Maybe you can resubmit when you fix it and we can take a look?

I like how you use a variable for the distance of the cell from the piece to check 4 cells at once, it's very smart. However, this exercise could be solved without `malloc()`, this would have been much better for you, because your evaluating logic is very nice, but it's the allocation part that is making your program not perform correctly. It requires good math, but it could be solved on the original `board` directly without re-organising it.

If you ever resubmit it, some good practices could be:

- Using `static` keyword more often for your functions.
- Try to use structs, some of your functions have too many variables, even if there's no norm here.
- I think `ft_movements()` can be re-organised and simplified in a more understandable group of functions.

For the rest, good job.

PR #212, opened 2020-04-20
Asgard - vsempere

victorsempere:
First of all the check of the map
Convert the P, B, R, Q chars to 1, 2, 4, 8 to be able to realize bitwise operations to check
Then starting from the Kings position I start to check the rings until I find a solution


victorsempere:
I forgot to fix something

PR #213, opened 2020-04-21
Asgard -  vsempere

victorsempere:
First of all the check of the map
Convert the P, B, R, Q chars to 1, 2, 4, 8 to be able to realize bitwise operations to check
Then starting from the Kings position I start to check the rings (north, north-east, east, ...) until I find a solution or check the whole board

### DIFF
--- a/chall06/vsempere.c
+++ b/chall06/vsempere.c
@@ -116,19 +116,24 @@ static int      ft_check_map(char *board)
     return (king != 1 ? 0 : 1);
 }
 
-int             ft_check_mate(char *board)
+int             ft_check_mate(char *board_param)
 {
     int hit;
     int king;
     int board_side = 0;
     int ring;
+    char *board;
 
-    if (!ft_check_map(board))
+    if (!ft_check_map(board_param))
         return (1);
+    if (!(board = malloc(sizeof(char) * strlen(board_param))))
+        return (1);
+    strcpy(board, board_param);
     ft_prepare_map(board, &king, &board_side);
     ring = 1;
     while (!(hit = ft_check_mate_ring(board, king, ring, board_side)))
         ring++;
+    free(board);
     if (hit == 1)
         return (0);
     else
--- a/chall06/vsempere.c
+++ b/chall06/vsempere.c
@@ -1,0 +1,136 @@
+#include <string.h>
+#include <stdlib.h>
+
+#define VSEMPERE_P 1
+#define VSEMPERE_B 2
+#define VSEMPERE_R 4
+#define VSEMPERE_Q 8
+
+static void ft_prepare_map(char *board, int *king, int *board_side)
+{
+    int offset;
+    int real;
+
+    real = 0;
+    offset = 0;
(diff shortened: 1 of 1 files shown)


chall06/vsempere.c
```diff
@@ -0,0 +1,218 @@
...
+{
+    int hit;
+    int king;
+    int board_side = 0;
+    int ring;
+    char *board;
+
+    if (!ft_check_map(board_param))
+        return (1);
+    if (!(board = malloc(sizeof(char) * strlen(board_param))))
```
aguimbao, line 206: You forgot to add the `+ 1` for the trailing `\0` on the malloc argument.

chall06/vsempere.c
```diff
@@ -0,0 +1,218 @@
...
+        return (1);
+    strcpy(board, board_param);
```
aguimbao, line 208: `strncpy` > `strcpy`. Your `strcpy` call heap-overflow's because you miscalculated the size for the malloc :( this is very dangerous, you are writing in random memory.

This leads to all sort of undefined behaviour. Because of this invalid write, you probably fail the checks for `board = "K.....\n......\n......\n......"`. You return `0`. This board is obviously invalid, and also the King is safe, so it's a double error.

chall06/vsempere.c
```diff
@@ -0,0 +1,218 @@
+#include <string.h>
+#include <stdlib.h>
+
+#define VSEMPERE_P 1
```
aguimbao, line 4: Maybe use some more generic names? Like without your login, but this is a bit subjective.

chall06/vsempere.c
```diff
@@ -0,0 +1,218 @@
...
+        ft_check_south_west(board, ring, board_size, king) ||
+        ft_check_north_west(board, ring, board_size, king) ||
+        ft_check_north_east(board, ring, board_size, king) ||
+        ft_check_north(board, ring, board_size, king) ||
+        ft_check_south(board, ring, board_size, king) ||
+        ft_check_east(board, ring, board_size, king) ||
+        ft_check_west(board, ring, board_size, king) ? 1 : 0);
+}
+
+static int      ft_check_map(char *board)
```
aguimbao, line 163: There's some missing checks in this function, like for example, a `\n` character at the end or the beginning of the board would make it invalid, or also this `board = "K..\n\n...\nR.."` is invalid cause of double `\n`. But this tests are a bit hardcore so I'm not sure they will be added.

aguimbao (commented):
I really like your code. You have good mastery of keywords. I was waiting to see some `#define`s/`typedef`s in some of your submissions. Also this code was a good fit a nice structure, but I'm not sure it was 100% the best approach. Nice structure for your code too, it's really easily understandable.

I think your logic is nice, checking from the king is the good way to go, instead of checking every cell in the board.

It's just sad the malloc corrupted your checks. This exercise could be solved without `malloc()`, it would have been less riskier, but harder to perform the checks.

All in all, good job !

PR #214, opened 2020-04-21
chall06 - gbudau

gbudau:
First validate the board string to make sure the string contains only of the allowed characters and it's a "square" less than 9x9.

I made three functions to check if a `PAWN`/`ROOK`/`BISHOP`/`QUEEN` intersect the `KING`. For the `QUEEN` i'm using same functions as for `ROOK` + `BISHOP`.
If any of those function `return 0` it means the piece intersects with the KING and will return 0, else will `return 1`.

**Short description of your code:**
`ft_check_mate` 
Check if the string is NULL or empty then `validate_board`.

`validate_board`
Using `size` to store first line length, this will be the size of the board if valid, N x N.
Use `strchr()` to compare all characters with the `VALID_PIECES`
Store the length of the first line to `size` and verify to not be more than 9 (`MAX_SQUARE_SIZE`)
When encounter a `LINE` increment the `lines` variable and compare it with the first line, 
then reset the `line_length` variable and verify to have less lines than the length of first line.
When encounter a `KING` increment the `king_found` variable and check to not be different than 1 (Only one king allowed).
Increment the `line_length` only when not pointing at a line character.
If there is only one line (no new line character), assign the total length of the string to size, check if the size is less than `MAX_SQUARE_SIZE`, check if there is one `KING`, and the number of lines coincide with the `size` (both should be 1 in this case).
If any condition fails here will `return 0`, else will `return 1`

If board is valid pass to `check_board`.
While iterating on the `board` string whenever an `ATTACKER` is found the coresponding function will be called.

`verify_pawn`
By dividing the position in the string with the length of the row we get the row number.
Using modulo we get the column number.
So for example here we only check the upper row if the `PAWN` is not in the first row.
And we check the right side if its not in the first column.
Check the left side if the `PAWN` is not in the last column.
`Return 0` if its intersecting with the KING, 1 otherwise.

`verify_rook`
Using division to get row position and modulo to get column position.
`diff `is the difference between the closest "wall" and the `ROOK` so for example when checking upward is the diference between the upper limit (first row) and the current row, but since the first row is 0 its the current row.
And checking upward, downward, left and right.
Also making sure there is no other `ATTACKER` in the path.
`Return 0` if its intersecting with the `KING`, 1 otherwise.

`verify_bishop`
Its the same as `verify_rook` but in diagonals and the limit will be the closest "wall".
For example for the upper-right corner will be the minimum between the position - first row and last column - position.
Check upward-right(NE), upward-left(NW), downward-right(SE) and downward-left(SW).
Also making sure there is no other `ATTACKER` in the path.
`Return 0` if its intersecting with the `KING`, 1 otherwise.

### DIFF
--- a/chall06/gbudau.c
+++ b/chall06/gbudau.c
@@ -6,7 +6,7 @@
 /*   By: gbudau <gbudau@student.42.fr>              +#+  +:+       +#+        */
 /*                                                +#+#+#+#+#+   +#+           */
 /*   Created: 2020/04/20 23:10:50 by gbudau            #+#    #+#             */
-/*   Updated: 2020/04/21 08:10:45 by gbudau           ###   ########.fr       */
+/*   Updated: 2020/04/21 08:44:37 by gbudau           ###   ########.fr       */
 /*                                                                            */
 /* ************************************************************************** */
 
@@ -56,7 +56,7 @@ static int	validate_board(char *board, int *size)
 		i++;
 	}
 	if ((*size == 0 && (*size = i) && *size > MAX_SQUARE_SIZE) ||
-		king_found != 1 || line_length != *size || lines != *size)
+		king_found != 1 || lines != *size)
 		return (0);
 	return (1);
 }
--- a/chall06/gbudau.c
+++ b/chall06/gbudau.c
@@ -1,0 +1,288 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   gbudau.c                                           :+:      :+:    :+:   */
+/*                                                    +:+ +:+         +:+     */
+/*   By: gbudau <gbudau@student.42.fr>              +#+  +:+       +#+        */
+/*                                                +#+#+#+#+#+   +#+           */
+/*   Created: 2020/04/20 23:10:50 by gbudau            #+#    #+#             */
+/*   Updated: 2020/04/21 08:10:45 by gbudau           ###   ########.fr       */
+/*                                                                            */
+/* ************************************************************************** */
+
+#include <string.h>
+#define VALID_PIECES	"PBRQK.\n"
+#define ATTACKERS		"PBRQ"
+#define KING 			'K'
+#define PAWN 			'P'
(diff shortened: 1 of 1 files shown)


aguimbao (commented):
Ok this is a really good piece of code, for the moment it's the best one i've seen for this challenge.
Really good naming, really good keyword usage, good definitions, almost looks like it's normed, it's very well commented, looks like it's self-documented code even, you don't fail any test, even the most vicious ones, you board sanitation is really good, and you don't go out of bounds a single time. And all of that without using `malloc()`, which is what gave problems to most of your peers.
Also, really well explained.

Maybe the only thing I would do differently is try to find a smart way to reduce the size of your `verify_rook()` and `verify_bishop()` functions, but that would take some time, and they already are very well organised.  
Also maybe use `size_t`s more often, I think it's good practice.
You could have also tried to implement a struct of some kind, but honestly, I don't think it was necessary the way you approached it, you managed your parameters really well.

Really really good job !

aguimbao:
Nvm we actually found a check you don't pass 😜.

```
.K....PQ
P.......
........
........
........
........
........
.......
```

This should return invalid (`1`), because the length is not equal for all the rows, but being the last row, and not having a `\n` appended at the end, you actually `+ 1` it's length, giving it the same length as the other rows, and actually bypassing your row length check.

gbudau:
Thank you a lot Abraham, i appreciate your feedback and kind words!

I apologize for packing many things into `validate_board`  and `verify_X` functions, tried to refactor them before the deadline but i was doing more harm than good, i've been awake for too long at that point.

Also now i realize that checking all the pieces was not the best approach, could have saved the king position and only check if there are any `ROOK`/`QUEEN` in his vertical+horizontal path, `BISHOP`/`QUEEN` in diagonals or `PAWN` one row below on the right and left side of his column.

 
> Nvm we actually found a check you don't pass

Well spotted, thank you! :+1: 

I didn't make a test for that case and when writing the PR description i believed that one check (`line_length != *size`) was unnecessary so i removed it and commited again, lesson learned.

PR #217, opened 2020-04-21
ChessMate ssacrist

esepunto:
First of all, I'm sorry about the written code (and my English). I don't master loops and other functions yet, which sometimes means that the code is complicated to understand.

I've tried to solve the problem **without working with a 2D-matrix.** 

In this way, I have tried to find the algorithms that, **just by going through the received string from main**, would inform us if a certain piece was giving check to the King. 

I think it's more difficult but also much more interesting. Unfortunately, it is very difficult to me to check if the function works in every case. So I keep my fingers crossed... ;)

The function first performs various checks to make sure that the pieces, the size of the board, etc. are correct.

Once these checks have been made...
Each piece requires a different algorithm to find out if it checks the King (except the Queen, which uses the Bishop and Rook algorithms). 

**The Bishop algorithm:** 
To find out if it checks the King, it must check the following positions of the character string. From the initial position of the Bishop, and without counting the positions = '\n' we perform these operations. Every time we make an addition or a subtraction, we check if in that square is the King, in which case we return (0).
- We add successively the variable columns_sum, whose value = nº of columns + 1.
- We successively subtract the variable columns_sum mentioned.
- We successively add the variable columns_minus, whose value = no. of columns - 1.
- We successively subtract the variable columns_minus mentioned.

A series of exceptions must also be applied when the Bishop is located at one of the ends of the board. 

**The Rook's algorithm:**
From the initial position, and without counting the '\n' positions:
- We successively add the columns variables, whose value = number of columns on the board. This way we can check if the King is in the same column as the Rook.
- We add and subtract one by one until arriving at "\n". This is how we check if the King is in the same row as the Rook.

**The Pawn algorithm:** 
A simpler version of the Bishop's algorithm.
We just go back in the string, and only subtract once column_sum and once columns_minus. 
Exceptions also apply when the pawn is at one end of the board. 

As I said before, the code does not solve all situations, as it is not able to handle when another piece comes between the King and the piece that is giving it check. (Solved at the last minute in the last commit).

I think that by fine-tuning the loops, the code could be more efficient and take up much fewer lines. Maybe some other time ;)


esepunto:
**Well, after checking my code I found a mistake. I'm mentioning it so you don't have to look for it.** 

This solution checks if there is a different "piece" of the King or of the space (.) in the "path" of the attacking piece. If it finds it, it stops the loop. The error is that it also checks the initial position of the attacking piece, which is obviously neither a King nor a space (.) 

Example: in this position it checks the positions marked with X, but it also checks the initial position (B). Therefore it always finds a piece "between". 

...X
..X.
.X..
B...

The correction is very simple. just add: `&& i != init_pos`. 
```
if (str[i] != '.' && str[i] != 'K' && i != init_pos)
    break;
```
My problem is that I changed this last one at 10:40am, confident it would work, no time to test. It's my fault. What a pity!

_Addenda:_ my function not runs well in some cases with Bishop in 2x2 board and when pawn is in lower-right corner. Examples:

K  .
B  .    
(My result == 0)


.   .   .
.  K   .
.   .   P
(My result == 1)


K  .
.   P
(My result == 1)

chall06/ssacrist.c
```diff
@@ -0,0 +1,430 @@
...
+/*
+** To check errors from pieces
+*/
+int    ft_check_errors(char *str)
+{
+    if (check_board(str) == 0)
+        return(1);
+    
+    /* To find unsuppoerted pieces */
+    if (count_pieces('P', str) + count_pieces('B', str)
```
aguimbao, line 357: This is too slow, probably would have been better to verify that the King is present with `strchr()`, then look for at least one enemy piece with `strchr()` for all supported pieces (but once one is found, you stop), and then you go through the board, and if you find an unsupported piece, you exit.

chall06/ssacrist.c
```diff
@@ -0,0 +1,430 @@
...
+
+    if (ft_check_errors(str) == 1)
+        return(1);
+    
+    /* Size is the number os colums/rows of the board */
+    size = ( ( strlen(str) ) / ( count_pieces('\n', str) + 1) );
+    j = 1;
+    
+    /* Are Pawns check the King? */
+    i = count_pieces('P', str);
```
aguimbao, line 386: Here and in all the rest of `count_pieces()` calls in this function, you could have used `strchr()`, much more efficient.

aguimbao (commented):
Ok, first of all, really good job. The hard way to do the exercise was without `malloc()`, a lot of your peers crashed their programs because they didn't use `malloc()` or `free()` correctly, so that's an advantage of not using them.

I see you saw your own fail, it happens also for Queens for diagonals. That's a shame :/. Also I think you have some collision problems as you said. For example for `board = ".K....PQ\n........\n........\n........\n........\n........\n........\n........"`. It doesn't check that the Pawn is in the middle of the Queen and the King.

A really smart approach to this exercise, that would eliminate a lot of your redundant code and excessive loops, would be to save the King position at the beginning, and perform checks, based on his position, because the king can only be attacked from the front, back, sides and diagonally, you don't really have to check all the board all the time.

Also remember to use `static` keyword for your functions, it's good practice.
Check the code from gbudau, it's really well written and you will learn a lot <3
https://github.com/42madrid/remote-challs/pull/214

You can think about that and try to reduce your lines :) good job though. And don't worry, your english is very good.

esepunto:
Thanks, problems with the Queen are caused because I was an incredible error of typing in the line      `  pos = find_position('R', j, str);`, because I typed R instead of Q. OMG!!

```
    /* Are Queen check the King? */
    j = 1;
    i = count_pieces('Q', str);
    while (j <= i)
    {
        pos = find_position('R', j, str);
        if (B_checkmate(size, pos, str) == 0)
            return(0);
        if (R_checkmate(size, pos, str) == 0)
            return(0);
        j++;
    }
    return(1);
}
```

Many thanks for all your comments... even though I sometimes forget them... (for example to avoid the use of strlen several times in the same function by replacing it with a variable , as you say me in a previous chall)

PR #218, opened 2020-04-21
Chall06 by Thor

fvega-tr:
First I check errors, then I divide the string into a matrix, I fix the position of the king, to later check first if there is a check position of the queen or the rook in line, later of the queen or bishop on the diagonal and finally the pawn.
I put it in a way and in the end I ended up changing the approach perhaps that's why the code is not as good as it should be but at least it has helped me to try to improve

### DIFF
--- a/chall06/fvega-tr.c
+++ b/chall06/fvega-tr.c
@@ -1,0 +1,281 @@
+#include <unistd.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+
+typedef struct	s_struct
+{
+				int column;
+				int row;
+				int len_row;
+				int len_column;
+				char **board;
+}				t_struct;
+
+size_t	ft_strlcpy(char *dst, const char *src, size_t dstsize)
+{
+	size_t	i;
+	size_t	j;
+	char	*source;
+
+	i = 0;
+	j = 0;
+	if (dst == NULL)
+		return (dstsize);
+	source = (char *)src;
+	while (source[j] != '\0')
+		j++;
+	if (dstsize > 0)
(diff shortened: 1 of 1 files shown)


chall06/fvega-tr.c
```diff
@@ -0,0 +1,281 @@
...
+char		**ft_split(char const *s, char c)
+{
+	int		i;
+	int		size;
+	int		j;
+	char	**pointer;
+
+	i = 0;
+	size = ft_wordscounter(s, c, &i);
+	if (!(pointer = (char **)malloc(sizeof(char *) * (size + 1))) || size == -1)
```
aguimbao, line 87: I would inverse the parameters of this condition.

chall06/fvega-tr.c
```diff
@@ -0,0 +1,281 @@
...
+			if (t->board[t->row + 1][t->column - 1] == 'P')
+				return (1);
+	}
+	return (0);
+}
+
+int ft_check_mate(char *str)
+{
+	t_struct t;
+	int i;
```
aguimbao, line 264: Not used variables? 😲

aguimbao (commented):
Good and safe approach. I like the usage of a struct, it's not the most performant way to solve it but it works. 

- You have a lot of leaks at the end as you end up not freeing your splitted board, this could be bad if we performed a lot of consecutive tests.
- I would also use `static` for all your functions, it's a good practice.

Overall it's a good approach, checking from the king's position. And your code is really easy to understand. :) Good job !

aguimbao:
Sadly your PR was out of the deadline ;( can't merge it.

PR #219, opened 2020-04-21
Chall06. mzomeno-

veleita:
I couldn't make it in time again. I am so sorry please revise my code :(

This is how my function works.
- Checks that the argument is non-null.
- Counts the length of the first line and saves it in a variable.
- Counts the number of lines. Checks that this number is equal to the length of the line. Also checks that all lines have the same length and that there are no invalid pieces.
- Checks that there is one and only one king and records its position.
- For every piece it encounters, checks if the king is in capture position for that piece.

### DIFF
--- a/chall06/README.md
+++ b/chall06/README.md
@@ -1,0 +1,119 @@
+<h1 align="center"><code>Asgard (Chall06) / CheckMate Problem</code></h1>
+
+<div align="center">
+  <sub>Created by <a href="https://github.com/abguimba">Abraham Guimbao (lincoln)</a>, <a href="https://github.com/vesta-nna">Anna Dominguez (analogia)</a>, <a href="https://github.com/clafoutis42">Charles Labourier (pistache)</a></sub>
+</div>
+<div align="center">
+  <sub>From <a href="https://42madrid.com">42 Madrid</a> for all the 42 Network</sub>
+</div>
+
+---
+
+### Instructions
+<sub>**Turn-in directory:** `chall06/`</sub><br />
+<sub>**Files to turn in:** `<xlogin>.c`</sub><br />
+<sub>**Language:** `C`</sub><br />
+<sub>**Prototype:** `int ft_check_mate(char *board);`</sub>
+
+<sub>**Deadline:** 21.04.2020 - 11:42am</sub>
+<br /><br />
+
+### Subject
+
+<p align="center">
+  <img src="https://i.imgur.com/zKzCCoN.jpg">
+</p>
+
+Write a function that takes a chessboard as argument and checks if your 
+King is in a check position.
+
+Chess is played on a chessboard, a squared board of 8-squares length with 
+different pieces on it.
+
+For this exercice, you will only play with Pawns (`P`), Bishops (`B`), Rooks (`R`), Queen (`Q`), and obviously a King (`K`). `.` is considered an empty space.
+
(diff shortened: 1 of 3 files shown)


chall06/README.md
```diff
@@ -0,0 +1,119 @@
+<h1 align="center"><code>Asgard (Chall06) / CheckMate Problem</code></h1>
```
aguimbao, line 1: Don't submit other files than the ones that were asked for in your PR next time or we won't merge it :(

chall06/mzomeno-.c
```diff
@@ -0,0 +1,173 @@
...
+
+static int rook_capture(int line_length, int line_number,
+			int position, int king_position)
+{
+  int n;
+
+  n = 1;
+  while (n < (line_number - (position - line_length)))
+    {
+      if (king_position == (position + n * (line_length + 1)) ||
```
aguimbao, line 104: There's probably an error in this condition and you never find the King ('`K`') in the same line as a Rook ('`R`'). Probably should have stored the line the king is in for an easier Rook ('`R`') check.

chall06/mzomeno-.c
```diff
@@ -0,0 +1,173 @@
...
+  return (king_position);
+  /*
+   * Note that the function will return -1 if there is no king,
+   * as the king_position variable’s initial value remains untouched
+   */
+}
+
+static int pawn_capture(int line_length, int position, int king_position)
+{
+  if (king_position == (position - line_length - 2) ||
```
aguimbao, line 64: The Pawn can only capture the King from "the bottom" (check the capture patterns again). Which means `board = "P.\n.K"` should return `1` but and `board = "K.\n.P" should return `0`.

chall06/mzomeno-.c
```diff
@@ -0,0 +1,173 @@
...
+   ** - Record king position
+   **/
+  king_position = check_king(board);
+  if (king_position == -1)
+    return (1);
+  /*
+   * Check mate
+   */
+  position = 0;
+  while (board[position] != '\0')
```
aguimbao, line 155: This way of going through the board can give you some problems, without resetting `position` when you go through `\n`'s I think. For example, you fail your Pawn check for `board = ".K....PQ\nP.......\n........\n........\n........\n........\n........\n........"` .

aguimbao (commented):
Hey, sorry for taking a long time to finish your correction.

First of all, good job, don't worry if you submit after the deadline, we will still take a look at your code. I won't be able to merge it, mainly because of the deadline, but also because you are submitting changes to another file (`README.md`), so be careful next time 😅 .

I commented some of the errors I saw in your code. Also you forgot to take into account collisions between pieces, this is very important for all the pieces except the Pawn.
I really liked that you tried without to solve it without using `malloc()`, it gave a lot of problems to your friends in this last challenge, and it was much harder to do it without allocating memory and rearranging the board. For passing information around, probably using a well defined struct would have helped you, I think structs help overall readability even if they can impact performance, I think it's usually a decent trade off. You could have used it to store the King's and other piece's positions. You maybe could have used a couple `typedef`s or `#defines`s for your common conditions in your checks.

Last of all, in your approach, you check every single case in the board, in a smarter approach, you could have saved the king's position, and knowing that he can only be captured from the front, back, sides, and diagonally, you wouldn't have had to check every case, but more like, going through these paths from the King's position, trying to find a corresponding enemy piece, would have made your code much more performant.

PR #220, opened 2020-04-23
chall07 : estina

xplore93:
The idea is to split the response line and then passing Atoi to each value, transform it to hexa and concatenate to a string. Then join the id and that hexa to send the message (with GET). Then print the response of the server.

### DIFF
--- a/chall07/estina.go
+++ b/chall07/estina.go
@@ -1,0 +1,36 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strconv"
+	"strings"
+)
+
+func throw_error(err error) {
+	if err != nil {
+		panic(err)
+	}
+}
+
+func main() {
+	resp, err := http.Get("https://chall07.42madrid.com/")
+	throw_error(err)
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
+	url := strings.Split(string(body), " ")
+	result := strings.Split(string(url[0]), ",")
+	hexa_color := ""
+	for i := 1; i < 4; i++ {
+		color, err := strconv.Atoi(result[i][2:])
+		throw_error(err)
+		hexa_color = hexa_color + fmt.Sprintf("%x", color)
+	}
+	final := "https://chall07.42madrid.com/?" + result[0] + "&resp=" + hexa_color
+	send, err := http.Get(final)
+	throw_error(err)
+	defer send.Body.Close()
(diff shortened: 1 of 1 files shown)


chall07/estina.go
```diff
@@ -0,0 +1,48 @@
...
+	"strings"
+)
+
+func throw_error(err error) {
+	if err != nil {
+		panic(err)
+	}
+}
+
+func send_get_from_server(s string) []byte {
```
clafoutis42, line 29: The norm that we all use in Go is `PascalCase` (for exported functions, which can be used from outside) and` camelCase` (for functions that cannot be used from outside). This means that in this case, your function should be called `sendGetFromServer` because it is a function from the main that you don't want to export.

chall07/estina.go
```diff
@@ -0,0 +1,48 @@
...
+
+func main() {
+	url := strings.Split(string(send_get_from_server("https://chall07.42madrid.com/")), " ")
+	result := strings.Split(string(url[0]), ",")
+	hexa_color := ""
+	for i := 1; i < 4; i++ {
+		color, err := strconv.Atoi(result[i][2:])
+		throw_error(err)
+		hexa_color = hexa_color + fmt.Sprintf("%x", color)
+	}
```
clafoutis42, line 45: A most simple way of doing this would have been by using [fmt.Sscanf](https://golang.org/pkg/fmt/#Sscanf). As you know exactly how the input should be:
```suggestion
	var id, red, green, blue int
	body := send_get_from_server("https://chall07.42madrid.com/")
	_, err := fmt.Sscanf(string(body), "id=%d,r=%d,g=%d,b=%d - Send your response here: chall07.42madrid.com/?id=<id>&resp=<hex>", &id, &red, &green, &blue)
	throw_error(err)
	body = send_get_from_server(fmt.Sprintf("https://chall07.42madrid.com/?id=%d&resp=%02x%02x%02x", id, red, green, blue))
	fmt.Println(string(body))
```

clafoutis42 (commented):
Congrats! You have a working solution. There is still some improvements you could do by taking advantages of the golang's standard lib.  
There is a naming convention in Golang, you should always use `PascalCase` or `camelCase`, `snake_case` will look weird among the whole Golang norm ;)

chall07/estina.go
xplore93: Good to know that. Thank you!

chall07/estina.go
xplore93: It's difficult to understand the code, but I will try to reproduce that.

PR #221, opened 2020-04-23
Chall07-finished

migferna42:
### Learning Go
To learn a little about Go and how to make requests I have used this Medium post: https://medium.com/@masnun/making-http-requests-in-golang-dd123379efe7

### First impressions
I've been seeing that almost everything returns 2 parameters, the success case and the error case.The other thing that has caught my attention is that you have to "close" the request to get

### How does it work
As you have to make 2 requests to the same url, I found it useful to create a function to make the requests, that we pass it by argument.
Each request to the server returns a response, so we pass the output of the first Get as input to the rgbToHexadecimal function.

The rgaToHexadecimal function doesn't do much, it just cleans up the string by spliting over the spaces, ',' and '='. and stores the useful data in the variables that come out of the function.

Finally Sprintf allows us to format an int to hexadecimal, using the arguments %02x.
And we send this data to the server and vuala!, We receive an Ok

### DIFF
--- a/chall07/migferna.go
+++ b/chall07/migferna.go
@@ -1,0 +1,47 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"strconv"
+	"strings"
+)
+
+func rgbToHexadecimal(response string) (int, int, int, int) {
+	str := strings.Split(response, " ")
+	data := str[0]
+	values := strings.Split(data, ",")
+	id, error := strconv.Atoi(strings.Split(values[0], "=")[1])
+	red, error := strconv.Atoi(strings.Split(values[1], "=")[1])
+	green, error := strconv.Atoi(strings.Split(values[2], "=")[1])
+	blue, error := strconv.Atoi(strings.Split(values[3], "=")[1])
+
+	if error != nil {
+		log.Fatalln(error)
+	}
+
+	return id, red, green, blue
+}
+
+func makeRequest(url string) string {
+	response, error := http.Get(url)
+	body, error := ioutil.ReadAll(response.Body)
+
+	if error != nil {
+		log.Fatalln(error)
+	}
+
+	defer response.Body.Close()
+	return string(body)
+}
+
+func main() {
+	urlBase := "https://chall07.42madrid.com/"
+	body := makeRequest(urlBase)
+	id, red, green, blue := rgbToHexadecimal(body)
+	urlResponse := fmt.Sprintf(urlBase+"?id=%d&resp=%02x%02x%02x", id, red, green, blue)
+	body = makeRequest(urlResponse)
+	fmt.Println(body)
+}


xplore93:
I like how you have structured it.

chall07/migferna.go
```diff
@@ -0,0 +1,47 @@
...
+	data := str[0]
+	values := strings.Split(data, ",")
+	id, error := strconv.Atoi(strings.Split(values[0], "=")[1])
+	red, error := strconv.Atoi(strings.Split(values[1], "=")[1])
+	green, error := strconv.Atoi(strings.Split(values[2], "=")[1])
+	blue, error := strconv.Atoi(strings.Split(values[3], "=")[1])
+
+	if error != nil {
+		log.Fatalln(error)
+	}
```
clafoutis42, line 23: A most simple way of doing this would have been by using [fmt.Sscanf](https://golang.org/pkg/fmt/#Sscanf). As you know exactly how the input should be:
```suggestion
	var id, red, green, blue int
	if _, err := fmt.Sscanf(response, "id=%d,r=%d,g=%d,b=%d - Send your response here: chall07.42madrid.com/?id=<id>&resp=<hex>", &id, &red, &green, &blue); err != nil {
		log.Fatalln(err)
	}
```

chall07/migferna.go
```diff
@@ -0,0 +1,47 @@
...
+	"net/http"
+	"strconv"
+	"strings"
+)
+
+func rgbToHexadecimal(response string) (int, int, int, int) {
+	str := strings.Split(response, " ")
+	data := str[0]
+	values := strings.Split(data, ",")
+	id, error := strconv.Atoi(strings.Split(values[0], "=")[1])
```
clafoutis42, line 16: It is wrong to call a variable `error` in Go. `error` is the [builtin error type interface](https://golang.org/pkg/builtin/#error), which means that from this point in your function you can't use this type anymore.  
Here it doesn't cause any problems, but it might in fact cause a lot of problem on error handling on more complicated projects.

It is common to call it `err`.

chall07/migferna.go
```diff
@@ -0,0 +1,47 @@
...
+
+	if error != nil {
+		log.Fatalln(error)
+	}
+
+	return id, red, green, blue
+}
+
+func makeRequest(url string) string {
+	response, error := http.Get(url)
```
clafoutis42, line 29: Again, watch out on the error variable naming ;)

chall07/migferna.go
```diff
@@ -0,0 +1,47 @@
...
+	body, error := ioutil.ReadAll(response.Body)
+
+	if error != nil {
+		log.Fatalln(error)
+	}
+
+	defer response.Body.Close()
```
clafoutis42, line 36: The goal of defer is to close the reader, NO MATTER WHAT HAPPENS. Here if the `ioutil.ReadAll`  fails for some reason you won't be closing the body.
Also, you are not handling the error sent by `http.Get`, which means you will potentially try to access `response.Body` whilst `response` being `nil`. This will cause an unexpected `panic`.

You should handle the error sent by `http.Get` and make a defer right after that:
```suggestion
	response, error := http.Get(url)
	if error != nil {
		log.Fatalln(error)
	}
	defer response.Body.Close()
	
	body, error := ioutil.ReadAll(response.Body)
	if error != nil {
		log.Fatalln(error)
	}
```

clafoutis42 (commented):
Congrats! You have a working solution. However there are some little errors regarding the golang implementation, but that's alright, you'll get to it.

PR #222, opened 2020-04-23
edramire - Planeta Vegeta

edramir18:
The script has two functions, one to log the time of execution and the request/response from the server, and other to create the request and parsed the response from the server. After processed successfully the request the function return the string with the message sending by the server and the status code obtained. To get the value of the rgb parsed the string splited by ',' and '=' character. Using the sprintf function, i convert the number obtenied to an hexadecimal representation. That result is send back to the server with the id recevied, and the response is show on the terminal.

### DIFF
--- a/chall07/edramire.go
+++ b/chall07/edramire.go
@@ -1,0 +1,55 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+)
+
+func logrequest(start time.Time, req string, res string, status int) {
+	duration := time.Since(start)
+	fmt.Printf("%d ms - GET %s\n", duration.Nanoseconds()/1000000, req)
+	if len(res) > 0 {
+		fmt.Printf("\t %d: %s\n", status, res)
+	}
+}
+
+func getrequest(request string) (string, int) {
+	res, err := http.Get(request)
+	if err != nil {
+		log.Fatal(err)
+	}
+	stringBody, err := ioutil.ReadAll(res.Body)
+	res.Body.Close()
+	if err != nil {
+		log.Fatal(err)
+	}
+	return string(stringBody), res.StatusCode
+}
+
+func main() {
+	home := "https://chall07.42madrid.com/"
+	start := time.Now()
+	logrequest(start, home, "", 0)
+	answer, status := getrequest(home)
+	if status != 200 {
+		log.Fatal("Wrong response from server")
+	}
+	logrequest(start, home, answer, status)
+	values := strings.Split(answer[0:strings.Index(answer, " ")], ",")
+	r, _ := strconv.Atoi(values[1][2:])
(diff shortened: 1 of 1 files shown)


chall07/edramire.go
```diff
@@ -0,0 +1,55 @@
...
+	logrequest(start, home, "", 0)
+	answer, status := getrequest(home)
+	if status != 200 {
+		log.Fatal("Wrong response from server")
+	}
+	logrequest(start, home, answer, status)
+	values := strings.Split(answer[0:strings.Index(answer, " ")], ",")
+	r, _ := strconv.Atoi(values[1][2:])
+	g, _ := strconv.Atoi(values[2][2:])
+	b, _ := strconv.Atoi(values[3][2:])
```
clafoutis42, line 46: This is really wrong ! You should always handle errors, and never ignore them, what if the response is not formatted as usual?

clafoutis42, line 46: A most simple way of doing this would have been by using [fmt.Sscanf](https://golang.org/pkg/fmt/#Sscanf). As you know exactly how the input should be:
```suggestion
	var id, red, green, blue int
	if _, err := fmt.Sscanf(answer, "id=%d,r=%d,g=%d,b=%d - Send your response here: chall07.42madrid.com/?id=<id>&resp=<hex>", &id, &red, &green, &blue); err != nil {
		log.Fatal(err)
	}
```

chall07/edramire.go
```diff
@@ -0,0 +1,55 @@
...
+	rgb := (r & 0xFF << 16) | (g & 0xFF << 8) | (b & 0xFF)
+	hex := fmt.Sprintf("%06x", rgb)
+	resp := fmt.Sprintf("%s?%s&resp=%s", home, values[0], hex)
```
clafoutis42, line 49: You don't really need to make this conversion with `rgb`, you could have simply used the `%x` flag of printf:
```suggestion
	resp := fmt.Sprintf("%s?%s&resp=%02x%02x%02x", home, values[0], r, g, b)
```

chall07/edramire.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+}
+
+func getrequest(request string) (string, int) {
+	res, err := http.Get(request)
+	if err != nil {
+		log.Fatal(err)
+	}
+	stringBody, err := ioutil.ReadAll(res.Body)
+	res.Body.Close()
```
clafoutis42, line 27: You should use the `defer` keyword to close `res.Body` no matter what happens. Here if `ioutil.ReadAll` panics for some reason, you will never close `res.Body`
```suggestion
	defer res.Body.Close()
	stringBody, err := ioutil.ReadAll(res.Body)
```
`defer` will automatically run the call it implies at the end of the function, it doesn't matter how it ends, it will call it.

clafoutis42 (commented):
Nice code! Your solution works. There's still a little improvements you could do though.

PR #223, opened 2020-04-23
chall07 - pcuadrad

pabloocg:
Because it is a new language and I had never programmed with it, I wanted to do the output as in the example to find the maximum possible information.
First I set a log that will write to the standard error output in case there were any errors when for example doing the get query or when converting to numbers with Atoi.
I also set a start time using Time.Now() so in the printRequestTime() function I can subtract it from the current time with the Sub() function which returns the difference in ms which is nice.
To parse the response of the get I make several splits, first for the spaces, then for the commas and finally I do them for the '=' to get the number corresponding to the id and the rgb.

### DIFF
--- a/chall07/pcuadrad.go
+++ b/chall07/pcuadrad.go
@@ -26,8 +26,8 @@ func printRequestTime(url string, answer string, start time.Time) {
 }
 
 func DoRequest(url string, l *log.Logger, start time.Time) (string) {
-	response, err := http.Get(url)
 	printRequestTime(url, "", start)
+	response, err := http.Get(url)
 	if err != nil {
 		l.Fatalln(err)
 	}
--- a/chall07/pcuadrad.go
+++ b/chall07/pcuadrad.go
@@ -1,0 +1,68 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strings"
+	"strconv"
+	"log"
+	"os"
+	"time"
+)
+
+func RGBToHex(r int, g int, b int) (string) {
+	hex := fmt.Sprintf("%x", r) + fmt.Sprintf("%x", g) + fmt.Sprintf("%x", b)
+	return hex
+}
+
+func printRequestTime(url string, answer string, start time.Time) {
+	now := time.Now()
+	elapsed := now.Sub(start)
+	if answer == "NO" {
+		fmt.Println(elapsed,"- GET ", url)
+	} else {
+		fmt.Println(elapsed,"- GET ", url, "\n\tanswer:", answer)
+	}
+}
+
+func DoRequest(url string, l *log.Logger, start time.Time) (string) {
+	response, err := http.Get(url)
+	printRequestTime(url, "NO", start)
+	if err != nil {
+		l.Fatalln(err)
+	}
+	defer response.Body.Close()
+	body, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		l.Fatalln(err)
+	}
+	printRequestTime(url, string(body), start)
+	return string(body)
+}
+
+func parseBody(body string, l *log.Logger) (string, int, int, int){
+	var arr [4]int
+	var err error
+	parse := strings.SplitN(body, " ", 2)
+	parse = strings.Split(parse[0], ",")
+	for i := 0; i < len(parse); i++ {
+		getNumbers := strings.Split(parse[i], "=")
+		arr[i], err = strconv.Atoi(getNumbers[1])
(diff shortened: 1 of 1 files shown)


chall07/pcuadrad.go
```diff
@@ -0,0 +1,67 @@
...
+	"net/http"
+	"strings"
+	"strconv"
+	"log"
+	"os"
+	"time"
+)
+
+func RGBToHex(r int, g int, b int) (string) {
+	return fmt.Sprintf("%x%x%x",r ,g ,b)
```
clafoutis42, line 15: Unfortunately, the RGB format asks for a 2 character precision. Which means that `9` should be writtent `09`. To give you an example, with this input `r=189,g=208,b=9` you should output `bdd009`, but the output you send is `bdd09`.  
To fix this it is quite simple, you just have to modify the printf's precision:
```suggestion
	return fmt.Sprintf("%02x%02x%02x",r ,g ,b)
```

chall07/pcuadrad.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func printRequestTime(url string, answer string, start time.Time) {
+	now := time.Now()
+	elapsed := now.Sub(start)
+	if answer == "" {
+		fmt.Println(elapsed,"- GET", url)
+	} else {
+		fmt.Println(elapsed,"- GET", url, "\n\tanswer:", answer)
+	}
```
clafoutis42, line 25: I really like the way you tried to respect the examples' output, I advise you looking at the builtin [Log Package](https://golang.org/pkg/log/) to make better logs. _(I noticed later that you used it, it is better to print logs with the same library,  personally, for big project I use `fmt.Print` functions to debug, but I never let them in my code.)_  
Or even, my favorite log package: [Logrus](https://github.com/sirupsen/logrus), it has many useful features, and a really nice formatting.

chall07/pcuadrad.go
```diff
@@ -0,0 +1,67 @@
...
+	parse := strings.SplitN(body, " ", 2)
+	parse = strings.Split(parse[0], ",")
+	for i := 0; i < len(parse); i++ {
+		getNumbers := strings.Split(parse[i], "=")
+		arr[i], err = strconv.Atoi(getNumbers[1])
+		if err != nil {
+			l.Fatalln(err)
+		}
+	}
+	return strconv.Itoa(arr[0]), arr[1], arr[2], arr[3]
```
clafoutis42, line 55: This is a nice parsing, however there is most simple way of doing this. You could have used [fmt.Sscanf](https://golang.org/pkg/fmt/#Sscanf). As you know exactly how the input should be:
```suggestion
	var id, red, green, blue int
	if _, err := fmt.Sscanf(body, "id=%d,r=%d,g=%d,b=%d - Send your response here: chall07.42madrid.com/?id=<id>&resp=<hex>", &id, &red, &green, &blue); err != nil {
		l.Fatalln(err)
	}
	return  strconv.Itoa(id), red, green, blue
```
Although I don't really understand the need of converting `id` into a string, I let it be this way.

chall07/pcuadrad.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func main() {
+	l := log.New(os.Stderr, "An error ocurred:\n", 0)
+	url := "https://chall07.42madrid.com/"
+	start := time.Now()
+	body := DoRequest(url ,l, start)
+	id, r, g, b := parseBody(body, l)
+	hex := RGBToHex(r, g, b)
+	url = url + "?id=" + id + "&resp=" + hex
```
clafoutis42, line 65: An equivalent to this would have been:
```suggestion
	url += "?id=" + id + "&resp=" + hex
```
You could also have used `fmt.Sprintf`

clafoutis42 (commented):
Nice code! Unfortunately there is a little mistake with the hexadecimal precision. That's too bad, this is a really clean code for a first time with Go :)

chall07/pcuadrad.go
pabloocg: It is true! I didn't pay attention :(

chall07/pcuadrad.go
pabloocg: Okay, thanks for the recommendation!

chall07/pcuadrad.go
pabloocg: I didn't know the existence of that function, so it is so much simpler.
The conversion of the id into string, was more than anything because to concatenate it later in the url string I had to parse string(id) and I wanted it to be as clean as possible.

PR #224, opened 2020-04-23
vde-dios.go

vdedios:
# SOLUTION

We follow a simple flow:

1. Making a GET request to suggested URI
2. Parsing the data. Although I though at first doing it with string parsing functions such as split etc. I decided to try it doing some regexp an actually found it easier to extract the info I wanted.
3. Converting data to hex. Just using `itoa` to convert strings to numbers as well as `Sprintf` to quickly convert from rgb to hex while. I found some problems when using atoi because the substrings I got from the regexp weren't string literals but type `[][]string`. So I had to use the join method `strings.Join(rgb[i],"")` before passing the string to the atoi function.
4. Making a new GET request with the id and hex code obtained from the first request
 

**A quick note,** I just added a simple log whenever a GET request is made to have a minimum tracking of what is happening.

### DIFF
--- a/chall07/vde-dios.go
+++ b/chall07/vde-dios.go
@@ -21,7 +21,7 @@ func toHex(rgb [][]string) string {
     for i := 1; i < 4; i++ {
         hex, err := strconv.Atoi(strings.Join(rgb[i],""))
         outputErr(err)
-        output += fmt.Sprintf("%x", hex)
+        output += fmt.Sprintf("%02x", hex)
     }
     return output
 }
@@ -31,6 +31,7 @@ func MakeRequest(uri string) string{
     outputErr(err)
     body, err := ioutil.ReadAll(resp.Body)
     outputErr(err)
+    defer resp.Body.Close()
     log.Println(string(body))
     return string(body)
 }
--- a/chall07/vde-dios.go
+++ b/chall07/vde-dios.go
@@ -40,5 +40,5 @@ func main() {
     response := MakeRequest("https://chall07.42madrid.com/")
     re := regexp.MustCompile(`\d+`)
     rgb := re.FindAllStringSubmatch(response, -1)
-    MakeRequest("https://chall07.42madrid.com/?id=" + strings.Join(rgb[0],"") + "&resp=" + toHex(rgb))
+    MakeRequest(fmt.Sprintf("https://chall07.42madrid.com/?id=%s&resp=%s", strings.Join(rgb[0],""), toHex(rgb)))
 }
--- a/chall07/vde-dios.go
+++ b/chall07/vde-dios.go
@@ -1,0 +1,43 @@
+package main
+
+import (
+    "net/http"
+    "log"
+    "io/ioutil"
+    "fmt"
+    "regexp"
+    "strconv"
+    "strings"
+)
+
+func outputErr(err error){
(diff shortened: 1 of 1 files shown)


chall07/vde-dios.go
```diff
@@ -0,0 +1,44 @@
...
+    outputErr(err)
+    defer resp.Body.Close()
+    log.Println(string(body))
+    return string(body)
+}
+
+func main() {
+    response := MakeRequest("https://chall07.42madrid.com/")
+    re := regexp.MustCompile(`\d+`)
+    rgb := re.FindAllStringSubmatch(response, -1)
```
clafoutis42, line 42: It seems a little overkill using regular expressions here, but it works.  
I would find more adapted you use [fmt.Sscanf](https://golang.org/pkg/fmt/#Sscanf). As you know exactly how the input should be:
```suggestion
	var id, red, green, blue int
	if _, err := fmt.Sscanf(response, "id=%d,r=%d,g=%d,b=%d - Send your response here: chall07.42madrid.com/?id=<id>&resp=<hex>", &id, &red, &green, &blue); err != nil {
		log.Fatalln(err)
	}
```
I didn't adapat all the code but it gives you a good idea of what to use :)

Later you could simply use `fmt.Sprintf` to format your hexadecimal output` fmt.Sprintf("%02x%02x%02x", red, green, blue)`

chall07/vde-dios.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func toHex(rgb [][]string) string {
+    output := ""
+    for i := 1; i < 4; i++ {
+        hex, err := strconv.Atoi(strings.Join(rgb[i],""))
+        outputErr(err)
+        output += fmt.Sprintf("%02x", hex)
+    }
+    return output
```
clafoutis42, line 26: As stated earlier, you could use something simpler and more readable to replace this function.

chall07/vde-dios.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func MakeRequest(uri string) string{
+    resp, err := http.Get(uri)
+    outputErr(err)
+    body, err := ioutil.ReadAll(resp.Body)
+    outputErr(err)
+    defer resp.Body.Close()
```
clafoutis42, line 34: `defer` helps you run a function, no matter what happens. Here you want to close de response's body when the function exits, would it be because of a panic, an error, etc.  
Which means you should make the `defer` call earlier, for example here if `ioutil.ReadAll` fails for some reason, the body will never be closed
```suggestion
    resp, err := http.Get(uri)
    outputErr(err)
    defer resp.Body.Close()
    body, err := ioutil.ReadAll(resp.Body)
    outputErr(err)
```

clafoutis42 (commented):
Congrats ! Your solutions works fine! You could however improve your golang code a little :)

chall07/vde-dios.go
vdedios: 📝

chall07/vde-dios.go
vdedios: I will definitely look at Sscanf!

About regular expressions, I have heard about it several times... Are regexp more resource consuming when compare with other string parsing methods? I mean in most languages (go, python, js...)

PR #225, opened 2020-04-23
dalba-de Vegeta

dalba-de:
## Planeta Vegeta (Chall07) / HTTP RGB to Hexadecimal

It is one of my first experiences with Golang so I understand that the structuring of the code is not correct, but I still do not understand correctly how the functions work.

First of all I use the start variable to return the number of milliseconds that have passed between each call to "time.Now (). Sub (start) .Milliseconds ()"

Then I make a first call to the url and read the result. I get the id, red, green and blue values ​​using Sscanf and create the hexadecimal value with Sprintf.

I make a second call again with the inclusion of "id" and the hexadecimal value and I get the answer: ok! - ko!

#### Making HTTP Requests in Golang:
- https://medium.com/@masnun/making-http-requests-in-golang-dd123379efe7
- http://polyglot.ninja/golang-making-http-requests/

#### Measuring execution time in Go:
- https://coderwall.com/p/cp5fya/measuring-execution-time-in-go

#### Parsing custom strings:
- https://stackoverflow.com/questions/4278430/convert-string-to-integer-type-in-go

#### RGB to Hex value in Golang:
- https://www.reddit.com/r/golang/comments/adm6l5/do_i_really_need_a_third_party_library_to_get_hex/


clafoutis42:
Well, I have nothing much to say, your code is clean, it shows that you understood Go's logic and your are applying it correctly on a small challenge.  
The only commentary I will make, is that it would help you gain time and make code shorter if you create functions instead of writing everything in the main.

PR #226, opened 2020-04-23
I think it works

byrd92:
### How it starts:

I made a http.get call, saving it on "response", I save the body of the http.call on a variable called : Content. I split all content per " " , then i split the info **(id, r, g and b)** every ",". 

### How I read the info:
i call id = string[0] , that means "id="id_number"" (I dont split the "id=" so I delete this on the call url.

Made atoi of each color, then the best way (and faster) to convert RGB to Hex is using this metod to make it String, I tried some others but they are slower so I cant pass the test.
`fmt.Sprintf("%02x%02x%02x", red, green, blue);`

### Last Call:

I made again a http.call now with the url updated using fmt.sprintf.


chall07/egarcia-.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+func main() {
+	response, err := http.Get("https://chall07.42madrid.com/")
+	contents, err := ioutil.ReadAll(response.Body)
+	body := strings.Split(string(contents), " ")
+	string := strings.Split(string(body[0]), ",")
+	red, err := strconv.Atoi(string[1][2:])
+	green, err := strconv.Atoi(string[2][2:])
+	blue, err := strconv.Atoi(string[3][2:])
```
clafoutis42, line 19: You should ALWAYS handle errors. Especially when you parse an external input. They are there for a reason.

clafoutis42, line 19: A most simple way of doing this would have been by using [fmt.Sscanf](https://golang.org/pkg/fmt/#Sscanf). As you know exactly how the input should be:
```suggestion
	var id, red, green, blue int
	if _, err := fmt.Sscanf(contents, "id=%d,r=%d,g=%d,b=%d - Send your response here: chall07.42madrid.com/?id=<id>&resp=<hex>", &id, &red, &green, &blue); err != nil {
		log.Fatalln(err)
	}
```

chall07/egarcia-.go
```diff
@@ -0,0 +1,32 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+	id := string[0]
+	hex := fmt.Sprintf("%02x%02x%02x", red, green, blue);  
+	context := fmt.Sprintf("https://chall07.42madrid.com/?%s&resp=%s", id, hex)
+	response, err = http.Get(context)
```
clafoutis42, line 26: Again, always handle the errors !

clafoutis42 (commented):
Your code works. You could do some improvement, and you should never forget to handle errors, do not wait for your code to crash!

PR #227, opened 2020-04-23
mpernia-

mmiguelx:
Hello!

1. Set the start variable with time.Now() in order to get the execution time in every print.
2. Transform the info we get by http.Get into a string so can be easily handled.
3. Use the strings.IndexByte to find the first comma so we know the the position so we can use atoi with the id variable.
4. Repeat the same but using the last variable and the next r, g or b respectively.
5. Transform the rgb in hex with the strncov.FormatInt.
6. Return the whole link with the id and hex string.
7. Http.Get and print.


chall07/mpernia-.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"io/ioutil"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+)
+
+func time_elapse(start time.Time) int {
```
clafoutis42, line 11: The norm that we all use in Go is `PascalCase` (for exported functions, which can be used from outside) and` camelCase` (for functions that cannot be used from outside). This means that in this case, your function should be called `timeElapse` because it is a function from the main that you don't want to export.

chall07/mpernia-.go
```diff
@@ -0,0 +1,75 @@
...
+	green, err := strconv.Atoi(body[k+2 : j-1])
+	if err != nil {
+		print(err)
+	}
+	k = strings.IndexByte(body, ' ')
+	blue, err := strconv.Atoi(body[j+2 : k])
+	if err != nil {
+		print(err)
+	}
+	hex := get_hex(red, green, blue)
```
clafoutis42, line 43: There are 2 things you should know here.
- First, it is good to print the errors, but you should exit in someway, make it panic, log a fatal error, it will help prevent unexpected behaviors.
- A most simple way of doing this would have been by using [fmt.Sscanf](https://golang.org/pkg/fmt/#Sscanf). As you know exactly how the input should be:
```suggestion
	var id, red, green, blue int
	if _, err := fmt.Sscanf(body, "id=%d,r=%d,g=%d,b=%d - Send your response here: chall07.42madrid.com/?id=<id>&resp=<hex>", &id, &red, &green, &blue); err != nil {
		panic(err)
	}
	return fmt.Sprintf("https://chall07.42madrid.com/?id=%d&resp=%02x%02x%02x", id, red, green, blue)
```

clafoutis42 (commented):
Your code works, congrats! There is still a few improvements you can make, like the error handling, or the naming convention, but it's good!

PR #228, opened 2020-04-24
chall07 - gbudau

gbudau:
**How you solved this problem / Short description of your code**
Using net/http to send a get request.
Read the response and use regex to create a string array with all the numbers in the string.
Use atoi to convert the 2nd, 3rd and 4th values to int and sprintf to create the answer.
Send a second get request with the generated answer.
Read the response.

Checking for basic errors while doing all the steps.

>PS: 
Not sure if `panic()` was the best option in case of errors, i read a bit about the `log` package.
First time using Go :smile:, seems pretty cool.

### DIFF
--- a/chall07/gbudau.go
+++ b/chall07/gbudau.go
@@ -25,7 +25,7 @@ func requestServer() {
 	}
 	fmt.Printf("%s\n", string(body))
 
-	/* Use regex to find the digits in the string and check to be more than 4 digits */
+	/* Use regex to find the digits in the string and check to be more than 3 digits */
 	/* Expecting the first 4 digits to be in correct order and valid */
 	re := regexp.MustCompile(`\d+`)
 	set := re.FindAllString(string(body), -1)
--- a/chall07/gbudau.go
+++ b/chall07/gbudau.go
@@ -23,7 +23,7 @@ func requestServer() {
 	if err != nil {
 		panic(err)
 	}
-	fmt.Printf("%s\n", body)
+	fmt.Printf("%s\n", string(body))
 
 	/* Use regex to find the digits in the string and check to be more than 4 digits */
 	/* Expecting the first 4 digits to be in correct order and valid */
--- a/chall07/gbudau.go
+++ b/chall07/gbudau.go
@@ -11,19 +11,19 @@ import (
 
 func requestServer() {
 	/* First GET request to get the required formated string */
-    resp, err := http.Get("https://chall07.42madrid.com/")
+	resp, err := http.Get("https://chall07.42madrid.com/")
 	if err != nil {
 		panic(err)
-    }
+	}
 	fmt.Println("Response status:", resp.Status)
-    defer resp.Body.Close()
+	defer resp.Body.Close()
 
 	/* Read and print response */
-    body, err := ioutil.ReadAll(resp.Body)
+	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
 		panic(err)
-    }
-    fmt.Printf("%s\n", body)
+	}
+	fmt.Printf("%s\n", body)
 
 	/* Use regex to find the digits in the string and check to be more than 4 digits */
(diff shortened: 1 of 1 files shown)


chall07/gbudau.go
```diff
@@ -0,0 +1,61 @@
...
+	if len(set) < 4 {
+		panic(err)
+	}
+
+	/* Convert second, third and forth digits to int and create an "answer" string*/
+	/* Expecting all colors to be between 0-255 */
+	r, _ := strconv.Atoi(set[1])
+	g, _ := strconv.Atoi(set[2])
+	b, _ := strconv.Atoi(set[3])
+	answer := fmt.Sprintf("https://chall07.42madrid.com/?id=%s&resp=%.2x%.2x%.2x",set[0] ,r ,g ,b)
```
clafoutis42, line 41: Never ignore errors when you the input comes from outside!

chall07/gbudau.go
```diff
@@ -0,0 +1,61 @@
...
+	set := re.FindAllString(string(body), -1)
+	if len(set) < 4 {
+		panic(err)
+	}
+
+	/* Convert second, third and forth digits to int and create an "answer" string*/
+	/* Expecting all colors to be between 0-255 */
+	r, _ := strconv.Atoi(set[1])
+	g, _ := strconv.Atoi(set[2])
+	b, _ := strconv.Atoi(set[3])
```
clafoutis42, line 40: A most simple way of doing this would have been by using [fmt.Sscanf](https://golang.org/pkg/fmt/#Sscanf). As you know exactly how the input should be:
```suggestion
	var id, r, g, b int
	body := send_get_from_server("https://chall07.42madrid.com/")
	if _, err := fmt.Sscanf(string(body), "id=%d,r=%d,g=%d,b=%d - Send your response here: chall07.42madrid.com/?id=<id>&resp=<hex>", &id, &r, &g, &b); err != nil {
		panic(err)
	}

File: chall07/gbudau.go [hunk 3]
```diff
@@ -0,0 +1,61 @@
...
+	answer := fmt.Sprintf("https://chall07.42madrid.com/?id=%s&resp=%.2x%.2x%.2x",set[0] ,r ,g ,b)
+
+	/* Second GET request using the answer and print the response status*/
+	resp, err = http.Get(answer)
+	if err != nil {
+		panic(err)
+	}
+	fmt.Println("Response status:", resp.Status)
+	defer resp.Body.Close()
```
clafoutis42, line 49: I'm gonna be really about the details here, because this is a print and it should not cause problems, but it would be better to do the `defer` before the `fmt.Println` :joy:
```suggestion
	defer resp.Body.Close()
	fmt.Println("Response status:", resp.Status)
```

chall07/gbudau.go
```diff
@@ -0,0 +1,61 @@
...
+	/* Read and print the response */
+	body, err = ioutil.ReadAll(resp.Body)
+	if err != nil {
+		panic(err)
+	}
+	fmt.Printf("answer: %s\n", string(body))
+}
+
+func main() {
+	requestServer()
```
clafoutis42, line 60: So that's really good you made a function to avoid putting all your code in the main! But maybe next time you could split this function so it is more readable?

clafoutis42 (commented):
Your code works, congrats! However there is still a few improvements you might make :)

Using `panic` was not a _bad_ call, however you are right, it would have been better to use the `log` package.  
Although on a very small code, it is not a real problem ;)

chall07/gbudau.go
gbudau: Thank you, will keep in mind for next time.

chall07/gbudau.go
gbudau: :smile:

That looks so much better and simpler, thank you!

gbudau:
Thank you a lot Charles, appreciate your feedback and advice, also the work you all put into this!

PR #229, opened 2020-04-24
vsempere

victorsempere:
Parse response from server to extract id, r, g, b and url to sendback hex code.
Build hex code
Build url to send
Send request with hex code

### DIFF
--- a/chall07/vsempere.go
+++ b/chall07/vsempere.go
@@ -1,0 +1,21 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+)
+
+func main() {
+	res, err := http.Get("https://chall07.42madrid.com/")
+	if err != nil {
+		log.Fatal(err)
+	}
+	robots, err := ioutil.ReadAll(res.Body)
+	res.Body.Close()
+	if err != nil {
+		log.Fatal(err)
+	}
+	fmt.Printf("%s", robots)
+}
--- a/chall07/vsempere.go
+++ b/chall07/vsempere.go
@@ -2,13 +2,29 @@ package main
 
 import (
 	"fmt"
-	"io/ioutil"
+	"strings"
+	"strconv"
+	"regexp"
 	"log"
+	"io/ioutil"
 	"net/http"
 )
 
-func main() {
-	res, err := http.Get("https://chall07.42madrid.com/")
+func toHex(dec []byte) string {
+	ri, err := strconv.Atoi(string(dec))
+	if err != nil {
+		log.Fatal(err)
+	}
+	str := strconv.FormatInt(int64(ri), 16)
+	if len(str) < 2 {
+		return "0"+ str
+	} else {
+			return str
+	}
+}
+
+func launchGet(url string) (string, string) {
+	res, err := http.Get(url)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


clafoutis42:
Maybe next time you could develop your explanation a little more :/  
Also you are sending an unwanted file, I am gonna have to close the PR. I will still review your code though.

chall07/vsempere.go
```diff
@@ -0,0 +1,59 @@
...
+	ri, err := strconv.Atoi(string(dec))
+	if err != nil {
+		log.Fatal(err)
+	}
+	str := strconv.FormatInt(int64(ri), 16)
+	if len(str) < 2 {
+		return "0"+ str
+	} else {
+			return str
+	}
```
clafoutis42, line 23: There is a much simpler way to do this, you could use `fmt.Sprintf` with flags:
```suggestion
	return fmt.Sprintf("%02x", ri)
```

chall07/vsempere.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+func launchGet(url string) (string, string) {
+	res, err := http.Get(url)
+	if err != nil {
+		log.Fatal(err)
+	}
+	robots, err := ioutil.ReadAll(res.Body)
+	res.Body.Close()
```
clafoutis42, line 32: You could benefit from [defer](https://tour.golang.org/flowcontrol/12) here. It helps you telling the program to run a function when the current function exits. Would that be because of an error, because it returns, because it panics...  
Here you would want to close the body even if `ioutil.ReadAll` panics for some reason:
```suggestion
	res, err := http.Get(url)
	if err != nil {
		log.Fatal(err)
	}
	defer res.Body.Close()
	robots, err := ioutil.ReadAll(res.Body)
```

chall07/vsempere.go
```diff
@@ -0,0 +1,59 @@
...
+
+func buildResponseUrl(response string) string {
+	re := regexp.MustCompile(`id=(\d+),r=(\d+),g=(\d+),b=(\d+).*here:\s+(.*)`)
+	m := re.FindSubmatch([]byte(response));
+
+	id := m[1]
+	r := toHex(m[2])
+	g := toHex(m[3])
+	b := toHex(m[4])
+	s := []string{r, g, b}
```
clafoutis42, line 48: A most simple way of doing this would have been by using [fmt.Sscanf](https://golang.org/pkg/fmt/#Sscanf). As you know exactly how the input should be:
```suggestion
	var id, r, g, b int
	_, err := fmt.Sscanf(response, "id=%d,r=%d,g=%d,b=%d - Send your response here: chall07.42madrid.com/?id=<id>&resp=<hex>", &id, &r, &g, &b); err != nil {
		log.Fatal(err)
	}

File: chall07/vsempere.go [hunk 4]
```diff
@@ -0,0 +1,59 @@
...
+
+	return "https://" + strings.Replace(strings.Replace(string(m[5]), "<id>", string(id), 1), "<hex>", strings.Join(s, ""), 1)
```
clafoutis42, line 50: Why not use `fmt.Sprintf` here?

clafoutis42 (commented):
Your code works, congrats! There are still a little improvements you might make though.

PR #230, opened 2020-05-02
chall08 - pcuadrad

pabloocg:
I have decided to do the exercise using objects to practice and understand them better, I have created three objects (Mine, Robot and Exit) inherited from an Object parent class that simply has the x and y position of the object.
The Robot class is the important one, since in it I have made several methods, one to check if it has collided with a mine, another to know if it has found the exit and to update the current position of the robot.
On the other hand I have created the MineField object which is where I first build the map and save the positions of all the elements it contains, as well as error checking (that there is only one robot and one output and that the square is closed).
Finally for the input commands to move the robot, I pass them one by one to the function MoveRobot so that it updates the position of the robot and checks if it has crashed.

### DIFF
--- a/chall08/pcuadrad.py
+++ b/chall08/pcuadrad.py
@@ -1,0 +1,163 @@
+#!/usr/bin/env python3
+import sys
+
+class Object:
+    def __init__(self, posX: int, posY: int):
+        self.x = posX
+        self.y = posY
+
+class Mine(Object):
+    def __init__(self, posX, posY):
+        super().__init__(posX, posY)
+
+class Exit(Object):
+    def __init__(self, posX, posY):
+        super().__init__(posX, posY)
+
+class Robot(Object):
+    def __init__(self, posX, posY):
+        super().__init__(posX, posY)
+        self.active = False
+
+    def explodes(self, mines: list) -> bool:
+        '''Return True if the robot has stepped on a mine.'''
+        for mine in mines:
+            if self.x == mine.x and self.y == mine.y:
+                return True, mine.x, mine.y
+        return False, 0, 0
+
+    def exitFound(self, oExit: Exit) -> bool:
+        '''Return true if the robot found the exit.'''
+        if self.x == oExit.x and self.y == oExit.y:
+            return True
+        return False
+    
+    def updatePosition(self, posX, posY):
+        self.x = posX
+        self.y = posY
+    
+    def startEngine(self):
+        self.active = True
+
+    def stopEngine(self):
+        self.active = False
+
+class MineField:
+    def __init__(self):
+        self.mines = []
+        self.map = []
+        self.invalidMap = False
+        self.exitMark = False
+        self.robotMark = False
+
+    def setMine(self, posX, posY):
+        self.mines.append(Mine(posX, posY))
+
+    def setExit(self, posX, posY):
+        self.exit = Exit(posX, posY)
+        self.exitMark = True
+    
+    def setRobot(self, posX, posY):
+        self.robot = Robot(posX, posY)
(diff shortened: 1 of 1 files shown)


chall08/pcuadrad.py
```diff
@@ -0,0 +1,185 @@
...
+            self.invalidMap = True
+
+    def makeRobotMove(self, befX, befY, newX, newY):
+        '''Check if the robot could move to the next x-y position.\
+            Also checks in the new position if stepped a mine.'''
+        if self.map[newX][newY] != '+' and self.robot.active:
+            self.robot.updatePosition(newX, newY)
+            self.map[befX][befY] = '0'
+            self.map[newX][newY] = 'M'
+            die, xMine, yMine = self.robot.explodes(self.mines)
```
aguimbao, line 137: `xMine` and `yMine` are unused.

chall08/pcuadrad.py
```diff
@@ -0,0 +1,185 @@
+#!/usr/bin/env python3
```
aguimbao, line 1: This is not an error per-se, just a recommendation, I think I already told you this, but I don't remember well.

You use features like `f-strings` that are only available in >= python3.6.
In my system this gives me an error because it tries to execute it with the first `python3` it finds in my `PATH`, which is `python3.5` and crashes.
But it's not an error because there's no way you could actually know what exact version of the python interpreter I had installed in my system. Just a heads up for you in case you end up having this error later on.

But good job on permissions and shebang.

chall08/pcuadrad.py
```diff
@@ -0,0 +1,185 @@
...
+                        return False
+                else:
+                    return False
+        return True if limit == 2 else False
+
+    def constructMap(self, mapi):
+        '''Create the map in a two-dimensional array and check for errors.'''
+        self.invalidMap = False
+        for line, linePos in zip(mapi, range(len(mapi) - 1)):
+            if linePos > 0 and len(line) != len(lastLine):
```
aguimbao, line 120: I know this won't happen normally, but reading the variable `lastLine` here before you initially assign it in line 126, is not good practice.

chall08/pcuadrad.py
```diff
@@ -0,0 +1,185 @@
...
+            If the map is already invalid it does nothing.'''
+        instr = instr.rstrip('\n')
+        if not self.invalidMap:
+            for command in instr:
+                self.moveRobot(command)
+
+def main():
+    game = MineField()
+    mapi = []
+    for line in sys.stdin:
```
aguimbao, line 171: So. A little trick here, where you wouldn't have had to append every line and construct the board yourself, would have been to use `sys.stdin.read()`, which will hang until `EOF`, and then just `.split()` the board and the instructions between `"\n\n"`, and then split every line of the board with `.split("\n")`, which would generate a list with each row for you.

You will see in the posted solution I use this method.

aguimbao (commented):
Overall your approach is nice. I also approached the exercise using classes, but personally, I think the way you used them is a bit overkill, what I mean is. Do you really need the position of all the mines stored in memory constantly? Even if most of them will never be checked?
The robot class is really nice, you will see mine is similar.

Your logic is very good and works nicely, there was just something you misunderstood. In the subject you are told that: `A valid map must be a closed square with a single exit point.`

We consider that a closed square means:

```
++++++
+0000+
+0000+
+0000+
++++++
```

But for us, :

```
++++++
+0000+
+0000+
+0000+
+++++E
```

Is an invalid map, because the "square" board, is not perfectly closed. So it respects the rule of having one exit point, but it's not a valid map, which means you should return `./xlogin.py: Invalid input.` instead of `Failure.`

PR #231, opened 2020-05-02
vsempere

victorsempere:
Hi!
I tried to be more "pytonic" this time but ... I think I need more practice.

The way I solved this challenge:
1) Read the input from stdin
2) I try to recover from the input the map and the movements.
3) Later on the data received is checked. No more, no less than one robot and exit. Board must be a square. Movements valid, ...
4) Later on process movements:
4.1) If no engine started, .., nothing will be done except for start the engine
4.2) If engine is strated, ..., we check the movement can be done: this is, we cannot walk into a wall.
4.3) If movement is valid, we move the robot 
4.4) Check the content of the pos to determine status. Everything goes on if we don't hit a bomb
5) When all the movements have been processed, we check the content of the cell were the robot is. If it's a E and the engine is stopped -> Success. In other case Failure.

### DIFF
--- a/chall07/build.sh
+++ b/chall07/build.sh
@@ -1,0 +1,1 @@
+docker run --rm -v "$PWD":/usr/src/myapp -w /usr/src/myapp golang:1.13-alpine go build -v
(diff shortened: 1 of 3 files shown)


victorsempere:
I will close the pull request because I've found a bug in the function that checks if the move is valid

PR #232, opened 2020-05-02
edramire - Monte Olimpo

edramir18:
To resolve the problem i created two classes, the first for manipulate the minefield and the second to operate the robot.
The lines are readed one by one from the standard input and validated after read the command line for the robot. In case any character is not valid according to the specifications given by the problem, the program show a message indicating that input was invalid.
After the input is validated, the robot is operated according to the intructionts  throught the move fuction,  using movement vectors to calculate the final position, and checking after each move if that position hold a mine, in that case the processing of the instructions is stopped and a message is show to the user.
On finished the processing of the instructions line, in case the robot is stopped correctly and the final position are the exit, the program show a message indicating the success of the instructions given to the robot. Otherwise a failure message is show.
After processing a test case, the program wait for new tests case or the closing of the standar input.


chall08/edramire.py
```diff
@@ -0,0 +1,138 @@
...
+                    if robot.destroyed:
+                        print('Exploted.')
+                        break
+                if not robot.destroyed:
+                    if robot.on_exit(minefield):
+                        print('Success.')
+                    else:
+                        print('Failure.')
+            else:
+                print('Invalid input.')
```
clafoutis42, line 131: That's unfortunate, the error was supposed to be formatted `./xlogin.py: Invalid input.`

clafoutis42, line 131: That's unfortunate, the error was supposed to be formatted `./xlogin.py: Invalid input.`

chall08/edramire.py
```diff
@@ -0,0 +1,138 @@
...
+                else:
+                    minefield.valid = False
+                    break
+            if (minefield.valid and minefield.exit and minefield.robot
+                    and len(line) > 0 and all(c in 'NESWIK' for c in line)):
+                robot = Robot(*minefield.robot)
+                for c in line:
+                    robot.move(minefield, c)
+                    if robot.destroyed:
+                        print('Exploted.')
```
clafoutis42, line 123: You should check your code more thoroughly, even with a correct program, your typos make you fail :(

chall08/edramire.py
```diff
@@ -0,0 +1,138 @@
...
+        elif c == 'K':
+            self.on = False
+        elif self.on:
+            row = self.row + Robot.moves[c][0]
+            col = self.col + Robot.moves[c][1]
+            if mine_field.is_valid_position(row, col):
+                self.row = row
+                self.col = col
+                if mine_field.position_has_bomb(row, col):
+                    self.destroyed = True
```
clafoutis42, line 87: To make a more readable code you could have separated this function in multiple functions.
For example: a function that inputs the instruction, a function that handles the on/off, and a function that handles movement and.
It would look like this:
```python3
    def instruction(self, mine_field: Minefield, instruction: str):
        if instruction in "IK":
            return self.switch(instruction)
        self.move(minefield, *self.moves[instruction])

    def switch(self, instruction: str):
        # handle switch instruction...

    def move(self, minefield: Minefield, xmov: int, ymov: int):
        # handle movement...
```

chall08/edramire.py
```diff
@@ -0,0 +1,138 @@
...
+        'W': (0, -1)
+    }
+
+    def __init__(self, row, col):
+        self.row = row
+        self.col = col
+        self.on = False
+        self.destroyed = False
+
+    def move(self, mine_field: Minefield, c: str):
```
clafoutis42, line 75: I do not like the fact that you have to pass the minefield to the robot, as the robot is a part of the minefield, I think it would be more common sense to have the robot be a part of the minefield class, so you can do the position checks from the minefield class.

chall08/edramire.py
```diff
@@ -0,0 +1,138 @@
...
+        'E': (0, 1),
+        'S': (1, 0),
+        'W': (0, -1)
+    }
+
+    def __init__(self, row, col):
+        self.row = row
+        self.col = col
+        self.on = False
+        self.destroyed = False
```
clafoutis42, line 73: You could just raise a custom error on destruction and catch it later !

chall08/edramire.py
```diff
@@ -0,0 +1,138 @@
...
+    def is_valid_position(self, row: int, col: int):
+        if (row < 0 or row >= self.rows or col < 0 or col >= self.cols
+                or self.lines[row][col] == '+'):
+            return False
+        return True
+
+    def position_has_bomb(self, row, col):
+        if self.lines[row][col] == '*':
+            return True
+        return False
```
clafoutis42, line 50: So basically you wrote `if True return True else return False`?
```suggestion
    def position_has_bomb(self, row, col):
        return self.lines[row][col] == '*'
```

chall08/edramire.py
```diff
@@ -0,0 +1,138 @@
...
+                return False
+        else:
+            return False
+        return True
+
+    def is_valid_position(self, row: int, col: int):
+        if (row < 0 or row >= self.rows or col < 0 or col >= self.cols
+                or self.lines[row][col] == '+'):
+            return False
+        return True
```
clafoutis42, line 45: `If not True return False else return True` here

chall08/edramire.py
```diff
@@ -0,0 +1,138 @@
...
+                    if 'E' in row:
+                        minefield.exit = (minefield.rows, row.index('E'))
+                    if 'M' in row:
+                        minefield.robot = (minefield.rows, row.index('M'))
+                        line = line.replace('M', '0')
+                    minefield.rows += 1
+                    minefield.lines.append(row)
+                else:
+                    minefield.valid = False
+                    break
```
clafoutis42, line 116: I think this whole part should be in the Minefield class, you should pass it the lines through the init.
Also it seems like you forgot to check if the the map was a closed square :/

chall08/edramire.py
```diff
@@ -0,0 +1,138 @@
...
+            return False
+        if all(c in '+E' for c in line_str):
+            if e == 1 and (line_str.startswith('E') or line_str.endswith('E')):
+                return False
+            if (self.rows > 1
+                    and all(c in '+E' for c in self.lines[self.rows - 1])):
+                return False
+        elif all(c in 'M0*' for c in line_str[1:len_line - 1]):
+            if line_str[0] not in '+E' or line_str[-1] not in '+E':
+                return False
```
clafoutis42, line 36: Using regular expressions here would have helped you make the code more readable I think.  
First and last line regex: `^\+[\+E]+\+$`  
Middle line regex `^[\+E][0M\*]+[\+E]$`

And then you check if the count of M and E is correct.

clafoutis42 (commented):
Your solution logic is good, it works on most valid maps. However you made a typo in `Exploted`, the `Invalid input` error is badly formatted and you forgot to check if the square was closed by walls :(

Too bad, but nice try! Better luck next time.

chall08/edramire.py
edramir18: thanks for the feedback, i will try to be more careful next time.

PR #233, opened 2020-05-03
vsempere - 2

victorsempere:
The process I do to solve the problem is:
Process user input until the EOF is received. Then process the data to get the map and moves.
After this, I check the input to assure it's valid: the map is a square, there is only one robot, movements are valid movements,  ...
If everything is ok then I start to process every movement. And If no mine is hit at the end I evaluate if the end has been reached and the engine has been stopped.
Robot movements won't apply if a wall is hit or if the engine has not been started.


victorsempere:
Gracias!
Thank you. I have fixed it. These means I'm out of the challenge?

migferna42:
You have sneaked a .go file ;)

victorsempere:
Fixed the problem with the call07 go file.

clafoutis42:
> Fixed the problem with the call07 go file.

Little advice about that: when you have this kind of problem just create a new branch from the current head, it is less polluting (although the files are not in your PR anymore, the old commits are (meaning that your file is here in someway))

chall08/vsempere.py
```diff
@@ -0,0 +1,124 @@
...
+                        return ('Exploded.', 2)
+    if (map[rpos] == 'E' and engine_started == 0):
+        return ('Success.', 0)
+    return ('Failure.', 1)
+
+def main():
+    user_input = sys.stdin.readlines()
+
+    map, map_line_size, moves, error = process_user_input(user_input)
+    if not error is None:
```
clafoutis42, line 113: In python it is more common to handle errors by raising exceptions and catching them.  

Also:
```suggestion
    if error is not None:
```

chall08/vsempere.py
```diff
@@ -0,0 +1,124 @@
...
+        if (not move in "IKNSEW"):
+            return 0
+    return 1
+
+def get_robot_pos(map):
+    pos = 0
+    for cell in map:
+        if cell=='M':
+            return pos
+        pos = pos + 1
```
clafoutis42, line 70: You already have a builtin for that ;) 
```suggestion
    return map.index("M")
```

chall08/vsempere.py
```diff
@@ -0,0 +1,124 @@
...
+                return 0
+            if (map_cell == 'M'):
+                robots = robots + 1
+        col = col + 1
+        if (col == map_line_size):
+            col = 0
+            row = row + 1
+    
+    if (row != rows or robots != 1 or exits != 1):
+        return 0
```
clafoutis42, line 58: Using regular expressions here would have helped you make the code more readable I think.  
First and last line regex: `^\+[\+E]+\+$`  
Middle line regex `^[\+E][0M\*]+[\+E]$`
If you want to check all lines at the sam

And then you check if the count of M and E is correct.

chall08/vsempere.py
```diff
@@ -0,0 +1,124 @@
...
+    
+    for move in moves:
+        if (not move in "IKNSEW"):
+            return 0
+    return 1
```
clafoutis42, line 63: You are using python, you have booleans, why not use them?

chall08/vsempere.py
```diff
@@ -0,0 +1,124 @@
...
+                elif (map_line_size != len(line)):
+                    return None, None, None, 'Invalid input.'
+                map = map + line
+        else:
+            if (moves_read == 0):
+                moves = line
+                moves_read = 1
+            else:
+                return None, None, None, 'Invalid input.'
+    return map, map_line_size, moves, None
```
clafoutis42, line 30: It seems to me like you are checking the input 2 times, you could have done everything in the same function!

clafoutis42 (commented):
You seem to have a working solution, although it looks a lot like C code.  
There's just one problem, the `Invalid input.` error format is wrong :(

Too bad, better luck next time!

PR #234, opened 2020-05-03
Chall08-migferna

migferna42:
### Validation of the minefield

- We check that the first and last row of the minefield, is all of walls ('+')
- From the second row to the penultimate, they must start and end with a wall or have an exit
- We only accept one robot and one output
- Only accept a map with the allowed symbols

 Note: * If it meets the above, then we say that the map is valid

### Movements

- If the destination position of the movement is a wall, we stay in the same place
- If we find a bomb, we don't move, and we return the position of the bomb

### Minifield
We look for the first instruction to start the robot (if it has one), and from that instruction instruction is extracted, and executing it, if we find an instruction to turn off the robot, the rest of the instructions are ignored until we find a new instruction to restart the robot.

_Note to reviewer: I know that the code is not optimal or pythonical, but I have little knowledge of the language, and every little thing I try to solve I have to go to the manuals_


chall08/migferna.py
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env python3
+
+import sys
+
+def first_and_last(line):
+    for c in line:
+        if c != '+':
+            return False
+    return True
```
clafoutis42, line 9: There's 2 simple solution to do this ;) :
- `return line.replace("+", "") == ""`
- Use this regex `^\+[\+E]+\+$`

chall08/migferna.py
```diff
@@ -0,0 +1,135 @@
...
+def move(minifield, robotx, roboty, newx, newy):
+    minifield[robotx][roboty] = 0
+    minifield[newx][newy] = 'M'
+    return minifield
+
+def play(minifield, instructions, robotx, roboty, robotState):
+    instruction = instructions.pop(0)
+    while instruction != 'I' and len(instructions) > 1:
+        instruction = instructions.pop(0)
+    robotState = True if instruction == 'I' else 0
```
clafoutis42, line 77: Why ? 😅  
`var = True if True else 0`

chall08/migferna.py
```diff
@@ -0,0 +1,135 @@
...
+
+def move(minifield, robotx, roboty, newx, newy):
+    minifield[robotx][roboty] = 0
+    minifield[newx][newy] = 'M'
+    return minifield
+
+def play(minifield, instructions, robotx, roboty, robotState):
+    instruction = instructions.pop(0)
+    while instruction != 'I' and len(instructions) > 1:
+        instruction = instructions.pop(0)
```
clafoutis42, line 76: Why not using the same function all along?

chall08/migferna.py
```diff
@@ -0,0 +1,135 @@
...
+        for posy, c in enumerate(line):
+            if c == 'E' and found:
+                return -1, -1
+            elif c == 'E':
+                exitx = posx
+                exity = posy
+    return exitx, exity
+
+def move(minifield, robotx, roboty, newx, newy):
+    minifield[robotx][roboty] = 0
```
clafoutis42, line 69: Watch out here`"0"` and `0` are different, it's not javascript :)

chall08/migferna.py
```diff
@@ -0,0 +1,135 @@
...
+            newy = newy + 1
+        elif instruction == 'W' and robotState:
+            newx = newx
+            newy = newy - 1
+        elif instruction == 'K':
+            robotState = False
+        elif instruction == 'I':
+            robotState = True
+        
+        if minifield[newx][newy] == '0' or minifield[newx][newy] == 'E':
```
clafoutis42, line 103: When you go passed the exit, you crash :/

chall08/migferna.py
```diff
@@ -0,0 +1,135 @@
...
+            move(minifield, robotx, roboty, newx, newy)
+        elif minifield[newx][newy] == '+':
+            move(minifield, robotx, roboty, robotx, roboty)
```
clafoutis42, line 106: Why not just do nothing ?

chall08/migferna.py
```diff
@@ -0,0 +1,135 @@
...
+    for c in line:
+        if c != '+':
+            return False
+    return True
+
+def check_line(line):
+    if (line[0] != '+' and line[0] != 'E') or (line[-1] != '+' and line[-1] != 'E'):
+        return False
+    for c in line[1:-1]:
+        if c != '*' and c != '0' and c != '+' and c != 'M' and c != 'E':
```
clafoutis42, line 15:
```suggestion
        if c not in "*0+ME":
```

chall08/migferna.py
```diff
@@ -0,0 +1,135 @@
...
+    for l in minifield:
+        checked = check_line(l)
+        if not checked:
+            return False
+    last = first_and_last(minifield[-1])
+
+    if robotx == -1 or exitx == -1 or not first or not last:
+        return False
+    else:
+        return True
```
clafoutis42, line 32: Just... `if not True return False else return True` ? :/

chall08/migferna.py
```diff
@@ -0,0 +1,135 @@
...
+    return True
+
+def check(minifield):
+    robotx, roboty = valid_robot(minifield)
+    exitx, exity = valid_exit(minifield)
+    first = first_and_last(minifield[0])
+    for l in minifield:
+        checked = check_line(l)
+        if not checked:
+            return False
```
clafoutis42, line 26:
```suggestion
        if not check_line(l):
            return False
```

clafoutis42 (commented):
Your solution is raising `IndexError` multiple times because it lacks safety conditions.  
I'm sorry if I did not place a comment on every error I met, but I'm seeing things I've already corrected in the previous challenges.  

It seems to me you're always looking for a complicated solution even when there's a simple one, try simplifying your code!

migferna42:
Thank you very much, you do not know what I am learning with these exercises, I am going to read all the corrections, both mine and my classmates, to see if I can adapt to the Python syntax

PR #235, opened 2020-05-03
chall08 - gbudau

gbudau:
    Get input from the user in an infinite loop, first for the `minefield` and second for the `actions`, then validate the input and control the `robot` based on the `actions` provided then output "Exploded.", "Success." or "Failure." depending on the position and state of the `robot`.

**Short description of your code:**

Notes:
Created some global constant variables to make the code easier to read.
`__file__` , was returning the relative path when testing so i'm using it to print the program name.

Get input from the user in an infinite loop for the minefield and actions, using a try/except for `KeyboardIntrerrupt`.
If there is an `EOFError` or `KeyboardIntrerrupt`, exit the program.
Validate the input, check that the outer grid has walls and one exit and the inner grid has a `robot` and contains valid characters.

In `remote_control` function, get the `robot_position` then execute the `actions` until either the `robot` steps on a `mine` or there are no more `actions` to execute.

Output "Exploded.", "Success." or "Failure." depending on whether the robot stepped on a mine, reached the exit and has the engine stopped or neither of those.
  
To quit the script press `CTRL + C` or `CTRL + D`.

### DIFF
--- a/chall08/gbudau.py
+++ b/chall08/gbudau.py
@@ -1,0 +1,173 @@
+#!/usr/bin/env python3
+
+_VALID_ACTIONS = "IKNSWE"
+_INNER_GRID = "M0*"
+_OUTER_GRID = "+E"
+_WALL = "+"
+_MINE = "*"
+_ROBOT = "M"
+_EXIT = "E"
+_INIT = "I"
+_STOP = "K"
+
+def get_minefield():
+    minefield = []
+    err = False
+
+    try:
+        line = input()
+        while line:
+            minefield.append(line)
+            line = input()
+    except EOFError:
+        err = True
+        return minefield, err
+    return minefield, err
+
+def get_actions():
+    err = False
+    actions = ""
+
+    try:
+        actions = input()
+    except EOFError:
+        err = True
+        return actions, err
+    return actions, err
+
+def equal_width(minefield):
+    length = len(minefield[0])
+
+    for row in minefield:
+        if len(row) != length: 
+            break
+    return len(row) == length
+
+def exit_in_corner(minefield):
+    height_idx = len(minefield) - 1
+    width_idx = len(minefield[0]) - 1
+
+    if minefield[0][0] == _EXIT or \
+       minefield[0][width_idx] == _EXIT or \
+       minefield[height_idx][0] == _EXIT or \
+       minefield[height_idx][width_idx] == _EXIT:
+        return True
+    return False
+
+def one_exit(minefield):
+    exits = minefield[0].count(_EXIT)
+    exits += minefield[-1].count(_EXIT)
+    
+    for row in minefield[1:-1]:
+        if exits > 1:
+            break
+        if row[0] in _EXIT or row[-1] in _EXIT:
+            exits += 1
+    return exits == 1
+
+def check_outer_grid(minefield):
+    if exit_in_corner(minefield) or \
+       not all(c in _OUTER_GRID for c in minefield[0][1:-1]) or \
+       not all(c in _OUTER_GRID for c in minefield[-1][1:-1]) or \
+       not all(row[0] in _OUTER_GRID and row[-1] in _OUTER_GRID for row in minefield[1:-1]) or \
+       not one_exit(minefield):
+        return False
+    return True
+
+def check_inner_grid(minefield):
+    robots = 0
+
+    for row in minefield[1:-1]:
+        robots += row[1:-1].count(_ROBOT)
+        if not all(c in _INNER_GRID for c in row[1:-1]):
+            return False
+    return robots == 1
+
+def validate_minefield(minefield):
+    if not minefield or len(minefield) < 3 or \
+       len(minefield[0]) < 3 or not equal_width(minefield) or \
+       not check_outer_grid(minefield) or not check_inner_grid(minefield):
+        return False
+    return True
+
+def validate_actions(actions):
+    if not actions or not all(c in _VALID_ACTIONS for c in actions):
+        return False
+    return True
+
+def robot_position(minefield):
+    for x, row in enumerate(minefield):
+        for y, col in enumerate(row):
+            if col in _ROBOT:
+                return x, y
+    return -1, -1
+
+def move_robot_exit(minefield, action, row, col):
+    height_idx = len(minefield) - 1
+    width_idx = len(minefield[0]) - 1
+
+    if action in "N" and row > 0 and col in range(1, width_idx):
+        row -= 1
+    elif action in "S" and row < height_idx and col in range(1, width_idx):
+        row += 1
+    elif action in "W" and col > 0 and row in range(1, height_idx):
+        col -= 1
+    elif action in "E" and col < width_idx and  row in range(1, height_idx):
+        col += 1
+    return row, col
+
+def move_robot(minefield, action, row, col):
+    if action in "N" and minefield[row - 1][col] not in _WALL:
+        row -= 1
+    elif action in "S" and minefield[row + 1][col] not in _WALL:
+        row += 1
+    elif action in "W" and minefield[row][col - 1] not in _WALL:
+        col -= 1
+    elif action in "E" and minefield[row][col + 1] not in _WALL:
+        col += 1
+    return row, col
+
+def remote_control(minefield, actions):
+    on = False
+
+    row, col = robot_position(minefield)
+    if row == -1 or col == -1:
+        return
+    for action in actions:
+        if minefield[row][col] in _MINE:
+            break
+        if action in _INIT and on == False:
+            on = True
+        elif action in _STOP and on == True:
+            on = False
+        elif on == True and minefield[row][col] in _EXIT:
+            row, col = move_robot_exit(minefield, action, row, col)
+        elif on == True:
+            row, col = move_robot(minefield, action, row, col)
+
+    if minefield[row][col] in _MINE:
+        print("Exploded.")
+    elif minefield[row][col] in _EXIT and on == False:
+        print("Success.")
+    else:
+        print("Failure.")
+
+def main():
+    try:
+        while True:
+            minefield, err = get_minefield()
+            if err:
+                return
+            actions, err = get_actions()
+            if err:
+                return
+            if not validate_minefield(minefield) or not validate_actions(actions):
+                print("{}: Invalid input.".format(__file__))
+                continue
+            remote_control(minefield, actions)
+    except KeyboardInterrupt:
+        print()
+        return
+
+if __name__ == "__main__":
+    main()


chall08/gbudau.py
```diff
@@ -0,0 +1,173 @@
...
+                return
+            actions, err = get_actions()
+            if err:
+                return
+            if not validate_minefield(minefield) or not validate_actions(actions):
+                print("{}: Invalid input.".format(__file__))
+                continue
+            remote_control(minefield, actions)
+    except KeyboardInterrupt:
+        print()
```
aguimbao, line 169: Why would you print a newline here ?

chall08/gbudau.py
```diff
@@ -0,0 +1,173 @@
...
+            if err:
+                return
+            actions, err = get_actions()
+            if err:
+                return
+            if not validate_minefield(minefield) or not validate_actions(actions):
+                print("{}: Invalid input.".format(__file__))
+                continue
+            remote_control(minefield, actions)
+    except KeyboardInterrupt:
```
aguimbao, line 168: Ok, so even though catching `KeyboardInterrupt` is good practice. We weren't asking for it.
I usually don't like using `try`, `except` and `finally` statements that often, and it's obvious that we `Ctrl-C`'ing your program is not part of our tests, it wouldn't be fair because we didn't ask you to protect your program against it.

chall08/gbudau.py
```diff
@@ -0,0 +1,173 @@
...
+    try:
+        while True:
+            minefield, err = get_minefield()
+            if err:
+                return
+            actions, err = get_actions()
+            if err:
+                return
+            if not validate_minefield(minefield) or not validate_actions(actions):
+                print("{}: Invalid input.".format(__file__))
```
aguimbao, line 165: Ok, so good practice here, because to get the filename of the module, `__file__` is much safer than `sys.argv[0]`, in case the module being executed is not the main module, only `__file__` will work.

For the formatting, this is just a suggestion, but `f-strings` are avalaible since `python3.6`, and I think personally they are more readable for simple and short print statements, and the string's `.format()` method is more readable and more useful for longer and more complex print statements.

So `print("{}: Invalid input.".format(__file__))` -> `print(f"{__file__}: Invalid input.")`

chall08/gbudau.py
```diff
@@ -0,0 +1,173 @@
+#!/usr/bin/env python3
+
+_VALID_ACTIONS = "IKNSWE"
```
aguimbao, line 3: So I know it's not the case for all of this global variables, but as you only use some of them 1 time in their respective functions, like `_VALID_ACTIONS` or `_INNER_GRID`, maybe it would have been better to declare them inside the function scope, it would maybe help readability, and also local variables are more performant than global variables.

chall08/gbudau.py
```diff
@@ -0,0 +1,173 @@
...
+_INNER_GRID = "M0*"
+_OUTER_GRID = "+E"
+_WALL = "+"
+_MINE = "*"
+_ROBOT = "M"
+_EXIT = "E"
+_INIT = "I"
+_STOP = "K"
+
+def get_minefield():
```
aguimbao, line 13: Instead of all this `try` and `except` clauses and that ugly `while True` in the `main()` function to try and catch that juicy `EOF` and a lot of functions to read the lines, you could have used `sys.stdin.read()`, this will read everything until the `EOF`, and return it to you as a string. You will be able to see it in the `solution.py` that I will submit later.

chall08/gbudau.py
```diff
@@ -0,0 +1,173 @@
...
+def remote_control(minefield, actions):
+    on = False
+
+    row, col = robot_position(minefield)
+    if row == -1 or col == -1:
+        return
+    for action in actions:
+        if minefield[row][col] in _MINE:
+            break
+        if action in _INIT and on == False:
```
aguimbao, line 139: Comparisons to `False` or `True` should be `x is False` or `x is True`, not `==`.

This is the same for lines 141, 143, 145 and 150.

chall08/gbudau.py
```diff
@@ -0,0 +1,173 @@
...
+
+def validate_minefield(minefield):
+    if not minefield or len(minefield) < 3 or \
+       len(minefield[0]) < 3 or not equal_width(minefield) or \
+       not check_outer_grid(minefield) or not check_inner_grid(minefield):
+        return False
+    return True
+
+def validate_actions(actions):
+    if not actions or not all(c in _VALID_ACTIONS for c in actions):
```
aguimbao, line 94: Little trick for this line and the similar ones: 

```python
if set(actions) - set(_VALID_ACTIONS):  # Bitwise operation
    # if True this means there's an invalid char
```

chall08/gbudau.py
```diff
@@ -0,0 +1,173 @@
...
+        actions = input()
+    except EOFError:
+        err = True
+        return actions, err
+    return actions, err
+
+def equal_width(minefield):
+    length = len(minefield[0])
+
+    for row in minefield:
```
aguimbao, line 41: This could be: `return all([len(row) == len(minefield[0]) for row in minefield])` in 1 line.

chall08/gbudau.py
```diff
@@ -0,0 +1,173 @@
...
+    return False
+
+def one_exit(minefield):
+    exits = minefield[0].count(_EXIT)
+    exits += minefield[-1].count(_EXIT)
+    
+    for row in minefield[1:-1]:
+        if exits > 1:
+            break
+        if row[0] in _EXIT or row[-1] in _EXIT:
```
aguimbao, line 64: There's a problem with your check here, if the board has two exits in both sides, and it's not in the last or first row, you only count one exit :/

Which means you take:

```
+++++++++++
+*0*0*0*0*+
+0*0*0*0*0+
+000000000+
+0000M0000+
E000000000E
+++++++++++
```

as a valid map !

chall08/gbudau.py
aguimbao, line 165: You should exit on `Invalid input.`. You print `Invalid input.` two times for this test case:

```
+++++++++++
+*0*0*0*0*+
+0*0*0*0*0+
+000000000+
+0000M0000+
+000000000+
++++++++++E
```

aguimbao (commented):
Hey ! Your code is good overall and your approach is nice, you only make only the necessary checks. But I think this challenge was asking for at least a `Robot` Class, it would have been a nice opportunity to try out OOP or just learn how Classes work in Python if you are interested in it.

You should take a look at the `solution.py` that will be provided a bit later to check it out. Good luck next time ;)

chall08/gbudau.py
gbudau: I think i misinterpreted how the script should work.
I interpreted that the input should come from stdin, as input prompt, for example the way `bc` gets input and that it should wait until the user close it.
The newline here is to put a newline after displaying `^C` (Ctrl-C), like when using `cat` without arguments and Ctrl-C.

chall08/gbudau.py
gbudau: I misinterpreted this line

> Your program shall wait for the user to close the stdin by his

I will try to clarify any questions i have before solving in the next challenges, thank you!

aguimbao:
I'm sorry I had duplicated some of the comments 😓

chall08/gbudau.py
aguimbao: It wasn't that clear :/ sorry. But yeah you should definitely ask it in the thread next time !

chall08/gbudau.py
gbudau: I will keep it in mind for the next time.
The workflow i used to solve the python challenges until now was something like: Search on python docs/learn x in y/search engine about how to do a specific thing in python and in this case input() was one of the first match on how to get input from stdin.
I guess better practice will come with more experience :smile: 

Also i checked the solution, learned a lot of new things from it :+1:

chall08/gbudau.py
gbudau: Thank you, will keep it in mind for next time.

gbudau (commented):
Thank you a lot for the feedback Abraham, very detailed. I'm learning so much from it and also from doing the challenges :heart:

gbudau:
I also duplicated some of the comments, at the moment github works in mysterious ways to me :smile: 

Edit: looks like i actually didn't duplicated them but since i thought i did, i removed the "duplicates" ... :facepalm:
Edit 2: All good writed the comments again :smile:

chall08/gbudau.py
gbudau: I prefered to use `.format` for the compatibility with all `python3` versions.
Will keep in mind `f-strings` for the next time.

chall08/gbudau.py
gbudau: I checked the `solution.py` and learned many things, thanks.

There are many ways to do something in python, the workflow i used was to search on docs/learn x in y/web and input() was one of the first matches for how to get input from stdin.

Will keep it in mind `sys.stdin.read()' for the next time :+1: .

chall08/gbudau.py
gbudau: Good point, can also be writed as `True/False` or `not True/not False` or in this case `on` / `not on`

chall08/gbudau.py
gbudau: Nice one! :+1: 
First need to wrap my head around how it works and will definately use it in the future.

chall08/gbudau.py
gbudau: Good point, one adition is to drop the list parenthesis `[]` so it will short-circuit the first time the condition is not meet instead of evaluating all rows.
`return all(len(row) == len(minefield[0]) for row in minefield)`

chall08/gbudau.py
gbudau: Thank you, well spotted.

chall08/gbudau.py
gbudau: If this is the test_case7.txt:

```
+++++++++++
+*0*0*0*0*+
+0*0*0*0*0+
+000000000+
+0000M0000+
+000000000+
++++++++++E

ISEEEEEK

asdadasdas
s% 
```

My script is looking for map+newline+instructions then repeating the process until the user press Ctrl-C or Ctrl-D.

For example:

```
+++
+ME
+++

IEK
+++
+ME
+++

IW
```
Will output first 'Success.' then 'Failure.'.

So it works the way i interpreted  the challenge and i misinterpreted the challenge :smile:

chall08/gbudau.py
aguimbao: Wow. Nice one 👍 Didn't think about it.

aguimbao:
We'll try to be more clear for the next challenge ! Good job and see you on the next one ❤️

PR #236, opened 2020-05-03
Ready to Try

szavarce42:
My program first reads from the stdin until the user closes it (ctrl+d).

First it checks for the lines that matches the valid box parameters and put them in a list. The same for the instructions, but if more than one line with instructions is given, it will throw an invalid input error.

We now call the check_box function that will check if:
- There is only one valid exit on top, botton, left or right walls
- There is only one robot
- The box is a closed square with only "+" signs boundaries
- There is only one line break after the box

Now we init a class  (Position) that will get the initial robot position.

We iterate over the instructions and call the method "next_pos" of the class to move the robot. It will only move if the state is "started". 

If the move to a mine it will explode, if we stop over the exit it will succeed. If it gets out without stopping it will fail.

### DIFF
--- a/chall08/szavarce.py
+++ b/chall08/szavarce.py
@@ -28,7 +28,9 @@ def check_box(box):
     count_upper = 0
     for lines in range(len(box)):
         count_m += box[lines].count("m")
-        count_e += box[lines].count("e")        
+        count_e += box[lines].count("e")
+        if len(box[lines]) != len(box[0]) and box[lines] != '\n':
+            return 0
         if re.match(r"^\+\+*\+$", box[lines]) or re.match(r"^\+[\+e]*\+$", box[lines]):
             count_upper += 1
         elif box[lines] == '\n':
@@ -53,8 +55,6 @@ for line in sys.stdin:
     else:
         instructions.append("-1")
 if not (instructions.count("-1") > 0 or len(instructions) != 1 or check_box(box) != 1):
-    #print ("Invalid Input.")
-    #exit()
     pos = Position(box)
     for move in instructions[0]:
         if move == "i":
--- a/chall08/szavarce.py
+++ b/chall08/szavarce.py
@@ -1,0 +1,75 @@
+#!/usr/bin/env python
+import sys
+import re
+
+class Position:
+    def __init__(self, box):
+        for lines in box:
+            if "m" in lines:
+                self.x = lines.index("m")
+                self.y = box.index(lines)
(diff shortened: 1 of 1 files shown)


chall08/szavarce.py
```diff
@@ -0,0 +1,75 @@
...
+        if move == "i":
+            started = True
+        elif move == "k":
+            started = False
+        else:
+            pos.next_pos(move, box, started)
+        result = 0
+        if pos.get_square(box, pos.x, pos.y) == "*":
+            print("Exploded.")
+            exit()
```
aguimbao, line 69: `sys.exit()` > `exit()`, btw you can print inside the exit statements, like `sys.exit("Exploded.")`

chall08/szavarce.py
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env python
+import sys
+import re
+
+class Position:
```
aguimbao, line 5: This is personal opinion, but I think the class should represent an Object, than has a position himself, like representing the robot, or a mine, etc. But good job on using classes.

chall08/szavarce.py
```diff
@@ -0,0 +1,75 @@
...
+            count_upper += 1
+        elif box[lines] == '\n':
+            if lines != len(box) - 1:
+                return 0
+        elif not re.match(r"^[\+e][0*m]*[\+e]$", box[lines]):
+            return 0
+    return 1 if count_e == 1 and count_m == 1 and count_upper == 2 else 0
+        
+
+box = []
```
aguimbao, line 44: Instead of having all this functionality in the global scope, I think it's better to always have a `main()` function, with it's corresponding ```
if __name__ == "__main__":
    main()
```

Just in case your module is ever imported in another module. (I know this will never be a case with this challenge, but it's good practice nonetheless.

File: chall08/szavarce.py [hunk 4]
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env python
```
aguimbao, line 1: This could have crashed if I had a `python2` interpreter in my `PATH` variable before my `python3` if you ever used only-python3 features like `f-strings` and such. I think it would be better to specify `#!/usr/bin/env python3`, just because we asked for `Python3.x`

aguimbao (commented):
There's not much to say about your code, it's understandable, it's short and easy to read, and it works.

Good use of regexes and classes to solve this little exercise. 👍

PR #239, opened 2020-05-09
Chall09 migferna

migferna42:
We check that seconds are passed to the function, and that they are valid parameters, otherwise we show "invalid input".

### Validation
We perform a split separating the words by spaces, and then each word is evaluated separately to see if all the characters are of the numeric type.
If an invalid character is found or there are multiple numeric words, the function is invalid.

### Format
We convert the input string to unsigned long long so that it can reach the maximum indicated in the statement. The time units have been unified in a struct, to make it easier to use.

### Output
Finally, the pieces of the final chain are co-forming, depending on whether they exist and whether they are singular or plural.

***NOTE**: I assume that the use of C library functions is allowed, since in previous challenges it was like this


chall09/migferna.c
```diff
@@ -0,0 +1,175 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <ctype.h>
+
+typedef	struct			s_time
+{
+	unsigned long long	years;
```
aguimbao, line 8: I think you needed at least a `long long` to store years, and `unsigned long long` for the seconds for your approach, but it would be more efficient to store the rest in regular `int`'s as they would fit no problem.

chall09/migferna.c
```diff
@@ -0,0 +1,175 @@
...
+	minutes_count = time.minutes > 1 ? "minutes" : "minute";
+	seconds_count = time.seconds > 1 ? "seconds" : "second";
+	hours_count = time.hours > 1 ? "hours" : "hour";
+	years_count = time.years > 1 ? "years" : "year";
+	days_count = time.days > 1 ? "days" : "day";
+
+	if (time.years)
+	{
+		sprintf(tmp, "%llu %s", time.years, years_count);
+		strcat(format, tmp);
```
aguimbao, line 42: `strncat()` is not safe, you should use `strncat()` and specify the sizes. You have a heap-buffer-overflow here for `long long` size values. Also for `unsigned long long` values like the maximum value you were asked to handle: `10000000000000000000`, or one less, `9999999999999999999`, for example.

We perform our checks with `fsanitize`, and so should you.

chall09/migferna.c
```diff
@@ -0,0 +1,175 @@
...
+char *ft_format_duration(char *seconds)
+{
+	char *out;
+	while (isspace(*seconds))
+		seconds++;
+
+	if (*seconds && is_valid(seconds))
+		out = parser_format(seconds);
+	else
+		out = "Invalid input.";
```
aguimbao, line 173: It would have been better to `out = strdup("Invalid input.")`. Even if it uses more memory and resources. You are returning a mix of dynamically allocated memory, and strings literals in other cases.
This is not good practice, as you normally would want to perform the same actions in the data you return. We obviously `free()` your returned variable to check for other leaks, which crashes sometimes for you (when you return a string literal like in this line).

chall09/migferna.c
```diff
@@ -0,0 +1,175 @@
...
+	if (time.seconds > 0)
+		count_units++;
+	
+	out = generate_output(time, count_units);
+	return (out);
+}
+
+int	is_valid(char *seconds)
+{
+	char *ptr = strtok(strdup(seconds), " ");
```
aguimbao, line 146: I think all the usages of `strtok()` are a bit useless here, you could have just advanced in the string instead of dividing it.

chall09/migferna.c
```diff
@@ -0,0 +1,175 @@
...
+	char *format = NULL;
+	char tmp[100];
+	char *minutes_count;
+	char *seconds_count;
+	char *hours_count;
+	char *years_count;
+	char *months_count;
+	char *days_count;
+
+	if (!(format = malloc(500)))
```
aguimbao, line 31: You could have done a bit better here, knowing that the longest string possible would have been 70 characters in length :/ But it would have been better to return a precisely allocated variable. Specially when you could concatenate everything...

aguimbao (commented):
Your approach is nice but you need to be careful when not dynamically allocating the variable you are returning, specially if you don't know what we are gonna do with it.

PR #240, opened 2020-05-09
chall09 - pcuadrad

pabloocg:
I first check the string for errors and make sure it is well formed in the function **check_error**.
In turn, this function returns the number.
In order to calculate large quantities like "10000000000000000000", I return the number in unsigned long long, if in the function something fails I return 0, and if I don't, return the number + 1 (to be able to make the case of the second '0').
I have created two structs one name **t_time** that will contain the number of separators ('comma' and 'and'), and also an array of five **t_unity** that each one of them is going to refer to the year, day, time ... etc.
In the function **calculate_time**, I calculate the seconds corresponding to the time unit of each instance t_unity, and if the final value is one I do a substring to remove the 's' from the end.
Once all are calculated, I return the result in make_result, here I loop through the five t_unities and concatenate the time units if they have been selected together with the separators.


chall09/pcuadrad.c
```diff
@@ -0,0 +1,215 @@
...
+    return (tmp);
+}
+
+char    *ft_format_duration(char *seconds)
+{
+    t_time              *Telescopium;
+    unsigned long long  second;
+
+    if ((second = check_error(seconds)) == 0)
+        return ("Invalid input.");
```
clafoutis42, line 199: If your function allocates memory, you should never return un-allocated adresses, it is really confusing for the user of your function.  
In this chall this won't be penalized, but don't take this habit !

chall09/pcuadrad.c
```diff
@@ -0,0 +1,215 @@
...
+long long   check_error(char *string)
+{
+    int     i;
+    char    *parse;
+    int     neg;
+    unsigned long long num;
+
+    neg = 1;
+    num = 0;
+    parse = ft_strtrim(string, " ");
```
clafoutis42, line 97: Why would you need to reallocate memory to do this? You can just ignore white-spaces at the beginning and the end!

chall09/pcuadrad.c
```diff
@@ -0,0 +1,215 @@
...
+    if (!parse || !*parse || !strlen(parse))
+        return (0);
+    i = 0;
+    while ((parse[i] == ' ' || parse[i] == '-' || parse[i] == '+') && parse[i])
+    {
+        if (parse[i] == '-')
+            neg *= -1;
+		i++;
+    }
```
clafoutis42, line 106: There was a change in the subject, so normally you would have needed to return `Invalid Input.` here, but given this change was made late, it won't be tested.

chall09/pcuadrad.c
```diff
@@ -0,0 +1,215 @@
...
+    Telescopium->sep += 1;
+    unity->selected = 1;
+    unity->value = seconds / unity->time_value;
+    seconds = seconds % unity->time_value;
+    if (unity->value == 1)
+    {
+        tmp = ft_substr(unity->time_unity, 0, strlen(unity->time_unity) - 1);
+        free(unity->time_unity);
+        unity->time_unity = tmp;
+    }
```
clafoutis42, line 162: It seems a little dirty reallocating memory just to remove one char. You could just replace the `s` by a `\0`.

clafoutis42 (commented):
That's an interesting algorithm. Although it makes some unnecessary memory allocations.  
You could have optimized it a little more !
It won't pass the test because you did not respect the limit of `10000000000000000000` :( 

One tiny suggestion I would have made, is building the string whilst parsing the numbers, to avoid allocating too much memory in structures ;)

chall09/pcuadrad.c
pabloocg: Oops ... I forgot to do it in this part, in the return 'now' I do it.

chall09/pcuadrad.c
pabloocg: Yes ... it is much better in matters of mermory, I write it down.

PR #241, opened 2020-05-09
vde-dios.c

vdedios:
### SOLUTION

As always, first of all input str is checked as parsed. Parsing is similar to atoi function (in this case atoull to manage all possible sizes) considering as valid input white spaces followed by a number and some extra white spaces. Since the max input number possible is 1e19, ir any error is encountered ULLONG_MAX is returned.

After parsing the input and converting it to ULL, is a matter of comparing the number against the different units required (years, days etc.) in seconds. If the number is greater it should be divided to extract its significative cifres. All the conversions are consequently concatenated using the `asprintf` function, which allocates memory on the go, and using a `tmp` pointer which stores previous string . Meanwhile everything is put together to form a proper sentence.

The function exits with no memory leaks except for the returned string.


chall09/vde-dios.c
```diff
@@ -0,0 +1,55 @@
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <limits.h>
+#include <ctype.h>
+
+unsigned long long     ft_ultimate_atoull(char *str)
```
aguimbao, line 7: It's good practice to use `static` keyword.

chall09/vde-dios.c
```diff
@@ -0,0 +1,55 @@
...
+{
+    unsigned long long num;
+    char               *aux;
+
+    while (isspace(*str))
+        str++;
+    if (!isdigit(*str))
+        return (ULLONG_MAX);
+    num = strtoull(str, &aux, 10);
+    while (*aux++)
```
aguimbao, line 17: Not the best way to check for good `strtoull` return in my opinion. It's better to use `errno` like explained here: https://stackoverflow.com/questions/40753285/strtoull-use-properly-in-c . You will see how we use it in our solution.

chall09/vde-dios.c
```diff
@@ -0,0 +1,55 @@
...
+        if (!isspace(*(aux - 1)))
+            return (ULLONG_MAX);
+    return (num);
```
aguimbao, line 20: You forgot something in your check, you forgot to check if the number is bigger than `10000000000000000000` which is still a `unsigned long long`, but you only needed to handle up to it. But this wasn't very clear in the subject so it probably won't be checked in the tests :), we will see.

chall09/vde-dios.c
```diff
@@ -0,0 +1,55 @@
...
+    unsigned long long  num = ft_ultimate_atoull(seconds);
+    unsigned long long  div;
+    long int            time[] = {31536000, 86400, 3600, 60, 1};
+    char                *units[] = {"year", "day", "hour", "minute", "second"};
+    char                *conversion;
+    char                *tmp = NULL;
+    int                 i = 0;
+
+    if (num == ULLONG_MAX)
+        return ("Invalid input.");
```
aguimbao, line 34: Please don't return a mix of `string literals`, and allocated memory. We usually free your string, so this will give all kinds of problems in case you return `"Invalid input."` or `"now"`.

chall09/vde-dios.c
```diff
@@ -0,0 +1,55 @@
...
+    if (!num)
+        return ("now");
+    while(i < 5)
+    {
+        if (num > time[i])
```
aguimbao, line 39: There's a fail in this check. You don't test for equality, so if `seconds = "60"`. You will return `60 seconds`, instead of `1 minute`. This happens for all the other time units.

chall09/vde-dios.c
```diff
@@ -0,0 +1,55 @@
...
+        {
+            div = num / time[i];
+            num = num % time[i];
+            if (!(num))
+                asprintf(&conversion, "%s%llu %s%s", !tmp ? "" : tmp, div, units[i], div > 1 ? "s" : "");
+            else if (!(num % time[i + 1]))
```
aguimbao, line 45: There's an edge case where this `if` condition never gets validated and your string never has a `" and "`.

For example for: `seconds = "8069197798646305224"` .

aguimbao (commented):
I really liked this approach, I solved the problem in a similar manner, but you really need to watch out for those pesky edge-cases ;)

I used `#define`'s for the time units instead of that smart `long int time[] = {31536000, 86400, 3600, 60, 1};` that you can iterate later, good job for that 👍 (I don't think you needed a `long int` to store those numbers though).

chall09/vde-dios.c
vdedios: Yes! I saw a few days ago that is also mandatory in Norminette to do it that way. I will do a little bit of research about this

chall09/vde-dios.c
vdedios: I will take a look at that, nice point

chall09/vde-dios.c
vdedios: From what it says on the subject, as you confirm is: 
> The biggest number of seconds that you need to handle is 10000000000000000000. Good luck :)

What I understood from that is numbers above 1e19 will never be passed... Maybe that smile after _good luck_ has double meaning? 😅🤣

chall09/vde-dios.c
vdedios: Yes, entirely my fault, I shouldn't mixed up memory types

chall09/vde-dios.c
vdedios: 😬

PR #242, opened 2020-05-09
chall09 | estina

xplore93:
My version of the solution.
First of all, the idea is to check first if the string is valid. Then transform it into a number. If it's 0, return "now", else transform seconds into years, days, hours, minutes and seconds. It solves dividing seconds for each type of time. Then join each time to one string that will be returned at the end. Also I use a variable to know how much types of time I have (not 0), so that I know when to put a coma or "and" into my string.


aguimbao:
Hey. I'm gonna start taking a look into this PR submission. Just letting you know I won't be able to merge until you fix the conflicts with the README.md file, we are just interested in the `<login>.c` file :)

chall09/estina.c
```diff
@@ -0,0 +1,96 @@
+/* ************************************************************************** */
```
aguimbao, line 1: It's a nice touch ❤️ but you don't need the header for the challenges.

xplore93:
Sure, I will try to solve it ASAP

chall09/estina.c
```diff
@@ -0,0 +1,96 @@
...
+	char				*new;
+	char				number_times;
+
+	years = transform_time(&seconds, 31536000);
+	days = transform_time(&seconds, 86400);
+	hours = transform_time(&seconds, 3600);
+	mins = transform_time(&seconds, 60);
+	number_times = (years ? 1 : 0) + (days ? 1 : 0) + (hours ? 1 : 0) +
+					(mins ? 1 : 0) + (seconds ? 1 : 0);
+	if (!(new = calloc(65, sizeof(char))))
```
aguimbao, line 75: Little fail here, for the biggest possible kind of string (for example: `151981058532 years, 332 days, 22 hours, 31 minutes and 39 seconds`), where every time unit is in plural, the total size of the string is 65 characters. So you still need one more allocated byte for the `\0` terminator. The `strcat()`in line `53` makes a heap-buffer-overflow, (better to use `strncat()`).

I recommend you check out gcc's `fsanitize` option for this kind of thing. You would have noticed it as this happens with the example case `seconds = 10000000000000000000`.

chall09/estina.c
```diff
@@ -0,0 +1,96 @@
...
+}
+
+char	*ft_format_duration(char *seconds)
+{
+	unsigned long long	number;
+
+	while (isspace(*seconds))
+		seconds++;
+	if (!check_input(seconds))
+		return ("Invalid input.");
```
aguimbao, line 92: I don't think it's good practice to return a mix of `string literals` and dynamically allocated memory, we normally `free()` your returned answer, so this obviously crashes our tester. Just returning the answer with `asprintf()` or `strdup()` in the case of `"Invalid input."` or `"now"` would have been better !

chall09/estina.c
```diff
@@ -0,0 +1,96 @@
...
+		return (NULL);
+	append_time(years, " year", " years", &number_times, &new);
+	append_time(days, " day", " days", &number_times, &new);
+	append_time(hours, " hour", " hours", &number_times, &new);
+	append_time(mins, " minute", " minutes", &number_times, &new);
+	append_time(seconds, " second", " seconds", &number_times, &new);
+	return (new);
+}
+
+char	*ft_format_duration(char *seconds)
```
aguimbao, line 85: You don't check if the value is over `10000000000000000000` which was the max number we asked you to handle. You also don't check for really big numbers where `strtoull()` will fail (bigger than `unsigned long long`, like `"4214032131247127411461641468712641471241241826481264821421"`). You should use `errno`, it's really useful. Take a look at https://stackoverflow.com/questions/40753285/strtoull-use-properly-in-c

chall09/estina.c
```diff
@@ -0,0 +1,96 @@
...
+	}
+}
+
+char	*human_time(unsigned long long seconds)
+{
+	unsigned long long	years;
+	short				days;
+	char				hours, mins;
+	char				*new;
+	char				number_times;
```
aguimbao, line 67: Nice memory saving using `char` instead of `int` :)

chall09/estina.c
```diff
@@ -0,0 +1,96 @@
...
+/*   Updated: 2020/05/10 15:34:40 by estina           ###   ########.fr       */
+/*                                                                            */
+/* ************************************************************************** */
+
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+#include <ctype.h>
+
+int		check_input(char *seconds)
```
aguimbao, line 18: For this and the rest of the functions except the asked function, you could have used the `static` keyword.

chall09/estina.c
```diff
@@ -0,0 +1,96 @@
...
+
+char	*human_time(unsigned long long seconds)
+{
+	unsigned long long	years;
+	short				days;
+	char				hours, mins;
+	char				*new;
+	char				number_times;
+
+	years = transform_time(&seconds, 31536000);
```
aguimbao, line 69: This is "subjective" but, `#define`'ing the time units, for example as `#define YEAR 31536000` at the top of the file, would make it easier to read for someone else that it taking a look at your code (obviously here it's obvious you are talking about the amount of seconds in a year, but I'm talking about a more complex program).

chall09/estina.c
```diff
@@ -0,0 +1,96 @@
...
+
+void	append_time(unsigned long long time, char *one, char *more, char *n, char **str)
+{
+	char	buf[13];
+
+	if (time)
+	{
+		sprintf(buf, "%llu", time);
+		*str = strcat(*str, buf);
+		*str = strcat(*str, time == 1 ? one : more);
```
aguimbao, line 53: You could have used `strncat()` or `asprintf()` and you probably wouldn't have had buffer problems (really powerful function).

chall09/estina.c
```diff
@@ -0,0 +1,96 @@
...
+		if (!isdigit(seconds[i]))
+			return (0);
+		i--;
+	}
+	return (1);
+}
+
+unsigned long long	transform_time(void *y, int z)
+{
+	unsigned long long	x;
```
aguimbao, line 38: I think `x`'s biggest needed size would be `long long` 😛

chall09/estina.c
```diff
@@ -0,0 +1,96 @@
...
+
+	x = *((unsigned long long *)y) / z;
```
aguimbao, line 40: I would personally avoid unnecessary casting and pass `y` directly as a pointer to `unsigned long long`, unless it's 100% necessary in this case and I don't see it.

aguimbao (commented):
Nice straightforward approach, there's not much performance to be gained here, maybe using `asprintf()` or similar to return the string in one go instead of concatenating everything, but I think you're fine 👍 

Too bad for the heap-buffer-overflow cause of the `calloc(65)` instead of `calloc(66)`.

chall09/estina.c
xplore93: I can't develop without it ;)

chall09/estina.c
xplore93: Yeah, I was thinking about it, but trying with that value, didn't have any problems and it remained as 65. Should check the fsanitize.

chall09/estina.c
xplore93: Shure, it's totally my fault.

chall09/estina.c
xplore93: Sometimes you tell me the function should fail if the condition is not maintained. Now you are saying I should check that limits. ;) So what should I do?

chall09/estina.c
xplore93: Thanks =)

chall09/estina.c
xplore93: I forgot it! 🤦‍♂️

chall09/estina.c
xplore93: 🐱‍🏍 Normally I do, but that time I didn't 🤷‍♀️🤷‍♀️

chall09/estina.c
xplore93: Should check that functions, thanks!

chall09/estina.c
xplore93: Ooops... My fault, sorry. You are goddamn right!

chall09/estina.c
xplore93: I will never do it. I promise!

chall09/estina.c
aguimbao: We weren't really clear about the condition for the maximum value. We will try to be more clear next time.

PR #243, opened 2020-05-09
edramire - Telescopium

edramir18:
According to the last information on the validation of the input, all cases on which exists characters others that the space and the digits from 0 to 9, are considered invalid input, and an error message is show to the user.
After validated the input, this is converted to a number, in case the number exceed the maximum allowed, an error message is showed.
Later, the number is splitted on years, days, hours, minutes and seconds, stored on a strutucture for easy manipulation. I use a variable as a collection of binary flags to determinate more easily if exists previous or next segments of values to print on the final string.
The final string is build using a function which has como paremeters, the value for each segment of time and the right way as has to be formatted, taken an account if exist a previous or next segment of time.


clafoutis42:
There are conflict with the README, please solve them before I can merge your branch!
I'll be reviewing your PR meanwhile.

chall09/edramire.c
```diff
@@ -0,0 +1,157 @@
...
+	return (time);
+}
+
+char 	*ft_format_duration(char *seconds)
+{
+	t_time	my_time;
+	char	*str;
+
+	if (ft_stc_validate_duration(seconds) > 0)
+		return strdup("Invalid input.");
```
clafoutis42, line 135: Maybe parsing the number of seconds into an int here would have helped you define if the input is valid or not in one condition only?

chall09/edramire.c
```diff
@@ -0,0 +1,157 @@
...
+	bzero(&my_time, sizeof(t_time));
+	my_time.years = ft_stc_convert_number(seconds);
+	if (my_time.years == ULLONG_MAX)
+		return strdup("Invalid input.");
+	my_time = ft_st_calculate_seconds(my_time);
+	my_time = ft_stc_set_flags(my_time);
+	if (my_time.flags == 0)
+		return (strdup("now"));
```
clafoutis42, line 143: I don't really understand, you could have guessed this earlier, no?  
Thus avoiding this much function calls before!

chall09/edramire.c
```diff
@@ -0,0 +1,157 @@
...
+static t_time	ft_st_calculate_seconds(t_time time)
+{
+	time.days = time.years % YEAR_SECONDS;
+	time.years /= YEAR_SECONDS;
+	time.hours = time.days % DAYS_SECONDS;
+	time.days /= DAYS_SECONDS;
+	time.minutes = time.hours % HOURS_SECONDS;
+	time.hours /= HOURS_SECONDS;
+	time.seconds = time.minutes % 60;
+	time.minutes /= 60;
```
clafoutis42, line 125: Nice proceeding!  
Although I would have built everything whilst parsing to avoid allocation memory on structures.

chall09/edramire.c
```diff
@@ -0,0 +1,157 @@
...
+{
+	char	*aux;
+
+	if (time == 0)
+		return ;
+	if (prev > 0)
+	{
+		aux = strdup(str);
+		if (aux == NULL)
+			return ;
```
clafoutis42, line 107: There should be a way to notify the caller that the malloc failed.  
Returning a code, or setting `str` to null (you would have to change it to a higher level pointer)

chall09/edramire.c
```diff
@@ -0,0 +1,157 @@
...
+		return (NULL);
+	bzero(str, sizeof(char) * 256);
+	ft_stc_format_time(str, "year", 0, 0, my_time.years);
+	ft_stc_format_time(str, "day", my_time.flags & 0x10, my_time.flags & 0x7,
+		my_time.days);
+	ft_stc_format_time(str, "hour", my_time.flags & 0x18, my_time.flags & 0x3,
+		my_time.hours);
+	ft_stc_format_time(str, "minute", my_time.flags & 0x1C, my_time.flags & 0x1,
+		my_time.minutes);
+	ft_stc_format_time(str, "second", my_time.flags & 0x1E, 0, my_time.seconds);
```
clafoutis42, line 155: Nice use of binary operators, but its a little overkill, given that you could simply have summed the next and previous values instead.

clafoutis42 (commented):
Nice solution, however I'm not a real fan of allocation this much memory on structures.  
But it works! Good job!

PR #245, opened 2020-05-10
vsempere - Telescopium

victorsempere:
First of all. Checks (nothing different than numbers. No more than one number, no negative number. Number max allowed value...)

Split number per unit of time: years, days, hours, ...
Build each value and prepend the conjunction (", " or " and ")
Concat all values
Return result


clafoutis42:
Please rebase your branch to only embed your chall09, and create another branch with your chall10 to create a new PR for it .

victorsempere:
I'm not quite sure if I did your request properly. But now there are two PR one from Chall09 and another from Chall10

clafoutis42:
No, you didn't do it properly, but that's alright, you're learning!  
You should have kept the old one for chall09 (by removing the undesired commits from it)
Look at the `rebase` and `reset` features of git :)

PR #246, opened 2020-05-10
chall09 - gbudau

gbudau:
**How i solved this problem:**
Validate the string then convert it to `unsigned long` which is enough to hold the `MAX_SECONDS`.
Then if its a positive number, split the `seconds` into `YEARS`, `DAYS`, `MINUTES` and `SECONDS` using divide and modulo.
Calculate and allocate enough memory to hold the formated string then i'm using `snprintf()` to format and copy the string  into the allocated memory.

**Short description of the code:**
Created some #define constants and a static constant array to make everything easier to read.

`get_seconds()`  is converting the string to unsigned long using `strtoul()` and also checking for errors.

In `create_date()` i create an array that contains the number of the units of time and use
divide and modulo to determine each of them.

In `format_time()` i calculate and allocate the memory needed to create the formated string.

`unit_length()` returns the length of a number for example 42 returns 2.
`total_length()` returns the memory that needs to be allocated for the formated string without counting the  separators `", "` and `" and "` and also counts the number of positive units of time.
`separator_length()` returns the memory that needs to be allocated for the separators `", "` and `" and "`.

Last step is to use `fill_time()` to create the formated string using the allocated memory, the array that contains the count of units of time, the length of the allocated memory and the number of positive units of time.
I format and copy into the the allocated memory with `snprintf()`, using the return value to aument the index and also to substract from the length of the allocated memory to make sure it's not going out of boundry.


chall09/gbudau.c
```diff
@@ -0,0 +1,164 @@
...
+	unsigned long	time[ARRAY_SIZE];
+
+	time[YEARS] = seconds / YEAR;
+	seconds %= YEAR;
+	time[DAYS] = seconds / DAY;
+	seconds %= DAY;
+	time[HOURS] = seconds / HOUR;
+	seconds %= HOUR;
+	time[MINUTES] = seconds / MINUTE;
+	time[SECONDS] = seconds % MINUTE;
```
clafoutis42, line 147: Nice, however I would have built the string whilst parsing to avoir using an array. The good point here is that you're not allocating any structures unnecessarily

chall09/gbudau.c
```diff
@@ -0,0 +1,164 @@
...
+	return (format_time(time));
+}
+
+char					*ft_format_duration(char *seconds)
+{
+	unsigned long	ul_seconds;
+	int				err;
+
+	if (*seconds == '\0')
+		return (strdup("Invalid input."));
```
clafoutis42, line 157: Maybe one function that parses the string into a number could have detected the error by allowing you to only have one error condition in here.

chall09/gbudau.c
```diff
@@ -0,0 +1,164 @@
...
+	if (positive == 1)
+		return (0);
+	return (strlen(", ") * (positive - 2) + strlen(" and "));
+}
+
+static char				*free_fmt(char *format)
+{
+	if (format != NULL)
+		free(format);
+	return (format = NULL);
```
clafoutis42, line 90: There's no use in doing `format = NULL` here. Format won't ever be used again as you exit de function.

chall09/gbudau.c
```diff
@@ -0,0 +1,164 @@
...
+			length -=ret;
+			if (positive >= 2)
+			{
+				ret = snprintf(&format[idx] , length, positive > 2 ? ", " : " and ");
+				if (ret < 0)
+					return (free_fmt(format));
+				idx += ret;
+				length -= ret;
+			}
+			positive--;
```
clafoutis42, line 115: Again, I think you could have done this whilst parsing (by parsing I mean the step where you divide the number of seconds to get the amount of each units.

clafoutis42 (commented):
Good job! You have a working solution. Although, it could have been optimized a little more :)

chall09/gbudau.c
gbudau: Good point, thank you.

chall09/gbudau.c
gbudau: You are right, this condition is not necessary here since it will be checked in `get_seconds()`

chall09/gbudau.c
gbudau: Thank you. That's a smart solution :smiley:

gbudau:
Thank you Charles!  :+1:

PR #247, opened 2020-05-10
Telescopium dalba-de

dalba-de:
# Telescopium Solution

First I go through the string to check its validity. Then I skip the spaces and use my own atoi function that allows me to check if there is something different to spaces after the entered number which would make it an invalid input.

Once verified that the string is correct, I enter the function that allows me to fill in the time in the returned string, checking for each space of time if the rest are 0, 1 or greater, so that I can correctly return the gap (seconds or second, and or comma...).

Finally I fill the string to return with the `sprintf` function.


clafoutis42:
There are conflicts with the README, please solve them before I can merge your PR :)

chall09/dalba-de.c
```diff
@@ -0,0 +1,189 @@
...
+			while (seconds[i])
+			{
+				if (seconds[i] != ' ')
+					return ("Invalid input");
+				i++;
+			}
+		}
+	}
+	ret = fill_time(num);
+	return (ret);
```
clafoutis42, line 188: Why not this:
```suggestion
return (fill_time(num));
```

chall09/dalba-de.c
```diff
@@ -0,0 +1,189 @@
...
+	unsigned long long		day;
+	unsigned long long		hour;
+	unsigned long long		minute;
+	unsigned long long 		second;
+	char					*str;
+	char					*cyear;
+	char					*cday;
+	char					*chour;
+	char					*cminute;
+	char					*csecond;
```
clafoutis42, line 17: This is a lot of elements for the same thing, you could have built the string whilst parsing to avoir allocating structures.

chall09/dalba-de.c
```diff
@@ -0,0 +1,189 @@
...
+		gap = " and ";
+	else
+		gap = "";
+  	t->csecond = malloc(sizeof(t->second) + 9 * sizeof(char));
+	if (t->second == 0)
+		sprintf(t->csecond, "");
+	else if (t->second == 1)
+		sprintf(t->csecond, "%s%llu second", gap, t->second);
+	else
+		sprintf(t->csecond, "%s%llu seconds", gap, t->second);
```
clafoutis42, line 65: All those functions are a little repetitive, don't you think? Maybe you could have made only one with different parameters !

chall09/dalba-de.c
```diff
@@ -0,0 +1,189 @@
...
+	t.hour = num % 60;
+	num /= 24;
+	t.day = num % 365;
+	t.year = num / 365;
+	fill_second(&t);
+	fill_minute(&t);
+	fill_hour(&t);
+	fill_day(&t);
+	fill_year(&t); 
+	sprintf(t.str, "%s%s%s%s%s", t.cyear, t.cday, t.chour, t.cminute, t.csecond);
```
clafoutis42, line 157: What bothers me here is that you're never freeing any of the memory you allocated to build the final one

chall09/dalba-de.c
```diff
@@ -0,0 +1,189 @@
...
+
+void      fill_hour(t_time *t)
+{
+	char *gap;
+
+	gap = "";
+	if ((t->year > 0 || t->day > 0) && (t->second == 0 && t->minute == 0))
+		gap = " and ";
+	else if ((t->year > 0 || t->day > 0) && (t->second > 0 || t->minute > 0))
+		gap = " , ";
```
clafoutis42, line 96: There should not be any space before de comma
```suggestion
		gap = ", ";
```

clafoutis42 (commented):
Your solution seems to work fine. Although you are printing a space before each commas, and this will make your program fail the tests. Sorry :(

PR #248, opened 2020-05-10
Changed error Handle and memory leaks

byrd92:
I create all returns with malloc so they can be free at main.
Also i solved error handle. I had a previous pull request explaining the code but I´m not sure if i cancelled it, i hope nope. 
Salut!

_I first check some errors like no numbers , symbols. Also check if its zero to return it.
then I read each (years, then days, hours, etc) making some division and saving it on 5 variables.
To handle the string I use a tmp string to help me. I know it could be done in a function and use some bucles but i got it faster like this. Also i check if the next variable its 0 (like minutes its 0, so it means hours is lasts, so i put the and before Hours).
Also i got some string help from sprintf.
thanks!!!_


chall09/egarcia-.c
```diff
@@ -0,0 +1,179 @@
...
+
+char   *ft_print_format(int y ,int d, int h, int m, int s)
+{
+    char *str;
+    char *tmp;
+    char *years = strdup("");
+    char *days = strdup("");
+    char *hours = strdup("");
+    char *minutes = strdup("");
+    char *seconds = strdup("");
```
clafoutis42, line 35: This is really awful, you should keep them to `NULL` and only allocate memory if you need it.

chall09/egarcia-.c
```diff
@@ -0,0 +1,179 @@
...
+	i = 0;
+	if (s1 && s2)
+	{
+		if (!(str = (char*)malloc((strlen(s1) + strlen(s2) + 1)
+			* sizeof(char))))
+			return (NULL);
+		while (*s1)
+			str[i++] = *s1++;
+		while (*s2)
+			str[i++] = *s2++;
```
clafoutis42, line 20: Why not using strcpy and strcat here?

chall09/egarcia-.c
```diff
@@ -0,0 +1,179 @@
...
+		str[i] = '\0';
+		return (str);
+	}
+	return (0);
+}
+
+char   *ft_print_format(int y ,int d, int h, int m, int s)
```
clafoutis42, line 27: This conversion doesn't fit the required type, here you are converting a signed int. Which means that if you go over `INT_MAX` you will go around and lose the initial value.  
For example, if you convert `9999999999999999999 seconds`, you will get `317097919837` which is higher than `2,147,483,647` (`INT_MAX`).  
Thus you won't pass this test.

Type choosing is really important in programmation, it has to fit the needs!

chall09/egarcia-.c
```diff
@@ -0,0 +1,179 @@
...
+    char *years = strdup("");
+    char *days = strdup("");
+    char *hours = strdup("");
+    char *minutes = strdup("");
+    char *seconds = strdup("");
+    if ( y > 0)
+    {
+
+        sprintf(years, "%d", y);
+        tmp = y > 1 ? ft_strjoin(years, " years") : ft_strjoin(years, " year");
```
clafoutis42, line 40: So many allocations! Why don't you simply do:
```suggestion
        sprintf(years, "%d %s", y, (y > 1 ? "years" : "year"));
```
This commentary is the same for every other cases.

chall09/egarcia-.c
```diff
@@ -0,0 +1,179 @@
...
+        {
+            m = n / 60;
+            n %= 60;
+        }
+        else if (n < 60)
+        {
+            s = n;
+            n = 0;
+        }
+    }
```
clafoutis42, line 175: Using macros to define the numbers by what they represent would have helped reading :)

chall09/egarcia-.c
```diff
@@ -0,0 +1,179 @@
...
+    long long int n = atoll(seconds);
+    int d = 0;
+    int h = 0;
+    int m = 0;
+    int s = 0;
+    int y = 0;
+    int i = 0;
+    
+    char *invalidinput = strdup("Invalid input.");
+    char *now = strdup("now");
```
clafoutis42, line 125: Why would you allocate things you are not sure to use?
Especially since you're only freeing `now` when you return a valid result, and you are not freeing `invalidninput` when you return `now`.

clafoutis42 (commented):
Your algorithm seems to work. However you really need to work on optimizing the way you use the memory.
Also you won't pass the test because you are using a type int for the year count when it might actually get higher.

PR #249, opened 2020-05-10
chall09 - jbaringo

baringo11:
First I check if the seconds string is okey, if it is right, then I pass the number of the string to a int number. then I calculate the minutes, hours... and pass each time to a string. After that I join each string to the result string with the designated name breaking them up with "," or ,if it is the final time, with "and".

### DIFF
--- a/chall09/jbaringo.c
+++ b/chall09/jbaringo.c
@@ -1,0 +1,194 @@
+#include <stdlib.h>
+#include <string.h>
+
+int         ft_isdigit(char c)
+{
+    if (c < '0' || c > '9')
+        return (0);
+    return (1);
+}
+
+
+int             get_size(unsigned long long num)
+{
+    int i = 0;
+
+    while (num >= 10)
+    {
+        i++;
+        num /= 10;
+    }
+    return (i + 1);
+}
+
+char			*ft_long_itoa(unsigned long long num)
+{
+	char		    *str;
+	int	index;
+	int	size;
+    
+	size = get_size(num);
+	index = 0;
+	if (!(str = malloc(sizeof(char) * (size + 1 ))))
+		return (0);
+
+	index = size - 1;
+	while (num >= 10)
+	{
+		str[index--] = (num % 10 + 48);
+		num /= 10;
+	}
+	str[index] = (num % 10 + 48);
+	str[size] = '\0';
+	return (str);
+}
+
+unsigned long long  ft_long_atoi(char *str)
+{
+    unsigned long long num = 0;
+    int i = 0;
+
+	while (str[i] == ' ' || (str[i] >= 9 && str[i] <= 13))
+		i++;
+	while (str[i] && str[i] >= '0' && str[i] <= '9')
+	{
+		num = num * 10 + (str[i] - 48);
+		i++;
+	}
+	return (num);
+}
+
+char        *ft_strjoin(char *s1, char*s2)
+{
+    int i = 0;
+    int j = 0;
+    char *s3;
+    if (!(s3 = malloc(sizeof(char) * (strlen(s1) + strlen(s2) + 1))))
+        return (NULL);
+    while (s1[i])
+    {
+        s3[i] = s1[i];
+        i++;
+    }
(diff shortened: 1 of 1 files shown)


chall09/jbaringo.c
```diff
@@ -0,0 +1,194 @@
...
+    }
+    return (1);
+}
+
+char        *ft_format_duration(char *seconds)
+{
+    unsigned long long time = ft_long_atoi(seconds);
+
+    if (!check_seconds(seconds) || !strlen(seconds))
+        return ("invalid input");
```
aguimbao, line 186: You had to return `"Invalid input."`, not `"invalid input"`. :(((

chall09/jbaringo.c
```diff
@@ -0,0 +1,194 @@
+#include <stdlib.h>
+#include <string.h>
+
+int         ft_isdigit(char c)
```
aguimbao, line 4: You can use the regular `isdigit()` function, no need to code your own 👍

chall09/jbaringo.c
```diff
@@ -0,0 +1,194 @@
...
+char        *ft_format_duration(char *seconds)
+{
+    unsigned long long time = ft_long_atoi(seconds);
+
+    if (!check_seconds(seconds) || !strlen(seconds))
+        return ("invalid input");
+    if (time > 0)
+        return(get_time(time));
+    else if (time == 0)
+        return ("now");
```
aguimbao, line 190: Also, don't return a mix of `string literals` and dynamically allocated memory, it's not good practice. What would happen if we `free()`'d your returned answer at the end of your program to see if you have other leaks in the case where you `return ("now");` ?

chall09/jbaringo.c
```diff
@@ -0,0 +1,194 @@
...
+		i++;
+	while (str[i] && str[i] >= '0' && str[i] <= '9')
+	{
+		num = num * 10 + (str[i] - 48);
+		i++;
+	}
+	return (num);
+}
+
+char        *ft_strjoin(char *s1, char*s2)
```
aguimbao, line 61: Your `ft_strjoin()` can be simplified to:

```C
char* result = malloc(strlen(s1) + strlen(s2) + 1);
// Check for malloc error
strcat(strcpy(result, s1), s2);
```

chall09/jbaringo.c
```diff
@@ -0,0 +1,194 @@
...
+	{
+		str[index--] = (num % 10 + 48);
+		num /= 10;
+	}
+	str[index] = (num % 10 + 48);
+	str[size] = '\0';
+	return (str);
+}
+
+unsigned long long  ft_long_atoi(char *str)
```
aguimbao, line 46: Just a hint, there was a special function to transform from `char *` to `unsigned long long` called `strtoull()` that could also help you with error handling, recoding your own version of a better `atoi()` was a bit of overkill, but good job nonetheless.

The problem here and in your error checking, is that you forgot to check if the provided `char *` was `> 10000000000000000000` and if it was superior to `ULLONG_MAX` (from `<limits.h>`), which `strtoull()` would have mostly handled for you.

chall09/jbaringo.c
```diff
@@ -0,0 +1,194 @@
...
+            i++;
+        else
+            return (0);
+    }
+    return (1);
+}
+
+char        *ft_format_duration(char *seconds)
+{
+    unsigned long long time = ft_long_atoi(seconds);
```
aguimbao, line 183: You do unnecessary work in case `seconds` is invalid.

aguimbao (commented):
Your solution was nice, there's a lot of "unnecessary work", because there were basically functions that could do most of the heavy work for you, like `strtoull()`, `sprintf()` and `asprintf()` (I recommend you check them out), etc. Code is simple and readable and I think the way you handle the amount of time units in lines 138, 139, 140, 141 and 142 is very nice !

It's a shame you didn't really check how `"Invalid input."` had to be formatted, because your program worked flawlessly except for that and the `ULLONG_MAX` and `10000000000000000000` checks.

Good job :)

PR #250, opened 2020-05-11
chall 09

szavarce42:
Sorry about the mess, i think it works. It was too late to clean the code and I have to go to bed.


clafoutis42:
I really would have liked you to explain your code :(  
I will close this PR and will not review it.

PR #251, opened 2020-05-11
Incomplete solution

bazuara:
the function receives a string with seconds in it. It first converts it to a int via atoi, then stores it in sc variable.
Then,dividing by the corresponding ratios, it stores minutes, hours, days and years. The remainder is stored in the same variables.
Using fprintf, it converts from int to string storing it into a temp variable.
Using strjoin_free we allocate, join and free all the variables in to a final string.
TODO: condition the text variables depending on units (singular or plural) and the links between them based on empty units.

### DIFF
--- a/chall09/bazuara.c
+++ b/chall09/bazuara.c
@@ -6,12 +6,13 @@
 /*   By: bazuara <bazuara@student.42madrid.>        +#+  +:+       +#+        */
 /*                                                +#+#+#+#+#+   +#+           */
 /*   Created: 2020/05/09 14:07:16 by bazuara           #+#    #+#             */
-/*   Updated: 2020/05/09 15:42:44 by bazuara          ###   ########.fr       */
+/*   Updated: 2020/05/09 16:38:58 by bazuara          ###   ########.fr       */
 /*                                                                            */
 /* ************************************************************************** */
 
 #include <stdlib.h>
 #include <stdio.h>
+#include <string.h>
 
 char	*ft_format_duration(char *seconds)
 {
@@ -23,6 +24,13 @@ char	*ft_format_duration(char *seconds)
 	char	*result;
 
 	sc = atoi(seconds);
+	if (sc == 0)
+	{
+		if (strcmp(seconds, "0") == 0)
+			return ("now");
+		else
+			return ("Invalid input.");
+	}
 	mn = sc / 60;
 	hr = mn / 60;
 	dy = hr / 24;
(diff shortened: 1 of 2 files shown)


chall09/bazuara.c
```diff
@@ -0,0 +1,131 @@
...
+			new[i] = s1[i];
+			i++;
+		}
+		j = 0;
+		while (j < strlen(s2))
+		{
+			new[i + j] = s2[j];
+			j++;
+		}
+		new[i + j] = '\0';
```
clafoutis42, line 41: You could have used `strcpy` and `strcat` here ;)

chall09/bazuara.c
```diff
@@ -0,0 +1,131 @@
...
+	temp = ft_strjoin(str1, str2);
+	if (i == 1)
+		free(str1);
+	else if (i == 2)
+		free(str2);
+	if (i == 3)
+	{
+		free(str1);
+		free(str2);
+	}
```
clafoutis42, line 59: A little advice on making this shorter and easier to read
```suggestion
	if (i == 1 || i == 3)
		free(str1);
	if (i == 2 || i == 3)
		free(str2);
```

chall09/bazuara.c
```diff
@@ -0,0 +1,131 @@
...
+		result = ft_strjoin_free( result, ft_strjoin_free( temp, " días, ", 0), 3);
+	sprintf(temp, "%d", hr);
+	if (hr != 0)
+		result = ft_strjoin_free( result, ft_strjoin_free( temp, " horas, ", 0), 3);
+	sprintf(temp, "%d", mn);
+	if (mn != 0)
+		result = ft_strjoin_free( result, ft_strjoin_free( temp, " minutos y ", 0), 3);
+	sprintf(temp, "%d", sc);
+	if (sc != 0)
+		result = ft_strjoin_free( result, ft_strjoin_free( temp, " segundos.", 0), 3);
```
clafoutis42, line 104: Instead of using this much `strjoin`'s you could have use `sprintf` to format your string directly! e.g:
```c
sprintf(temp, "%d %s", years, (years > 1 ? "years" : "year"));
```

chall09/bazuara.c
```diff
@@ -0,0 +1,131 @@
...
+	}
+	mn = sc / 60;
+	hr = mn / 60;
+	dy = hr / 24;
+	yr = dy / 365;
+	//printf("respre: s:%d m:%d, h:%d, d:%d,  y:%d\n", sc, mn, hr, dy, yr);
+	sc = sc % 60;
+	mn = mn % 60;
+	hr = hr % 24;
+	dy = dy % 365;
```
clafoutis42, line 89: To help you write faster, you could use this syntax:
```suggestion
	sc %= 60;
	mn %= 60;
	hr %= 24;
	dy %= 365;
```
This works exactly the same with all operators ;)

chall09/bazuara.c
```diff
@@ -0,0 +1,131 @@
...
+	char	*result;
+	char	temp[50];
+
+	sc = atoi(seconds);
+	if (sc <= 0)
+	{
+		if (strcmp(seconds, "0") == 0)
+			return ("now");
+		else
+			return ("Invalid input.");
```
clafoutis42, line 79: As you are going to need to allocate memory in most cases, you should always return allocated memory, here the user might be confused trying to free the memory that was not allocated.

clafoutis42 (commented):
Although it is not complete, you did try to solve it, and it's something valuable! You'll get the point for submitting something ;)

PR #252, opened 2020-05-11
chall09 isoria-g

nsoria042:
The first thing is to analyze the input string and verify that it is a valid number to work with and I use ATOI (controlling characters before and after). Then I check the "Invalid input".
The number I work with ATOI. The first thing is to analyze the input string and verify that it is a valid number to work with and then I use ATOI (controlling characters before and after). Then I check the "Invalid input".
I work the number dividing by the units in seconds and I build the string with strjoin and ITOA.

### DIFF
--- a/chall09/isoria-g.c
+++ b/chall09/isoria-g.c
@@ -1,0 +1,333 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   isoria-g.c                                         :+:      :+:    :+:   */
+/*                                                    +:+ +:+         +:+     */
+/*   By: yo <yo@student.42.fr>                      +#+  +:+       +#+        */
+/*                                                +#+#+#+#+#+   +#+           */
+/*   Created: 2020/05/11 00:02:38 by marvin            #+#    #+#             */
+/*   Updated: 2020/05/11 10:19:38 by yo               ###   ########.fr       */
+/*                                                                            */
+/* ************************************************************************** */
+
+#include <unistd.h>
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+
+size_t      ft_strlen(const char *s)
+{
+    size_t  a;
+
+    a = 0;
+    while (s[a] != '\0')
+        a++;
+    return (a);
+}
+
+char                *ft_strjoin(char *s1, char *s2)
+{
+    unsigned int    a;
+    unsigned int    l1;
+    unsigned int    l2;
+    char            *t;
+    char            *str1;
+
+    a = 0;
+    if (s1 == NULL || s2 == NULL)
+        return (NULL);
+    l1 = (unsigned int)ft_strlen((char *)s1);
+    l2 = (unsigned int)ft_strlen((char *)s2);
+    if (!(t = (char *)malloc((l1 + l2) * sizeof(char) + 1)))
+        return (NULL);
+    str1 = (char *)t;
+    while (a < (l1 + l2))
+    {
+        if (a < l1)
+            str1[a] = s1[a];
+        else
+            str1[a] = s2[a - l1];
+        a++;
+    }
+    str1[a] = '\0';
+    return (t);
+}
+
+static unsigned long long   ft_value(unsigned long long n)
+{
+    unsigned long long      b;
+
+    if (n < 0)
+        b = n * (-1);
+    else
+        b = n;
+    return (b);
+}
+
+static unsigned long long   ft_size(unsigned long long b)
+{
+    unsigned long long      a;
+
+    a = 0;
+    while (b > 9)
+    {
+        b = b / 10;
+        a++;
+    }
+    return (a);
+}
+
+static unsigned long long   ft_end(unsigned long long n)
+{
+    unsigned long long      c;
+
+    if (n < 0)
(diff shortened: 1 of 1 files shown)


nsoria042:
The first thing is to analyze the input string and verify that it is a valid number to work with and I use ATOI (controlling characters before and after). Then I check the "Invalid input".
The number I work with ATOI. The first thing is to analyze the input string and verify that it is a valid number to work with and then I use ATOI (controlling characters before and after). Then I check the "Invalid input".
I work the number dividing by the units in seconds and I build the string with strjoin and ITOA.

chall09/isoria-g.c
```diff
@@ -0,0 +1,335 @@
...
+
+size_t      ft_strlen(const char *s)
+{
+    size_t  a;
+
+    a = 0;
+    while (s[a] != '\0')
+        a++;
+    return (a);
+}
```
clafoutis42, line 28: Why not use standard library's `strlen` ?

chall09/isoria-g.c
```diff
@@ -0,0 +1,335 @@
...
+    str1 = (char *)t;
+    while (a < (l1 + l2))
+    {
+        if (a < l1)
+            str1[a] = s1[a];
+        else
+            str1[a] = s2[a - l1];
+        a++;
+    }
+    str1[a] = '\0';
```
clafoutis42, line 54: You could have used `strcpy` and `strcat` to do this ;)

chall09/isoria-g.c
```diff
@@ -0,0 +1,335 @@
...
+    unsigned int    l1;
+    unsigned int    l2;
+    char            *t;
+    char            *str1;
+
+    a = 0;
+    if (s1 == NULL || s2 == NULL)
+        return (NULL);
+    l1 = (unsigned int)ft_strlen((char *)s1);
+    l2 = (unsigned int)ft_strlen((char *)s2);
```
clafoutis42, line 42: Why are you casting `s1` and `s2` ? They already are of type `char *`

chall09/isoria-g.c
```diff
@@ -0,0 +1,335 @@
...
+static unsigned long long   ft_value(unsigned long long n)
+{
+    unsigned long long      b;
+
+    if (n < 0)
+        b = n * (-1);
+    else
+        b = n;
+    return (b);
+}
```
clafoutis42, line 67: Unsigned variables can never be negative

chall09/isoria-g.c
```diff
@@ -0,0 +1,335 @@
...
+
+char                    *ft_format_duration(char *seconds)
+{
+    unsigned long long  nb;
+    unsigned long long  yy;
+    unsigned long long  dd;
+    unsigned long long  hh;
+    unsigned long long  mm;
+    unsigned long long  ss;
+    char *str = "";
```
clafoutis42, line 188: It is better to initiate it with `NULL`

chall09/isoria-g.c
```diff
@@ -0,0 +1,335 @@
...
+
+    if (seconds == NULL)
+        return("\"Invalid input.\"");
+    nb = ft_atoi(seconds);
+    if (nb == -1)
+        return("\"Invalid input.\"");
+    if (nb == 0)
+        return("\"now\"");
```
clafoutis42, line 196: Why did you return this values with a quote?

chall09/isoria-g.c
```diff
@@ -0,0 +1,335 @@
...
+    yy = nb / 31536000;
+    dd = (nb % 31536000) / 86400;
+    hh = ((nb % 31536000) % 86400) / 3600;
+    mm = (((nb % 31536000) % 86400) % 3600) / 60;
+    ss = (((nb % 31536000) % 86400) % 3600) % 60;
```
clafoutis42, line 201: Using macros to define those numbers by what they represent would have helped readability

chall09/isoria-g.c
```diff
@@ -0,0 +1,335 @@
...
+        if (yy == 1)
+        {
+            str = ft_strjoin(str, ft_itoa(yy));
+            str = ft_strjoin(str, " year");
+        }
+        else
+        {
+            str = ft_strjoin(str, ft_itoa(yy));
+            str = ft_strjoin(str, " years");
+        }
```
clafoutis42, line 214: This seems overly repetitive. You should have made a function for this.  
Even better you could have used functions like `sprintf` or other variations of it to format your string easily

chall09/isoria-g.c
```diff
@@ -0,0 +1,335 @@
...
+            str = ft_strjoin(str, ft_itoa(ss));
+            str = ft_strjoin(str, " second");
+        }
+        else
+        {
+            str = ft_strjoin(str, ft_itoa(ss));
+            str = ft_strjoin(str, " seconds");
+        }
+    }
+    str = ft_strjoin(str, "\"");
```
clafoutis42, line 333: Why are you quoting the result ? :/

clafoutis42 (commented):
Your code seems to work, but you are overly repetitive and you added quotes for a reason I don't get.  
Unfortunately this won't work, but nice try!

clafoutis42:
I just noticed your commit was initially late, I'm gonna have to revert this PR, sorry.

PR #253, opened 2020-05-16
edramire - Kepler

edramir18:
For this challenge i made an array of 6 rows by 7 columns which will be held the value for each cell of the game. Using an enumeration class i choose between empty, red or blue cell to place inside the array.
Using a regular expression i check the input line is according to the schema a letter from A to G, one underscore character and a word starting for a capitalized letter.
In case of any invalid input a error message is show. I take as invalid any input wich has more of 42 plays, more of 6 moves for column, consecutive plays for the same player or plays after a player has win the match. As well, if more of two kind of player is supplied, that input is checked as invalid.
The chips are stored in the array from the last row to the first, to emulate the original game, that way the 6th row is filled first for each column of the array
After each play the winner is verified starting from the position of the last chip, in case more of 4 chip are connected that player is declared as the winner of the match. To check the horizontal and the diagonals i use the line equation to generate each position to be check, and for the vertical just checked each position for that column.


aguimbao:
Hey, I'm gonna start checking out this PR, but sadly I won't be able to merge it unless you fix the conflicts ! We only need the `<login>.py` file, not your test files which were also added in the commit :)

Edit: Fixed it myself.

chall10/edramire.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def load_line(self, line: str, cell):
+        if re.search('^[A-G]_[A-Z][a-z]+$', line) is not None:
+            idx, name = line.split('_')
+            if self.red is None:
+                self.red = name
+            elif self.blue is None:
+                self.blue = name
+            elif not (name == self.blue or name == self.red):
+                raise ValueError
```
aguimbao, line 58: Don't worry for the moment we won't ask you to raise specific exceptions. Just to handle graciously the input and return `"Invalid input ."` when needed. But it's still good practice !

chall10/edramire.py
```diff
@@ -0,0 +1,110 @@
...
+            return
+    if game.win is None:
+        print('Draw.')
+    else:
+        print(f'{game.red if game.win == Cell.RED else game.blue}.')
+    game.print_grid()
+
+
+if __name__ == '__main__':
+    connect42()
```
aguimbao, line 110: The convention here is to execute `main()`, not a custom function, but it's not really an error !

chall10/edramire.py
```diff
@@ -0,0 +1,110 @@
...
+            else:
+                raise ValueError
+        except ValueError:
+            print('Invalid input.')
+            return
+    if game.win is None:
+        print('Draw.')
+    else:
+        print(f'{game.red if game.win == Cell.RED else game.blue}.')
+    game.print_grid()
```
aguimbao, line 106: This is probably a debugging statement, but we were only asking you to print who won, or `"Draw."`, accordingly :( so you fail most tests with this.

chall10/edramire.py
```diff
@@ -0,0 +1,110 @@
...
+        try:
+            if game.win is None and count < 42:
+                cell = Cell.RED if count % 2 == 0 else Cell.BLUE
+                idx, cell = game.load_line(line[:-1], cell)
+                game.play(idx, cell)
+                count += 1
+            else:
+                raise ValueError
+        except ValueError:
+            print('Invalid input.')
```
aguimbao, line 100: You could do this in a simple statement with `sys.exit("Invalid input.")`. Don't be scared to `exit` a program, it doesn't always mean there has been an error. You usually don't `return` from the main function in python.

chall10/edramire.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def connect42():
+    game = Connect()
+    count = 0
+    for line in sys.stdin:
+        try:
+            if game.win is None and count < 42:
+                cell = Cell.RED if count % 2 == 0 else Cell.BLUE
+                idx, cell = game.load_line(line[:-1], cell)
```
aguimbao, line 94: So this has a couple problems. In the case where we fed your program line by line, this would probably be the input you would receive:

```zsh
</$> ./edramire.py
"A_Red\n
B_Yellow\n
A_Red\n
B_Yellow\n
A_Red\n
B_Yellow\n
G_Red\n
F_Yellow\n
<EOF>(Ctrl-D)"
```

Your line[:-1] works here because you trim the `\n` at the end of each line. But we actually feed our test cases from files that look like this:

```txt
A_Red\n
B_Yellow\n
A_Red\n
B_Yellow\n
A_Red\n
B_Yellow\n
G_Red\n
F_Yellow<EOF>
```

(As most .txt files end usually by an `EOF` character, rather by a blank line at the end followed by the `EOF` character.)

So your program fails because you interpret the last line as `"F_Yello"`.

chall10/edramire.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def print_grid(self):
+        for row in self.grid:
+            print(''.join([str(x) for x in row]))
+
+
+def connect42():
+    game = Connect()
+    count = 0
+    for line in sys.stdin:
```
aguimbao, line 90: This is better than my implementation where I use `sys.stdin.read()`, as you can catch errors in the input sooner this way 👍

chall10/edramire.py
```diff
@@ -0,0 +1,110 @@
+#! /usr/bin/env python3
+
+import sys
+import re
+from enum import Enum
+
+
+class Cell(Enum):
```
aguimbao, line 8: I didn't know there were enums in python 😅 I think they are a bit overkill here, but they would definitely be useful if you had 2 kind of cells like:

```python3
class FriendlyCell(Enum):
    EMPTY = '.'
    FULL = 'X'

class EnemyCell(Enum):
    EMPTY = '.'
    FULL = 'O'
```

But I would have used global definitions here, less overhead.

chall10/edramire.py
```diff
@@ -0,0 +1,110 @@
...
+        for row in self.grid:
+            print(''.join([str(x) for x in row]))
+
+
+def connect42():
+    game = Connect()
+    count = 0
+    for line in sys.stdin:
+        try:
+            if game.win is None and count < 42:
```
aguimbao, line 92: There's a small fail in your logic here, if there's a winner in the middle of the game, but there are still remaining lines to be fed to the program as input, for example:

```zsh
<$/> ./edramire.py
A_Yellow
A_Yellow
A_Yellow
A_Yellow
B_Red
```
You return `"Invalid input."` because there's lines after a team won.

chall10/edramire.py
```diff
@@ -0,0 +1,110 @@
...
+            if 0 <= row < 6:
+                if cell == self.grid[row][col]:
+                    count += 1
+                elif count < 4:
+                    count = 0
+            col += 1
+        return count == 4
+
+    def load_line(self, line: str, cell):
+        if re.search('^[A-G]_[A-Z][a-z]+$', line) is not None:
```
aguimbao, line 51: Nice regex ;), only fail I see is that `A_Team2` or `A_TEAM` should be valid too !

chall10/edramire.py
```diff
@@ -0,0 +1,110 @@
+#! /usr/bin/env python3
```
aguimbao, line 1: Be careful with no specificating >= `python3.5` in the shebang when you use features specific from python > 3.5, like `f-strings`.

aguimbao (commented):
Nice solution :) your checks are fine, I think using a class here was a bit of overkill, but it's up to debate !

chall10/edramire.py
edramir18: That part still bugging me a lot. I always have doubts about how to handle the end of file.

chall10/edramire.py
edramir18: i forgot to comment that line. Is part of my debugging tests

chall10/edramire.py
edramir18: I used enumerations mostly for debugging purpose, overwriting the string method to make possible  to the change of the string representation of each cell of the grid according to the content.

PR #254, opened 2020-05-16
chall10 | estina

xplore93:
This is my try for Connect 4 game.
I save all the strings into a list, then check each line for errors meanwhile splitting them. Then I check if there are only 2 colors and put it into the board at the same time. Afterall I check if there is a Winner.
End.


chall10/estina.py
```diff
@@ -0,0 +1,88 @@
...
+		exit("Draw.")
+	elif lines_len > 42:
+		exit("Invalid input.")
+	board = [" "] * 42
+	colors = []
+	for line in lines:
+		split_line = line.split("_")
+		index = check_line(line, colors, split_line)
+		for i in range(COLS):
+			if board[index + i * COLS] == " ":
```
clafoutis42, line 79: Watch out, you did not protect the size, if you put more than 7 objects in the same column it crashes

chall10/estina.py
```diff
@@ -0,0 +1,88 @@
...
+	if len(split_line[0]) != 1:
+		exit("Invalid input.")
+	if not len(colors) or (len(colors) != 2 and colors[0] != split_line[1]):
+		colors.append(split_line[1])
+	elif colors[0] != split_line[1] and colors[1] != split_line[1]:
+		exit("Invalid input.")
+	index = ord(split_line[0]) - ord('A')
+	if index < 0 or index > 6 or len(split_line) != 2:
+		exit("Invalid input.")
+	return index
```
clafoutis42, line 18: You forgot to check the first letter of the column. It should be uppercase, you are failing one of the examples!

chall10/estina.py
```diff
@@ -0,0 +1,88 @@
...
+		split_line = line.split("_")
+		index = check_line(line, colors, split_line)
+		for i in range(COLS):
+			if board[index + i * COLS] == " ":
+				board[index + i * COLS] = split_line[1]
+				break
+	for cell in board:
+		if not board:
+			board = " "
+	print(check_winner(board, colors))
```
clafoutis42, line 85: I think you forgot something, as you only check for the winner at the end of the you lose the information of who one first. Meaning that if both players get a complete grid, you will declare winner the first YOU notice has terminated a line.
You should stop it before !

clafoutis42 (commented):
Your code seems to work, but there is some little edge cases you did not think about, better luck next time!

PR #255, opened 2020-05-17
Chall10

victorsempere:
Hi all!

First time I did the workflow as it must be done :dancers: 

Process:
1) I use two lists, one to set the players names. The other to write down the cells occupied . Writting 0 to say nobody occupies the cell, 1 means the player 1 is in the cell, 2 means the player 2 is in the cell.
2) At the first time something fails, None will be returned and Invalid input will be printed
3) If everything matches, the map becomes updated with the cell occupied by the move
4) The win condition is being checked when there is at least 7 movements and just checks from the cell currently occupied the horizontal , vertical and diagonals looking for 4 pieces of the same colour of the cell, in a row
5) None, Player 1 name or Player 2 name will be printed at this point

### DIFF
--- a/chall10/vsempere.py
+++ b/chall10/vsempere.py
@@ -1,0 +1,117 @@
+#! /usr/local/bin/python3
+
+import sys
+import re
+
+def check_east(i, map):
+    if (i < 0 or i >= 42 or (i % 7) > 3):
+        return 0
+    if (map[i] == 1 and (map[i:i + 4] == [1,1,1,1])):
+        return 1
+    elif (map[i] == 2 and (map[i:i + 4] == [2,2,2,2])):
+        return 2
+    return 0
+
+
+def check_west(i, map):
+    if (i < 0 or i >= 42 or (i % 7) < 3):
+        return 0
+    if (map[i] == 1 and (map[i-4:i] == [1,1,1,1])):
+        return 1
+    elif (map[i] == 2 and (map[i-4:i] == [2,2,2,2])):
+        return 2
+    return 0
+
+
+def check_north(i, map):
+    if (i < 0 or i >= 14):
+        return 0
+    if (map[i] > 0 and map[i + 7] == map[i] and map[i + 14] == map[i] 
+    and map[i + 21] == map[i]):
+        return map[i]
+    return 0
+
+def check_north_west(i, map):
+    if (i < 0 or i >= 14 or (i % 7) < 3):
+        return 0
+    if (map[i] > 0 and map[i + 6] == map[i] and map[i + 12] == map[i] 
+    and map[i + 18] == map[i]):
+        return map[i]
+    return 0
+
+def check_north_east(i, map):
+    if (i < 0 or i >= 14 or (i % 7) >= 2):
+        return 0
+    if (map[i] > 0 and map[i + 8] == map[i] and map[i + 16] == map[i] 
+    and map[i + 24] == map[i]):
+        return map[i]
+    return 0
+
+def check_winner(map):
+    i = 0
+    while (i < 42):
+        if (check_east(i, map) or check_west(i, map) or 
+        check_north(i, map) or check_north_west(i, map) or 
+        check_north_east(i, map)):
+            return map[i]
+        i = i + 1
+
+    return 0
+    
+def process_user_input(user_input):
+    if (not isinstance(user_input, list)):
+        return None, None;
+
+    plyrs = [None, None]
+    cols = [0, 0, 0, 0, 0, 0, 0]
+    turn = 0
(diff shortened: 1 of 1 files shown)


chall10/vsempere.py
```diff
@@ -0,0 +1,163 @@
+#! /usr/local/bin/python3
```
aguimbao, line 1: It's better, and more modular to use `#!/usr/bin/env python3`. For example, in my environment, I don't have any interpreter located in `/usr/local/bin/`, so your shebang won't work.

I will still correct your errors even if I can't run it.

chall10/vsempere.py
```diff
@@ -0,0 +1,163 @@
...
+    return 0
+    
+def process_user_input(user_input):
+    if (not isinstance(user_input, list) or len(user_input) > 42):
+        return None;
+
+    plyrs = [None, None]
+    cols = [0, 0, 0, 0, 0, 0, 0]
+    turn = 0
+    map = [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
```
aguimbao, line 111: I think it would have been easier to to this as a list of lists. Like `[[0,0,0,0,0,0,0],[0,0,0,0,0,0,0],[0,0,0,0,0,0,0],[0,0,0,0,0,0,0],[0,0,0,0,0,0,0],[0,0,0,0,0,0,0]]`.

You could achieve this easily with: `[[0 for _ in range(7)] for _ in range(6)]` (taken from edramire ;) ).

chall10/vsempere.py
```diff
@@ -0,0 +1,163 @@
...
+
+def main():
+    user_input = sys.stdin.readlines()
+
+    winner = process_user_input(user_input)
+    if (winner is None):
+        print("Invalid input.")
+    else:
+        print(winner + ".")
+    sys.exit(0)
```
aguimbao, line 160: You don't really need to do this in `python` like you would in `C`.

chall10/vsempere.py
```diff
@@ -0,0 +1,163 @@
...
+        if (turn == 0 and plyrs[0] != colour):
+            return None
+        elif (turn == 1 and plyrs[1] != colour):
+            return None
+        turn = (turn + 1) % 2
+        movs = movs + 1
+    return "Draw"
+
+def main():
+    user_input = sys.stdin.readlines()
```
aguimbao, line 153: Nice way to read the lines, but it maybe would have been better to read each line one by one, to check for invalid ones and exit directly (I solved the problem like you, but it's just a thought).

chall10/vsempere.py
```diff
@@ -0,0 +1,163 @@
...
+
+    winner = process_user_input(user_input)
+    if (winner is None):
```
aguimbao, line 156: Simple `if` statements don't need the parentheses, and by convention should be used without, like `if winner is None:`

chall10/vsempere.py
```diff
@@ -0,0 +1,163 @@
...
+        print("Invalid input.")
+    else:
+        print(winner + ".")
```
aguimbao, line 159: A really cool thing about python > 3.5 is `f-strings` functionality. This could be written as `print(f"{winner}.")` .

chall10/vsempere.py
```diff
@@ -0,0 +1,163 @@
...
+    
+def process_user_input(user_input):
+    if (not isinstance(user_input, list) or len(user_input) > 42):
+        return None;
+
+    plyrs = [None, None]
+    cols = [0, 0, 0, 0, 0, 0, 0]
+    turn = 0
+    map = [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
+    xpr_reg = re.compile(r"(\S+)_(\S+)")
```
aguimbao, line 112: Not a bad regex, but you would still match non printable characters, maybe `^[A-G]_[A-Z][a-zA-Z0-9]+$` would have been better? It respects all the examples in the subject.

chall10/vsempere.py
```diff
@@ -0,0 +1,163 @@
...
+    movs = 0
+    for line in user_input:
```
aguimbao, line 114: You could have had all valid matches with:

```python3
list_of_matches = [xpr_reg.match(line) for line in user_input if xpr_reg.match(line) is not None]
```

aguimbao (commented):
Overall it's a good solution that works in most cases, the problem is that it's not `pythonic` at all, which is normal when you're used to an old-school language like `C`. You have a lot of `while` loops, a lot of `if` statements, a lot of `indexes`.

You should instead research about how to take advantage of all those juicy builtin functionality and useful keywords that python provides you with !

PR #256, opened 2020-05-17
Chall10 - pcuadrad

pabloocg:
First I store all the user input and analyze if it has any error. If the first letter is between A-G and after the '_' it is capitalized and there are no more than two that play.
Then I create a 6 * 7 list to simulate the game board. I initialize it with '0'. One player fills in with '1' and the other with '2'.
In the game function, I fill the board with each movement in the list and then check in check_win () function, if the last movement that has been made means that there are more than 4 in a row, both horizontal, vertical or diagonal.
If this function returns True it means that there has been a winner and it will paint the player who made the last move.
If there is no winner Draw will be painted.


chall10/pcuadrad.py
```diff
@@ -0,0 +1,108 @@
...
+    for elem, n in zip(lstInput, range(len(lstInput))):
+        turnMove = '1' if not n % 2 else '2'
+        move = str(elem).split('_')
+        col = lettersValid.find(move[0])
+        row = retRow(table, col)
+        if row == -1:
+            continue
+        table[row][col] = turnMove
+        if check_win(table, row, col, turnMove):
+            print(players.get(turnMove)+'.')
```
aguimbao, line 84: At first when I read your comment I thought, well, checking who was the last player is smart. Then I realised that we are feeding your script with a random set of instructions, which may not end when a player wins. This means that there can be moves, after a player wins. So your answer can be wrong.

chall10/pcuadrad.py
```diff
@@ -0,0 +1,108 @@
...
+            return True
+    return False
+
+def main():
+    '''Save the input and remove '\\n' from the end.\n
+        @inputlst -> list containing user input.\n
+        @table -> board where the boxes will be filled in each turn.'''
+
+    inputlst = []
+    for line in sys.stdin:
```
aguimbao, line 94: You can get all the lines with `sys.stdin.readlines()` or `sys.stdin.read()`.

chall10/pcuadrad.py
```diff
@@ -0,0 +1,108 @@
...
+        line = line.rstrip('\n')
+        if len(line) > 0:
+            inputlst.append(line)
+    if not checker(inputlst):
+        print("Invalid input.")
+    else:
+        table = []
+        for _ in range(6):
+            table.append(['0'] * 7)
```
aguimbao, line 103: Little trick: `[['0' for _ in range(7)] for _ in range(6)]`

chall10/pcuadrad.py
```diff
@@ -0,0 +1,108 @@
...
+import sys
+
+lettersValid = "ABCDEFG"
+
+def checker(lstInput: list) -> bool:
+    '''Check the input user That the box is between A-G,
+        that the second parameter after '_' is capitalized and 
+        there are only two and they alternate '''
+
+    if len(lstInput) > 42:
```
aguimbao, line 11: A lot of this could have been much easier to read with a nice regex !

chall10/pcuadrad.py
```diff
@@ -0,0 +1,108 @@
...
+
+def game(lstInput: list, table: list) -> bool:
+    '''Function that fills the board according to the instructions that the user has entered.\n
+        lstInput -> list of the moves.
+        table -> table of the game 6*7.'''
+
+    players = {'1': str(lstInput[0]).split('_')[1], '2': str(lstInput[1]).split('_')[1]}
+    for elem, n in zip(lstInput, range(len(lstInput))):
+        turnMove = '1' if not n % 2 else '2'
+        move = str(elem).split('_')
```
aguimbao, line 77: Also would been easier with a regex !

chall10/pcuadrad.py
```diff
@@ -0,0 +1,108 @@
...
+        if newcolor[0] not in lettersValid or not newcolor[1].istitle():
+            return False
+        if i % 2 == 0:
+            if i > 0 and (fcolor[1] != newcolor[1] or newcolor[1] == scolor[1]):
+                return False
+            fcolor = newcolor
+        else:
+            if i > 1 and (scolor[1] != newcolor[1] or newcolor[1] == fcolor[1]):
+                return False
+            scolor = newcolor
```
aguimbao, line 26: Assigned but never used?

chall10/pcuadrad.py
```diff
@@ -0,0 +1,108 @@
...
+lettersValid = "ABCDEFG"
+
+def checker(lstInput: list) -> bool:
+    '''Check the input user That the box is between A-G,
+        that the second parameter after '_' is capitalized and 
+        there are only two and they alternate '''
+
+    if len(lstInput) > 42:
+        return False
+    for elem, i in zip(lstInput, range(len(lstInput))):
```
aguimbao, line 13: I think `for elem, i in enumerate(lstInput):` is more performant.

aguimbao (commented):
Hey, your answer is good, you forgot a couple checks, and your way to declare winner wasn't that much of a good idea. 

You have nice practices, like static typing, good docstrings, good shebang and permissions, etc. But it still looks a lot like C. Even if it looks like it's not possible, you can usually work around using `indexes` and `while` loops in python in most situations if you know all the good `builtins` and `keywords`.

chall10/pcuadrad.py
pabloocg: But in the event that there are more movements after a player wins, they would be ignored since someone has already won ... I don't know if I understood you correctly.

chall10/pcuadrad.py
pabloocg: Learning to make good regex is a pending task that I have... I always forgot!

chall10/pcuadrad.py
pabloocg: I use it in the conditions, comparing it with the previous fcolor or scolor simultaneously to check that they are not repeated.

chall10/pcuadrad.py
aguimbao: Edit: Please don't read that, I was thinking about `gbudau`'s program hahaha.

The problem with your program is that it doesn't ignore the moves after a player has already won. I'll give you an example.

With this test:

```txt
A_Yellow
B_Red
A_Yellow
B_Red
A_Yellow
B_Red
A_Yellow
B_Red
```

You should return `"Yellow."` as he was the first person to win, but there's a problem in your logic, and you return `"Red."` because you still check the last move and you return the last person who moved.

aguimbao: Edited with explanation.

PR #257, opened 2020-05-18
chall10 - gbudau

gbudau:
**How i solved the problem:**

First i create an empty 7x7 board and initialize it to 0, i'm using the first row to store the total pieces in each column.
Also i create a list containing the two teams using the first two lines from the input.
Validate each line from the input and return "Invalid input." if its not valid.
For each input add the piece on the board using the team identifier and check if it can make a 4 piece line
If we find 4 pieces in a row on horizontal, vertical or diagonal print the team and exit.
If no team won, print "Draw."

**Short description of my code:**
`init_board()` initiate a 7x7 board.
Then i create a `teams` list that will contain the two teams from the first two lines of the input (if they are valid and don't coincide)
Then for each line in the input, first validate and check if its the right order of turns, then use the first character as index of columns and aument the pieces for that  column and verify to not be more than 6 (max rows)
Then in `connect_four` first place the piece on the board as 1 for first team of 2 for second team.
In `check_connections`  i look on horizontal, vertical and diagonals for 4 pieces of the  same team in a line.
If there are 4 pieces connected in a line print the team and exit.
If we reach the end of file print `Draw.`

> NOTE: I considered a valid input with the format atleast `COLUMN(A to G)_UPPERLETTER` that could be followed by any characters or none at all, so teams like `A` , `A1` or `A+` etc.. are valid in my program.

### DIFF
--- a/chall10/gbudau.py
+++ b/chall10/gbudau.py
@@ -1,0 +1,141 @@
+#!/usr/bin/env python3
+
+import fileinput
+import sys
+import curses.ascii
+
+POSITION = "ABCDEFG"
+TEAMS = [1, 2]
+NOT_CONNECTED = 3
+CONNECTED = 4
+ROWS = 6
+COLUMNS = 7
+
+def init_board():
+    board = [[0] * COLUMNS for i in range(COLUMNS)]
+    return(board)
+
+def valid_input(action):
+    if not action or len(action) < 3 or action[0] not in POSITION or \
+       action[1] not in "_" or not curses.ascii.isupper(action[2]):
+        return False
+    return True
+
+def check_horizontal(board, team, row, col):
+    connect = 0
+    for i in board[row][col:]:
+        if i == TEAMS[team]:
+            connect += 1
+        else:
+            break
+    for i in board[row][:col][::-1]:
+        if i == TEAMS[team]:
+            connect += 1
+        else:
+            break
+    return team if connect >= CONNECTED else NOT_CONNECTED
+
+def check_vertical(board, team, row, col):
+    connect = 0
+    for i in range(1, row)[::-1]:
+        if board[i][col] == TEAMS[team]:
+            connect += 1
+        else:
+            break
+    for i in range(row, ROWS + 1):
+        if board[i][col] == TEAMS[team]:
+            connect += 1
+        else:
+            break
+    return team if connect >= CONNECTED else NOT_CONNECTED
+
+def check_diagonal_nesw(board, team, row, col):
+    connect = 0
+    j = col
+    for i in range(1, row)[::-1]:
+        j += 1
+        if j < COLUMNS and board[i][j] == TEAMS[team]:
+            connect += 1
+        else:
+            break
+    j = col
+    for i in range(row, ROWS + 1):
+        if j >= 0 and board[i][j] == TEAMS[team]:
+            connect += 1
+            j -= 1
+        else:
+            break
+    return team if connect >= CONNECTED else NOT_CONNECTED
+
+def check_diagonal_nwse(board, team, row, col):
+    connect = 0
+    j = col
+    for i in range(1, row)[::-1]:
+        j -= 1
+        if j >= 0 and board[i][j] == TEAMS[team]:
+            connect += 1
+        else:
+            break
+    j = col
+    for i in range(row, ROWS + 1):
+        if j < COLUMNS and board[i][j] == TEAMS[team]:
+            connect += 1
+            j += 1
+        else:
+            break
+    return team if connect >= CONNECTED else NOT_CONNECTED
+
+def check_diagonals(board, team, row, col):
+    connected = check_diagonal_nesw(board, team, row, col)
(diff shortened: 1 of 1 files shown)


chall10/gbudau.py
```diff
@@ -0,0 +1,141 @@
...
+    for turn, line in enumerate(fileinput.input()):
+        action = line.rstrip("\n")
+        if not valid_input(action):
+            sys.exit("Invalid input.")
+        if not turn:
+            teams.append(action[2:])
+        elif turn == 1 and teams[0] != action[2:]:
+            teams.append(action[2:])
+        elif teams[prev] == action[2:]:
+                sys.exit("Invalid input.")
```
gbudau, line 129: I forgot to add a condition to check if the team on turns 3+ is already in teams, so instead of printing `Invalid  input.`  when there are more than 2 teams it will evaluate other teams as one of the first two based on turn % 2.
```
elif teams[prev] == action[2:] or action[2:] not in teams:
                sys.exit("Invalid input.")
```

chall10/gbudau.py
```diff
@@ -0,0 +1,141 @@
...
+ROWS = 6
+COLUMNS = 7
+
+def init_board():
+    board = [[0] * COLUMNS for i in range(COLUMNS)]
+    return(board)
+
+def valid_input(action):
+    if not action or len(action) < 3 or action[0] not in POSITION or \
+       action[1] not in "_" or not curses.ascii.isupper(action[2]):
```
aguimbao, line 20: I think this was a good opportunity to use a good regex, for example: `"(^[A-G])_([A-Z][a-zA-Z0-9]+$)"`. They are pretty useful in languages like python.

chall10/gbudau.py
```diff
@@ -0,0 +1,141 @@
...
+def connect_four(board, team, position):
+    for i, row in enumerate(board[:ROWS - 1], 1):
+        if board[i + 1][position]:
+            board[i][position] = TEAMS[team]
+            return check_connections(board, team, i, position)
+    board[i + 1][position] = TEAMS[team]
+    return check_connections(board, team, i + 1, position)
+
+def main():
+    board = init_board()
```
aguimbao, line 118: I think the `init_board()` function is a bit redundant, you could have defined `board` globally as `BOARD = [[0 for _ in range(7)] for _ in range(6)]` (from `edramire`) 😄

chall10/gbudau.py
```diff
@@ -0,0 +1,141 @@
...
+def check_diagonals(board, team, row, col):
+    connected = check_diagonal_nesw(board, team, row, col)
+    if connected != NOT_CONNECTED:
+        return team
+    connected = check_diagonal_nwse(board, team, row, col)
+    if connected != NOT_CONNECTED:
+        return team
+    return NOT_CONNECTED
+
+def check_connections(board, team, row, col):
```
aguimbao, line 97: I think there's a lot of overhead with this way of approaching the problem, you try to `return` the answer to your `main()` function, when you could have simply `sys.exit(f"{winner}.")` in the `check_horizontal()`, check_vertical()` or check_diagonals()` functions.

Also I don't really like the `NOT_CONNECTED` variable, it's a bit confusing.

chall10/gbudau.py
```diff
@@ -0,0 +1,141 @@
...
+import curses.ascii
+
+POSITION = "ABCDEFG"
+TEAMS = [1, 2]
+NOT_CONNECTED = 3
+CONNECTED = 4
+ROWS = 6
+COLUMNS = 7
+
+def init_board():
```
aguimbao, line 14: Redundant (IMO).

aguimbao (commented):
Nice solution !

You solved it in a simple way that's easy to understand, there were a couple different ways to check for the winner, like using regexes, or pre-computing all the possible answers, but checking one by one is one of the simple ways to do it.

Maybe importing `fileinput` and `curses.ascii` is a bit overkill when you already have `sys` for the input and a good opportunity to try regexes. 
Your code still looks a lot like C, a lot of `break`'s, a lot of implicit loops, and `if` forests, I guess you guys want more `C` than `python` 😓 .

Too bad there was a little fail when you're checking the teams. :(

chall10/gbudau.py
gbudau: Thanks, will keep in mind for next time.

chall10/gbudau.py
gbudau: Agree, there was no need top create an additional function, also defining the board globally would have made updating it easier :slightly_smiling_face:

chall10/gbudau.py
gbudau: I prefered to return instead of exiting from those functions because it was easier for me to test them.
Could have used -1 instead of NOT_CONNECTED, it is a bit confusing :+1:

gbudau:
Thank you Abraham! Will try to use regex for similar cases in the future. Sorry if the code looks a lot like `C`, i think `Python` is amazing and i need more practice with both of them :heart:

PR #258, opened 2020-05-19
vsempere - chall09 - late 

victorsempere:
First of all. Checks (nothing different than numbers. No more than one number, no negative number. Number max allowed value...)

Split number per unit of time: years, days, hours, ...
Build each value and prepend the conjunction (", " or " and ")
Concat all values
Return result

### DIFF
--- a/chall09/vsempere.c
+++ b/chall09/vsempere.c
@@ -1,0 +1,160 @@
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <stdarg.h>
+#include <ctype.h>
+
+#define YEAR_UNIT       0
+#define DAY_UNIT        1
+#define HOUR_UNIT       2
+#define MINUTE_UNIT     3
+#define SECOND_UNIT     4
+#define YEAR_SECONDS    31536000
+#define DAY_SECONDS     86400
+#define HOUR_SECONDS    3600
+#define MINUTE_SECONDS  60
+#define MAX_SECONDS     10000000000000000000ul
+
+static void extract_unit(long long *num, int unit_seconds, long *units)
+{
+    *units = *num / unit_seconds;
+    *num = *num - (*units * unit_seconds);
+}
+
+static char *build_particle(long seconds, char *unit, char *use_separator,
+    int *remaining_particles)
+{
+    char    *str;
+
+    str = 0;
+    if (seconds <= 0 || !(str = malloc(sizeof(char) * 25)))
+        return (str);
+
+    if (*use_separator && *remaining_particles > 1)
+        sprintf(str, ", %ld %s", seconds, unit);
+    else if (*use_separator && *remaining_particles == 1)
+        sprintf(str, " and %ld %s", seconds, unit);
+    else
+        sprintf(str, "%ld %s", seconds, unit);
+
+    *use_separator = 1;
+    (*remaining_particles)--;
+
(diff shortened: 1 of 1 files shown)


chall09/vsempere.c
```diff
@@ -0,0 +1,160 @@
...
+    str = 0;
+    if (seconds <= 0 || !(str = malloc(sizeof(char) * 25)))
+        return (str);
+
+    if (*use_separator && *remaining_particles > 1)
+        sprintf(str, ", %ld %s", seconds, unit);
+    else if (*use_separator && *remaining_particles == 1)
+        sprintf(str, " and %ld %s", seconds, unit);
+    else
+        sprintf(str, "%ld %s", seconds, unit);
```
clafoutis42, line 38: There's a debate here, always allocating `25` will lead you to unnecessary memory, but allows you to only make one call.
Whilst using `asprintf` would allow you to allocate exactly what you need, with more system calls

chall09/vsempere.c
```diff
@@ -0,0 +1,160 @@
...
+    char    *years;
+    char    *days;
+    char    *hours;
+    char    *minutes;
+    char    *seconds;
+    char    use_separator;
+    
+    use_separator = 0;
+    years = build_particle(unit_values[YEAR_UNIT],
+        unit_values[YEAR_UNIT] > 1 ? "years" : "year", &use_separator, &particles);
```
clafoutis42, line 59: To improve readability, you should not just break the line, but organize it as follows:
```suggestion
    years = build_particle(
        unit_values[YEAR_UNIT],
        unit_values[YEAR_UNIT] > 1 ? "years" : "year",
        &use_separator,
        &particles
    );
```
Especially since you have long and complicated arguments passed to your functions

chall09/vsempere.c
```diff
@@ -0,0 +1,160 @@
...
+    else
+        sprintf(str, "%ld %s", seconds, unit);
+
+    *use_separator = 1;
+    (*remaining_particles)--;
+
+    return (str);
+}
+
+static char *build_human_time(long unit_values[5], int particles)
```
clafoutis42, line 46: You are using `long` to store your values, but years might exceed `LONG_MAX`. This will cause your program not to work :(

chall09/vsempere.c
```diff
@@ -0,0 +1,160 @@
...
+    if (minutes)
+        len += strlen(minutes);
+    if (seconds)
+        len += strlen(seconds);
+    human_time = calloc(len + 1, sizeof(char));
+    if (years) {
+        if (human_time) 
+            strcat(human_time, years);
+        free(years);
+    }
```
clafoutis42, line 85: Maybe you could have made a function for this repetitive tasks ?

chall09/vsempere.c
```diff
@@ -0,0 +1,160 @@
...
+        i++;
+    return (i && seconds[i] == '\0' ? 1 : 0);
+}
+
+static char *allocate_string(const char *str) {
+    char *astr;
+    
+    astr = calloc(strlen(str) + 1, sizeof(char));
+    return (strcat(astr, str));
+}
```
clafoutis42, line 130: This is basically strdup?

clafoutis42 (commented):
You have a good solution with nice ideas, but the type problem will cause your solution not to work, too bad

PR #259, opened 2020-05-19
vsempere - chall10 - upload file from an empty branch

victorsempere:
Process input until EOF
Then parse each line. If something is wrong an Invalid input is returned.
If not the movements are done one at a time. The check for win condition will occur after from the 7th move on when it could be the first time someone can win.
The check involves only the currently occupied cell and its surrounding cells (verticals, horizontal, diagonals)

### DIFF
--- a/chall10/vsempere.py
+++ b/chall10/vsempere.py
@@ -1,0 +1,163 @@
+#! /usr/local/bin/python3
+
+import sys
+import re
+
+def cell_matches_value(map, valueToMatch, row, col):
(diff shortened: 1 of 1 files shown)


aguimbao:
I merged your other PR (by accident), but next time it will be better in a different branch like this :)

victorsempere:
Thank you @abguimba . Next time I'll do it the right way. I'll have a look to the rebase and reset you mentioned.

PR #261, opened 2020-05-21
vde-dios.c

vdedios:
### SOLUTION

With `strtok` the string is split using `\n` as a delimiter. After that, each sentence is analysed. First number is checked, if we get an `errno` it means we have a goto sentence. In this case we look for the matching line using `strstr` to encounter the following line.

For the output string, `asprintf` function is used. Each sentence is split with `strtok` and properly contatenated.

The function exits with no leaks apart from the returned string. All output strings are allocated and there is no string literals involved.

### DIFF
--- a/chall11/vde-dios.c
+++ b/chall11/vde-dios.c
@@ -14,7 +14,7 @@ char *ft_goto_parser(const char *code)
     long n_line = 0;
     char *input = strdup(code);
 
-    next = strtok(input, "\n");
+    next = strtok_r(input, "\n", &input);
     while (next)
     {
         if (!strtol(next, &statement, 10) && errno == EINVAL)
@@ -25,15 +25,15 @@ char *ft_goto_parser(const char *code)
                 return (strdup("Infinite loop !"));
             else
             {
-                next = strtok(NULL, "\n");
+                next = strtok_r(input, "\n", &input);
                 while(!strstr(next, statement))
-                    next = strtok(NULL, "\n");
+                    next = strtok_r(input, "\n", &input);
             }
         }
         else
         {
             n_line = strtol(next, &statement, 10);
-            word = strtok(statement, " ");
+            word = strtok_r(statement, " ", &statement);
             while(word)
             {
                 while(isspace(*word))
@@ -41,9 +41,9 @@ char *ft_goto_parser(const char *code)
                 asprintf(&output, "%s%s%s", tmp ? tmp: "", tmp ? " ": "", word);
                 free(tmp);
                 tmp = output;
-                word = strtok(NULL, "\n");
+                word = strtok_r(statement, " ", &statement);
             }
-            next = strtok(NULL, "\n");
+        next = strtok_r(input, "\n", &input);
         }
     }
     free(input);
--- a/chall11/vde-dios.c
+++ b/chall11/vde-dios.c
@@ -8,7 +8,7 @@ char *ft_goto_parser(const char *code)
 {
     char *next;
     char *word;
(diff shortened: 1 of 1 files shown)


chall11/vde-dios.c
```diff
@@ -0,0 +1,54 @@
...
+                    asprintf(&output, "%s%s%s", tmp ? tmp: "", tmp ? " ": "", word);
+                    free(tmp);
+                    tmp = output;
+                }
+                word = strtok_r(next, " ", &next);
+            }
+            next = strtok_r(input, "\n", &input);
+        }
+    }
+    free(input);
```
aguimbao, line 52: Be careful because this works in macOs, but will crash in any normal environment that runs under linux. As glibc's `strtok_r` modifies your pointer (`input`) and returns you a pointer to the `'\0'` char.

I will comment this line out for the tests as it's not fair.

Also we don't really check leaks as it's a bit hard when correcting you guys, but good practice :)

chall11/vde-dios.c
```diff
@@ -0,0 +1,54 @@
...
+            while (isdigit(*next))
+                next++;
+            word = strtok_r(next, " ", &next);
+            while(word)
+            {
+                while(*word && isspace(*word))
+                    word++;
+                if (*word){
+                    asprintf(&output, "%s%s%s", tmp ? tmp: "", tmp ? " ": "", word);
+                    free(tmp);
```
aguimbao, line 44: We don't really check leaks as it's a bit hard when correcting you guys, but good practice :)

chall11/vde-dios.c
```diff
@@ -0,0 +1,54 @@
...
+        {
+            aux = next;
+            while(!isdigit(*aux))
+                aux++;
+            next = strtok_r(input, "\n", &input);
+            while(!strstr(next, aux))
+            {
+                next = strtok_r(input, "\n", &input);
+                if (!next)
+                    return (strdup("Infinite loop !"));
```
aguimbao, line 30: Not the best way to detect infinite loops.

For example you detect infinite loops if there's a line like this: `"goto  10  \n"` with more than one space after the line number.

aguimbao (commented):
Overall I think your approach is really nice. `strtok()` is a really nice function and I think you used it really well 👍 . You mention `errno`, but you never really use it in your implementation...

Also, you didn't account for "false" infinite loops. For example if the string is:

```C
const char  *code = 
"5 Hello\n"
"goto         50\n"
"10 my\n"
"20 name is Marvin\n"
"goto         80\n"
"30 AND\n"
"40 I love 42\n"
"50 Hello2\n"
"goto 20\n"
"60 my2\n"
"70 name is Marvin2\n"
"80 AND2\n"
"90 I love 422\n";
```

Here we make a backwards jump, but there's not an infinite loop !

vdedios:
About the errno my fault.. In last commits I altered some part of the code removing that part and forgot updating the description...

chall11/vde-dios.c
vdedios: Ooops.. in this challenge I just tried the examples you provided (and similar ones) worked. Never though that was asked but makes sense  🤷‍♂️

chall11/vde-dios.c
vdedios: I considered it at first (though about using an auxiliar pointer), but after making some tests and check it did not crash I left it that way. But it is nice to know, I hate those slight differences across different OS 😓

aguimbao:
I submitted your code manually cause it didn't merge because you grouped your PRs for different challenges.

PR #263, opened 2020-05-21
Chall11 migferna

migferna42:
We divide the initial code into several sentences separated by \ n, this I do with strtok.
Each of these phrases, I put it inside an Arrangement of instructions, to be able to detect if any one is repeated, for this part I use the function strstr. If it is repeated, we return that it is a loop.
In the opposite case, the instructions are formatted, eliminating digits and spaces.
When a goto instruction arrives, we look it up in the entire code block, if the destination instruction is above the goto instruction, we have the localized instruction, otherwise we will have to do another search.
All the words I concatenate with a space, and later I put them to the output string.


chall11/migferna.c
```diff
@@ -0,0 +1,83 @@
...
+	
+	if (!(parser = malloc(sizeof(parser) * strlen(code))))
+		return (NULL);
+	if (!(list = malloc(sizeof(list) * strlen(code))))
+		return (NULL);
+	str = strdup(code);
+	ptr = strtok(str, "\n");
+	while(ptr != NULL)
+	{
+		if (strstr(list, ptr) == NULL)
```
aguimbao, line 47:
```txt
"The strstr() function locates the first occurrence of the null-terminated string needle in the null-terminated string haystack."
```

When you get to this line, `list` has just been malloced, there's no null-termination inside of it. You get a segfault here with the example tests even when you compile without flags 😞

chall11/migferna.c
```diff
@@ -0,0 +1,83 @@
...
+			word = clean_word(ptr);
+			ptr = strtok(NULL, "\n");
+			if (ptr != NULL)
+				strlcat(word, " ", sizeof(word));
+			strlcat(parser, word, strlen(word) * sizeof(parser));
+		}
+		else
+			return strdup("Infinite loop !");
+	}
+	free(str);
```
aguimbao, line 80: You don't need to worry about leaks for the challenges :)

chall11/migferna.c
```diff
@@ -0,0 +1,83 @@
...
+	free(list);
```
aguimbao, line 81: You don't need to worry about leaks for the challenges :)

chall11/migferna.c
```diff
@@ -0,0 +1,83 @@
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <ctype.h>
+
+char *clean_word(const char *code)
```
aguimbao, line 6: `static` 👍

chall11/migferna.c
```diff
@@ -0,0 +1,83 @@
...
+{
+	char *clean;
+	char *str;
+	char *ptr;
+	char *word;
+
+	if (!(clean = malloc(sizeof(clean) * strlen(code))))
```
aguimbao, line 13: I think you are missing the last byte here for the null-terminator `'\0'` character if I'm not mistaken?

chall11/migferna.c
```diff
@@ -0,0 +1,83 @@
...
+{
+	char *parser;
+	char *word;
+	char *str;
+	char *ptr;
+	char *subcode;
+	char *list;
+	char *tmp;
+	
+	if (!(parser = malloc(sizeof(parser) * strlen(code))))
```
aguimbao, line 39: Same as for line 13.

chall11/migferna.c
```diff
@@ -0,0 +1,83 @@
...
+		return (NULL);
+	if (!(list = malloc(sizeof(list) * strlen(code))))
```
aguimbao, line 41: Same as for line 13.

chall11/migferna.c
```diff
@@ -0,0 +1,83 @@
...
+		return (NULL);
+	str = strdup(code);
+	ptr = strtok_r(str, " ", &str);
+	while (ptr != NULL)
+	{
+		word = strdup(ptr);
+		
+		ptr = strtok_r(str, " ", &str);
+		if (ptr != NULL)
+			strlcat(word, " ", sizeof(word));
```
aguimbao, line 23: Don't really worry about using `sizeof` in `char`'s. They are always 1 byte.

chall11/migferna.c
```diff
@@ -0,0 +1,83 @@
...
+			while (isspace(*ptr))
+				ptr++;
+			word = clean_word(ptr);
+			ptr = strtok(NULL, "\n");
+			if (ptr != NULL)
+				strlcat(word, " ", sizeof(word));
+			strlcat(parser, word, strlen(word) * sizeof(parser));
+		}
+		else
+			return strdup("Infinite loop !");
```
aguimbao, line 78: You didn't take into account "fake" infinite loops. For example:
```C
const char  *code = 
"5 Hello\n"
"goto         50\n"
"10 my\n"
"20 name is Marvin\n"
"goto         80\n"
"30 AND\n"
"40 I love 42\n"
"50 Hello2\n"
"goto 20\n"
"60 my2\n"
"70 name is Marvin2\n"
"80 AND2\n"
"90 I love 422\n";
```

Here we make a backwards jump, but there's not an infinite loop !

aguimbao (commented):
I couldn't really test your submission with  a segfault so early on in the tests...

PR #264, opened 2020-05-21
chall11 | estina

xplore93:
My solution for this problem.
First, I make a strsep by "\n".
Then I check if first char isdigit() and if so, make asprintf() into my result string. If the first char is not a digit, I assume that it's "goto" and pass through all next strings searching the label. If so, make asprintf() into my result string. If not, throw and "Infinite Loop!".


chall11/estina.c
```diff
@@ -0,0 +1,80 @@
...
+	char	*duplicate = strdup(code);
+	char	*token = strsep(&duplicate, "\n");
+
+	while (token)
+	{
+		if (isdigit(token[0]))
+			put_line(&response, token);
+		else
+		{
+			label = atoi(token + 4);
```
aguimbao, line 52: In your last loop, you call `atoi()` here with an empty string, this causes a buffer-overflow (you can, and you should always) verify this with `gcc`'s `-g` and `-fsanitize=address` options. It's a shame cause you submitted a really nice solution.

chall11/estina.c
```diff
@@ -0,0 +1,80 @@
...
+			}
+		}
+		line = strsep(&token, " ");
+	}
+}
+
+static char	*return_infinite(char **duplicate, char **response)
+{
+	if (*response)
+		free(*response);
```
aguimbao, line 34: Don't worry about leaks with the challenges, but good practice :)

chall11/estina.c
```diff
@@ -0,0 +1,80 @@
...
+			label = atoi(token + 4);
+			if (loop == -1)
+				loop = label;
+			else if (loop == label)
+				return (return_infinite(&duplicate, &response));
+			token = strsep(&duplicate, "\n");
+			while (token)
+			{
+				if (isdigit(token[0]))
+					if (atoi(token) == label)
```
aguimbao, line 61: Maybe `atoi()` was a bit unnecessary ?

aguimbao (commented):
I don't have much to say about your solution, it's very optimal and you have really nice practices as usual !

It's a shame you didn't sanitise addresses with `fsanitize` :( Cause you pass all the tests otherwise. But your extra `atoi()` makes you overflow on every test.

chall11/estina.c
xplore93: Shame on me

PR #265, opened 2020-05-21
dalba-de solution

dalba-de:
# Saffar Solution

I go through the string and while I can't find the `goto` command, I eliminate the numbers and spaces and fill in the string that I will return later. Within that same function I manage that there is only one space between each word.

When I find the `goto` command, I first extract the label and then look it up in the rest of the text with the `strstr` function. If I can't find it in the rest of the string, it means that that this label was previously in the text and therefore I assume it's an infinite loop.

I am aware that maybe in some inputs with a lot of goto, it might not work correctly, but at least I wanted to provide a solution for the simplest cases.


chall11/dalba-de.c
```diff
@@ -0,0 +1,73 @@
+#include <stdio.h>
+#include <ctype.h>
+#include <stdlib.h>
+#include <string.h>
+
+int		ft_malloc_str(const char *code)
```
aguimbao, line 6: You don't really need this function. You can find the index of the next occurrence of the `'\n'` in the string with this:

```C
char *str = "Hello\nWorld";
char *dest = strchr(str, '\n');
int index_of_next_newline = dest - str;
```

chall11/dalba-de.c
```diff
@@ -0,0 +1,73 @@
...
+	char	*ret;
+	int		j;
+	int		k;
+
+	str = malloc(strlen(code));
+	num = NULL;
+	j = 0;
+	while (*code)
+	{
+		if (num == NULL)
```
aguimbao, line 29: For me
```C
if (!num) {
 // code
}
```
Is more readable

chall11/dalba-de.c
```diff
@@ -0,0 +1,73 @@
...
+				num = malloc(ft_malloc_str(code) + 2);
+				k = 0;
+				while (*code >= '0' && *code <= '9')
+				{
+					num[k++] = *code++;
+				}
+				num[k++] = ' ';
+				num[k] = '\0';
+			}
+			while (*code >= '0' && *code <= '9')
```
aguimbao, line 44: You can use 
```C
isdigit()
```

For this 😉

chall11/dalba-de.c
```diff
@@ -0,0 +1,73 @@
...
+
+char    *ft_goto_parser(const char *code)
+{
+	char	*str;
+	char	*num;
+	char	*ret;
+	int		j;
+	int		k;
+
+	str = malloc(strlen(code));
```
aguimbao, line 24: Verify your `malloc()`'s, even in challenges, it's good practice :D

chall11/dalba-de.c
```diff
@@ -0,0 +1,73 @@
...
+	num = NULL;
+	j = 0;
+	while (*code)
+	{
+		if (num == NULL)
+		{
+			if (*code == 'g')
+			{
+				while (*code < '0' || *code > '9')
```
aguimbao, line 33: You can use
```C
isdigit()
```

For this 😉

chall11/dalba-de.c
```diff
@@ -0,0 +1,73 @@
...
+					code++;
+				num = malloc(ft_malloc_str(code) + 2);
+				k = 0;
+				while (*code >= '0' && *code <= '9')
```
aguimbao, line 37: You can use
```C
isdigit()
```

For this 😉

chall11/dalba-de.c
```diff
@@ -0,0 +1,73 @@
...
+				{
+					num[k++] = *code++;
+				}
+				num[k++] = ' ';
+				num[k] = '\0';
+			}
+			while (*code >= '0' && *code <= '9')
+				code++;
+			while (*code == ' ')
```
aguimbao, line 46: You can use
```C
isblank()
```

or

```C
isspace()
```

For this 😉

chall11/dalba-de.c
```diff
@@ -0,0 +1,73 @@
...
+		else
+		{
+			ret = strstr(code, num);
+			code = ret;
+			if (code == NULL)
+				return ("Infinite loop !");
+			free (num);
+			num = NULL;
+		}
+	}
```
aguimbao, line 70: You forgot to null-terminate (`'\0'`) the string that you return (I think). And it overflows when trying to print it. 

You can verify all of this with `fsanitize`'s gcc option when compiling the binary. Sadly this will make it fail the tests :(

chall11/dalba-de.c
```diff
@@ -0,0 +1,73 @@
...
+			if (num == NULL && str[j - 1] != ' ' && str[j] != '\0')
+				str[j++] = ' ';
+			code++;
+		}
+		else
+		{
+			ret = strstr(code, num);
+			code = ret;
+			if (code == NULL)
+				return ("Infinite loop !");
```
aguimbao, line 66: Don't return a mix of string literals (like in this line), and dynamically allocated memory. Because you don't know what we will do with the returned variable. If we tried to free it. Our program would crash in the case that you return string literals like here.

Much better to return 
```C
strdup("Infinite loop !");
```

chall11/dalba-de.c
```diff
@@ -0,0 +1,73 @@
...
+			free (num);
```
aguimbao, line 67: Good practice, but we don't verify leaks in the challenges.

aguimbao (commented):
Good submission !

Short, understandable and concise, if it wasn't for the overflow, you would pass almost all the tests. Except one edge case:

Basically you didn't account for "false" infinite loops. For example if the string is:

const char  *code = 
"5 Hello\n"
"goto         50\n"
"10 my\n"
"20 name is Marvin\n"
"goto         80\n"
"30 AND\n"
"40 I love 42\n"
"50 Hello2\n"
"goto 20\n"
"60 my2\n"
"70 name is Marvin2\n"
"80 AND2\n"
"90 I love 422\n";

Here we make a backwards jump, but there's not an infinite loop ! You return: `"Infinite loop !"` but you should have returned: `"HERE HERE2 JUST rAnDoM TEXT AND2 MORE RANDOM text2"`

PR #266, opened 2020-05-21
Chall11 - gsanchez

GonzSanch:
First at all I parse the char code into an array parsed by newlines then in an iteration I just check if the first character is a number, thats mean that is a text to be added to the result to do that I parse again that text with strtok_r as parsed the original char code. When i parsed this again in this proccess spaces are removed and words are isolated later I use asprintf to add the text, checking first it is empty(first text added) and then for the rest of the text adding a space between words. At the end if the first character is not a number, that's mean that is a goto statement, for this I check the label with atoi and in a while loop if the label from goto its equal replace the value for iterator if the check iteration its complete just return "Infinite loop !" then continues untill return result.


GonzSanch:
you merged this without review ;(

clafoutis42:
It is because you pushed both PR on the same branch... Wait a little we will review it :)

chall11/gsanchez.c
```diff
@@ -0,0 +1,77 @@
...
+            asprintf(&result, "%s %s", result, token);
+            free(aux);
+          }
+        }
+      }
+      else if (strstr(arr[j], "goto"))
+      {
+        indextogo = atoi(arr[j]+4);
+        j = 0;
+        while (++j < i)
```
aguimbao, line 65: This loop in one line
```C
while (++j < i && indextogo != atoi(arr[j]));
```

Cause I hate `break` statements.

chall11/gsanchez.c
```diff
@@ -0,0 +1,77 @@
+/* ************************************************************************** */
```
aguimbao, line 1: Header not necessary 😶

chall11/gsanchez.c
```diff
@@ -0,0 +1,77 @@
...
+	rest = buf;
+    for (i=0; buf[i]; buf[i]=='\n' ? i++ : *buf++);
+    arr = ft_parse_nlarray(rest, i);
+    result = NULL;
+    j = 0;
+    while (j < i)
+    {
+      if (isdigit(*arr[j]))
+      {
+        while (isdigit(*arr[j]))
```
aguimbao, line 46:
```C
while (isdigit(*arr[j]++));
```
😝

chall11/gsanchez.c
```diff
@@ -0,0 +1,77 @@
...
+        rest = arr[j];
+        while (token = strtok_r(rest, " ", &rest))
+        {
+          aux = result;
+          if (!aux)
+            asprintf(&result, "%s", token);
+          else
+          {
+            asprintf(&result, "%s %s", result, token);
+            free(aux);
```
aguimbao, line 57: We don't usually check for leaks in the challenges, but good practice :)

chall11/gsanchez.c
```diff
@@ -0,0 +1,77 @@
...
+#include <string.h>
+#include <stdlib.h>
+#include <ctype.h>
+
+char  **ft_parse_nlarray(char *src, int newsize)
+{
+  char  **arr;
+  int   j;
+
+  arr = malloc(sizeof(char*)*newsize);
```
aguimbao, line 23: You should check if malloc fails 😇

chall11/gsanchez.c
```diff
@@ -0,0 +1,77 @@
...
+  while (j != newsize)
+      arr[j++] = strtok_r(src, "\n", &src);
+  return (arr);
+}
+
+char  *ft_goto_parser(const char *code)
+{
+    char *buf, *result, *aux, *rest, *token;
+    char **arr;
+    int i, j, indextogo;
```
aguimbao, line 34: You use snake case for your functions so I would use snake case for `indextogo` too. So -> `index_to_go`, more readable IMO.

chall11/gsanchez.c
```diff
@@ -0,0 +1,77 @@
...
+  arr = malloc(sizeof(char*)*newsize);
+  j = 0;
+  while (j != newsize)
+      arr[j++] = strtok_r(src, "\n", &src);
+  return (arr);
+}
+
+char  *ft_goto_parser(const char *code)
+{
+    char *buf, *result, *aux, *rest, *token;
```
aguimbao, line 32: Honestly, this is a bit nitpicky, but I would assign all my pointers to `NULL`, as to explicitly declare they are not pointing to anything yet. So:

```C
char *buf, *result, *aux, *rest, *token = NULL;
```

chall11/gsanchez.c
```diff
@@ -0,0 +1,77 @@
...
+      else if (strstr(arr[j], "goto"))
+      {
+        indextogo = atoi(arr[j]+4);
+        j = 0;
+        while (++j < i)
+          if (indextogo == atoi(arr[j]))
+              break;
+        if (j == i)
+          return (strdup("Infinite loop !"));
+        continue;
```
aguimbao, line 70: I don't really like `continue` statements either, they are a bit confusing and they hurt readability in my opinion.  Maybe try to refactor it without using it ?

chall11/gsanchez.c
```diff
@@ -0,0 +1,77 @@
...
+    result = NULL;
+    j = 0;
+    while (j < i)
+    {
+      if (isdigit(*arr[j]))
+      {
+        while (isdigit(*arr[j]))
+          arr[j]++;
+        rest = arr[j];
+        while (token = strtok_r(rest, " ", &rest))
```
aguimbao, line 49: Should be
```C
while ((token = strtok_r(rest, " ", &rest)))
```

Or compiler will kill you 😈

aguimbao (commented):
Your submission was really nice :)

Maybe your infinite loop check could be better if you recorded the jumps you have already done, as as soon as you are going to come back to somewhere you have already jumped to, just return `"Infinite loop !"` ? 

But overall really nice usage of `strtok_r()` and `asprintf()`, you pass all the tests !

GonzSanch:
Thank you so much for the correction, guys! I Don't know why I mess up the PR but looks like it doesn't merge to the master.
I make this one from myFork:master and the 12th from myFork:newBranch...

I could create new branches for any chall and PR to keep it up the good practiceses with github workflow.

aguimbao:
I submitted your code manually cause it didn't merge because you grouped your PRs for different challenges. Be sure to create new branches and new PRs next time to prevent conflicts <3

PR #267, opened 2020-05-21
Chall11 - cdorta-n

Top64:
Hello!

For this challenge, my first idea was to create a function that will do the goto jump (ft_goto_statement). I included it in a main function that when he encounter the statement, he will jump to the next point, and continue. There are 2 "main" function where this works, "goto_aux", and "goto_answer", one to know how long will be the string to return, and the second one, for fill this string.

The first part of both functions, is the jump system, the second one, the line check.


aguimbao:
Could you fix this PR or make a new one? There's multiple files submitted here. I only want the chall11 😄

chall11/cdorta-n.c
```diff
@@ -0,0 +1,197 @@
...
+/*
+** 'Label' -> Positive Integer
+** 'goto' -> goto-statement
+** 'whitespaces' -> only 1 space
+** 'end-of-line' -> \n
+*/
+
+char	*ft_goto_error()
+{
+	char *error = "Infinite loop !";
```
aguimbao, line 26: Don't return a mix of string literals (like in this line), and dynamically allocated memory. Because you don't know what we will do with the returned variable. If we tried to free it. Our program would crash in the case that you return string literals like here.

Much better to:
```C
return strdup("Infinite loop !");
```

chall11/cdorta-n.c
```diff
@@ -0,0 +1,197 @@
+/* ************************************************************************** */
```
aguimbao, line 1: Header is not needed don't worry 😛

chall11/cdorta-n.c
```diff
@@ -0,0 +1,197 @@
...
+** 'end-of-line' -> \n
+*/
+
+char	*ft_goto_error()
+{
+	char *error = "Infinite loop !";
+	return (error);
+}
+
+int		ft_goto_statement(const char *code, int num, int c)
```
aguimbao, line 30: You don't have to recode all this functions ! You can just use the official `atoi()` or similar functions in the challenges :) it's even advised, as you would have to use them in a real world environment.

chall11/cdorta-n.c
```diff
@@ -0,0 +1,197 @@
...
+		return (0);
+}
+
+int 	ft_goto_aux(const char *code)
+{
+	int c = 0;
+	int i = 0;
+	int num = 0;
+	
+	while (code[c] != '\0')
```
aguimbao, line 78: For me
```C
while (code[c])
```

is more readable 👍

chall11/cdorta-n.c
```diff
@@ -0,0 +1,197 @@
...
+	{
+		if (code[c] == 'g' && code[c + 1] == 'o' && code[c + 2] == 't' && code[c + 3] == 'o')
```
aguimbao, line 80: There's a couple ways to do this much cleaner, like 
```C
if (&code[c] + 4) == "goto" {
    // code
}
```

Or you could have used `sscanf()`, very powerful function.

chall11/cdorta-n.c
```diff
@@ -0,0 +1,197 @@
...
+		{
+			while (code[c] < '0' || code[c] > '9')
```
aguimbao, line 82: You should use `isdigit()`, much more readable.

chall11/cdorta-n.c
```diff
@@ -0,0 +1,197 @@
...
+				c++;
+			while (code[c] >= '0' && code[c] <= '9')
```
aguimbao, line 84: You should use `isdigit()`, much more readable.

chall11/cdorta-n.c
```diff
@@ -0,0 +1,197 @@
...
+			}
+			if ((ft_goto_statement(code, num, c) < c))
+				return (-1);
+			c = ft_goto_statement(code, num, c);
+		}
+		else
+		{
+			if (code[c] != '\0' && code[c] != '\n')
+			{
+				while ((code[c] >= '0' && code[c] <= '9') || code[c] == ' ')
```
aguimbao, line 97:
```C
while (isdigit(code[c]) || isblank(code[c])) {
    // code
}
```

Is better 👍

chall11/cdorta-n.c
```diff
@@ -0,0 +1,197 @@
...
+		{
+			while (code[c] < '0' || code[c] > '9')
+				c++;
+			while (code[c] >= '0' && code[c] <= '9')
+			{
+				num = num * 10 + (code[c] - 48);
+				c++;
+			}
+			if ((ft_goto_statement(code, num, c) < c))
+				return (-1);
```
aguimbao, line 90: This doesn't really work because the labels weren't always in incremental order. The order didn't really matter (there was an example about this).

A good way to check this was to record the previous labels that you have performed a jump to, and every time you are gonna perform a `goto`, you check if it's in the list.

Top64:
I think I deleted the other file. But i'm not clear

aguimbao (commented):
Hey ! Your solution is pretty nice and it works for most cases. The problems I see is a bit of redundant code, a lot of this code is repetitive, and you could have used some oficial functions from the C standard libraries. 

Good job :) and nice PR fix.

aguimbao:
You had to delete the Chall06 submission from the PR, not from the repo. I added it again :)

Top64:
Thanks for the correction! Im looking to improve from this^^

PR #269, opened 2020-05-21
chall11 - pcuadrad

pabloocg:
First, I divided the string by '\ n' and removed the blank space around the words, leaving only one space between them.
Then, in the **treat_code** function, I go through the instructions and concatenate them, make a regular expression to find the 'goto' instruction in the string (maybe the regular expression is not useful in this case, but I have no experience with this and wanted try something different).
If there is a goto instruction, I go through the array again from the beginning looking for the new instruction to follow. I save the goto statement in an array for later comparison with another goto statement to see if an infinite loop can happen.

### DIFF
--- a/chall11/pcuadrad.c
+++ b/chall11/pcuadrad.c
@@ -1,0 +1,254 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <regex.h>
+
+#define isinset(a, b) (a == b) ? 1 : 0
+
+static int		count_str(char const *str, char c)
+{
+	int		count;
+
+	count = 0;
+	while (*str)
+	{
+		while (*str && isinset(*str, c))
+			str++;
+		if (*str && !isinset(*str, c))
+		{
+			count++;
+			while (*str && !isinset(*str, c))
+				str++;
+		}
+	}
+	return (count);
+}
+
+static char		*malloc_string(char *str, char c)
+{
+	char	*word;
+	int		i;
+
+	i = 0;
+	while (str[i] && !isinset(str[i], c))
+		i++;
+	if (!(word = (char *)malloc(sizeof(char) * (i + 1))))
+		return (NULL);
+	i = 0;
+	while (str[i] && !isinset(str[i], c))
+	{
+		word[i] = str[i];
+		i++;
+	}
+	word[i] = '\0';
+	return (word);
+}
+
+static char		**free_split(char ***tabstr)
+{
+	int i;
+
+	i = 0;
+	while (*tabstr[i])
+	{
+		free(*tabstr[i]);
+		i++;
+	}
+	free(*tabstr);
+	return (NULL);
+}
+
+char			**ft_split(char const *s, char c)
+{
+	int		len;
+	char	**tabstr;
+	int		i;
+
+	if (!s)
+		return (NULL);
+	len = count_str(s, c);
+	if (!(tabstr = (char **)malloc(sizeof(char*) * (len + 1))))
+		return (NULL);
+	i = 0;
+	while (*s)
+	{
+		while (*s && isinset(*s, c))
+			s++;
+		if (*s && !isinset(*s, c))
+		{
+			if (!(tabstr[i] = malloc_string((char*)s, c)))
+				return (free_split(&tabstr));
+			i++;
+			while (*s && !isinset(*s, c))
+				s++;
+		}
+	}
+	tabstr[i] = NULL;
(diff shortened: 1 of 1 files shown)


chall11/pcuadrad.c
```diff
@@ -0,0 +1,254 @@
...
+}
+
+static char		**free_split(char ***tabstr)
+{
+	int i;
+
+	i = 0;
+	while (*tabstr[i])
+	{
+		free(*tabstr[i]);
```
aguimbao, line 54: You can obv do this in one line 😛 
```C
while (*tabstr[i]) {
    free(*tabstr[i++]);
}
```

To me it looks better to try to always reduce the amount of lines of code, as long as it's readable.

chall11/pcuadrad.c
```diff
@@ -0,0 +1,254 @@
...
+    int     i;
+
+    i = -1;
+    while(gotos[++i])
+        if (!strcmp(gotos[i], newGoto))
+            return (1);
+    return (0);
+}
+
+char    *treat_code(char **instr)
```
aguimbao, line 163: I think I would have used `strstr()` or `sscanf()` if you wanted to try something different 😓 . It's a very powerful function.

chall11/pcuadrad.c
```diff
@@ -0,0 +1,254 @@
...
+            }
+        }
+        idx++;
+    }
+    if (last_pattern)
+        free(last_pattern);
+    return (ret);
+}
+
+void    free_all(char **tofree)
```
aguimbao, line 222: (For me) this would be more understandable:

```C
static void    free_all(char **tofree)
{
    size_t     i = 0;

    while (free(tofree[i++]));
    free(tofree);
}
```

chall11/pcuadrad.c
```diff
@@ -0,0 +1,254 @@
...
+
+char    *ft_goto_parser(const char *code)
+{
+    char    **parseCode;
+    char    *ret;
+    int     i;
+
+    if (!(parseCode = ft_split(code, '\n')))
+        return (NULL);
+    i = -1;
```
aguimbao, line 240: I don't really like negative indexes, I would also use `size_t` for all of them, as it let's you clearly see they are indexes. And you eliminate the risk of your index turning into a negative number.

chall11/pcuadrad.c
```diff
@@ -0,0 +1,254 @@
...
+			i++;
+			while (*s && !isinset(*s, c))
+				s++;
+		}
+	}
+	tabstr[i] = NULL;
+	return (tabstr);
+}
+
+char	*ft_strjoin(char const *s1, char const *s2)
```
aguimbao, line 90: Forgot to define as `static` this function and the rest of them from this point.

chall11/pcuadrad.c
```diff
@@ -0,0 +1,254 @@
...
+        ret = parseSpaces(parseCode[i]);
+        free(parseCode[i]);
+        parseCode[i] = ret;
+    }
+    if (!(ret = treat_code(parseCode)))
+    {
+        free_all(parseCode);
+        return (strdup("Infinite loop !"));
+    }
+    free_all(parseCode);
```
aguimbao, line 252: No need to free anything on these challenges, it's good practice though !

chall11/pcuadrad.c
```diff
@@ -0,0 +1,254 @@
...
+    i = -1;
+    while(gotos[++i])
+        if (!strcmp(gotos[i], newGoto))
+            return (1);
+    return (0);
+}
+
+char    *treat_code(char **instr)
+{
+    char    *ret = NULL;
```
aguimbao, line 165: There's no norm here, so it's actually more readable (personally), to group declarations/initializations like

```C
    char    *ret, *pattern, *tmp, **last_pattern = NULL;
    int     countGoto = 1;
    int     idx = 0;
```

aguimbao (commented):
I think you can reduce a lot the number of lines you used to solve this challenge.

It makes your code very readable and easy to understand but it's also more "cumbersome" to navigate.
I think regex was a bit of overkill, you could have simply used `sscanf()` for example.
The one other thing I didn't like is that you do a lot of work, you could have simply allocated space for the answer string, go through the string one time, and store the different words while you go through it, without extra splitting or joining anything. I also don't really like negative indexes, I think they are dangerous and very error prone ! All in all, I think there's just a little too much overhead in your program, else it's fine 👍 

But good job, you pass all the tests ;). Sorry for the late correction !

PR #270, opened 2020-05-22
chall11 - gbudau

gbudau:
1. Create a `copy `of the string on the heap.
2. Count the lines in the string and create an array of char pointers (`split`) that will be used to access each line.
3. Using `strsep()` to split the lines in the string `copy` and point the `split` pointers to each newline.
4. Create an array of pointers (`unique`) that will be initialized to NULL and used as a `set`.
5. Looping over the lines:
5.1.  If the line starts with a digit and that label is not in the set it will get added to the set, else if there is already in the set it means an infinite loop will happen.
5.2. Each time its processing a label also add it to `chain` string, replacing multiple whitespaces with a single space.
5.3. If the line don't start with a `digit`, then it should start with a `goto`, search for the label in `split` array and update the index in the main loop, repeat until all lines were processed or an error occurred.
6. Return the newly formated string if there were no errors or `NULL`/"`Infinite loop !`" otherwise.

Short description of the code:

`total_lines()` - return the number of lines in a string.

`split_lines()` - split lines in `copy` string and add them to the `split` array.

`free_ret()` - free the `set` and return `chain`/`NULL`/`"Infinite loop !"` based on a `flag` and a couple of `defines`.

`count_digits()` - count digits in a string.

`walk_goto()` -  skip digits in a string.

`goto_index()` - use the `line` string (the label after a goto) and compare it with the labels on the` split` array, `return` an index in the `split` array if found or `-1` if not found.

`fill_chain()` - using strsep() split the line(src) into words separated by space and copy them to the `chain`(dst) string, separating them by only one space. Return the updated `chain` index.

### DIFF
--- a/chall11/gbudau.c
+++ b/chall11/gbudau.c
@@ -1,0 +1,169 @@
+#include <stdlib.h>
+#include <string.h>
+#include <ctype.h>
+#define ERROR -1
+#define INFINITE_LOOP 1
+#define SUCCESS 0
+#define MAX(a,b) (((a)>(b))?(a):(b))
+
+static size_t	total_lines(const char *code)
+{
+	size_t	length;
+
+	length = 1;
+	for (; *code; code++)
+	{
+		if (*code == '\n')
+			length++;
+	}
+	return (length);
+}
+
+static void		split_lines(char *copy, char **split)
+{
+	char	*line;
+	size_t	i;
+
+	i = 0;
+	while ((line = strsep(&copy, "\n")) != NULL)
+	{
+		if (*line)
+		{
+			split[i] = line;
+			i++;
+		}
+	}
+	split[i] = NULL;
+}
+
+static char		*free_ret(char **unique, size_t length, int flag, char *chain)
+{
+	for (size_t i = 0; i < length; i++)
+		free(unique[i]);
+	free(unique);
+	if (flag == ERROR || flag == INFINITE_LOOP)
+	{
+		free(chain);
+		chain = NULL;
+	}
+	if (flag == INFINITE_LOOP)
+		return (strdup("Infinite loop !"));
+	return (chain);
+}
+
+static size_t	count_digits(const char *str)
+{
+	size_t	count;
+
+	count = 0;
+	for (int i = 0; isdigit(str[i]); i++)
+		count++;
+	return (count);
+}
+
+static char		*walk_goto(char *str)
+{
+	while (*str && isdigit(*str) == 0)
+		str++;
+	return str;
+}
+
+static ssize_t		goto_index(char **split, char *line)
+{
+	ssize_t	i = 0;
+	size_t	len1;
+	size_t	len2;
+
+	while (split[i] != NULL)
+	{
+		if (isdigit(*(split[i])))
+		{
+			len1 = count_digits(split[i]);
+			len2 = strlen(line);
+			if (strncmp(split[i], line, MAX(len1, len2)) == 0)
+				break;
+		}
+		i++;
+	}
+	free(line);
+	line = NULL;
+	return (split[i] == NULL ? -1 : i);
+}
+
+static size_t	fill_chain(char *dst, char *src, size_t index)
+{
+	char	*word;
+	size_t	len;
+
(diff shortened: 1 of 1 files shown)


chall11/gbudau.c
```diff
@@ -0,0 +1,169 @@
...
+#define SUCCESS 0
+#define MAX(a,b) (((a)>(b))?(a):(b))
+
+static size_t	total_lines(const char *code)
+{
+	size_t	length;
+
+	length = 1;
+	for (; *code; code++)
+	{
```
aguimbao, line 15: You can skip these brackets too

chall11/gbudau.c
```diff
@@ -0,0 +1,169 @@
...
+		if (*code == '\n')
+			length++;
+	}
+	return (length);
+}
+
+static void		split_lines(char *copy, char **split)
+{
+	char	*line;
```
aguimbao, line 24: You should initialise all your pointers to null when you declare them (when there's no norm), it's good practice.

chall11/gbudau.c
```diff
@@ -0,0 +1,169 @@
...
+	size_t	i;
+
+	i = 0;
+	while ((line = strsep(&copy, "\n")) != NULL)
+	{
```
aguimbao, line 29: You can skip this brackets.

chall11/gbudau.c
```diff
@@ -0,0 +1,169 @@
...
+		if (*line)
+		{
+			split[i] = line;
+			i++;
+		}
+	}
+	split[i] = NULL;
+}
+
+static char		*free_ret(char **unique, size_t length, int flag, char *chain)
```
aguimbao, line 39: Good practice, but we don't check leaks 😛

chall11/gbudau.c
```diff
@@ -0,0 +1,169 @@
...
+{
+	for (size_t i = 0; i < length; i++)
+		free(unique[i]);
+	free(unique);
+	if (flag == ERROR || flag == INFINITE_LOOP)
```
aguimbao, line 44: A bit of overkill but it would have been really cool if you used bit flags/masks for your flags, it's much more performing than using an `int`.

chall11/gbudau.c
```diff
@@ -0,0 +1,169 @@
...
+	size_t	len2;
+
+	while (split[i] != NULL)
+	{
+		if (isdigit(*(split[i])))
+		{
+			len1 = count_digits(split[i]);
+			len2 = strlen(line);
+			if (strncmp(split[i], line, MAX(len1, len2)) == 0)
+				break;
```
aguimbao, line 84: I don't like `break` statements, I would have tried to approach this differently so I don't need to use it.

aguimbao (commented):
So the problem with your solution is that there's a lot of overhead, you didn't need to split the original array, the performant way to solve was to walk the string directly and store the words on the fly with `asprintf()` or `strcat()` or similar.
Your solution is easy to read and easy to understand, there's just a bit too much of overhead for a simple problem (even though for a more complex problem, your approach would have been probably better).
Also I would try to save lines omitting brackets (`{}`) whenever possible, I think it doesn't affect readability that often, and group declaration statements.

Good job though, you pass all the tests 😝 and sorry for the late correction.

chall11/gbudau.c
gbudau: Thanks, i will keep it in mind for next time.

chall11/gbudau.c
gbudau: Good to know, thank you.

chall11/gbudau.c
gbudau: Will keep it in mind for next time :+1:

chall11/gbudau.c
gbudau: The only ways i can think of writing it without a `break` statement is to add either a flag or to `return` instead of `break`, and both cases need writing redundant code.
I will keep it in mind nonetheless.

gbudau:
Thank you Abraham, appreciate the feedback, very thorough!

PR #271, opened 2020-05-22
edramire - Saffar

edramir18:
My solution is based on the graph theory to determine infinite loops in the string provided. According to this, upon resolving the path from the start node to the last node, in case i visited a node already in the path, the input string has a infinite loop.
The information for each node is stored emulating a hash map, using a pointer to the original string as key and value of the map. In addition the node has a flag to determinate that has not yet been visited.
After build all nodes, the final string is created recorring the graph and any extra white space found is discarded.

### DIFF
--- a/chall11/edramire.c
+++ b/chall11/edramire.c
@@ -1,0 +1,144 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   edramire.c                                         :+:      :+:    :+:   */
+/*                                                    +:+ +:+         +:+     */
+/*   By: edramire <edramire@student.42.fr>          +#+  +:+       +#+        */
+/*                                                +#+#+#+#+#+   +#+           */
+/*   Created: 2020/05/21 23:56:08 by edramire          #+#    #+#             */
+/*   Updated: 2020/05/22 10:30:15 by edramire         ###   ########.fr       */
+/*                                                                            */
+/* ************************************************************************** */
+
+#include <stdlib.h>
+#include <string.h>
+#include <ctype.h>
+#include <stdio.h>
+
+typedef	struct	s_list{
+	char const		*key;
+	size_t			key_len;
+	char const		*value;
+	size_t			value_len;
+	int				visited;
+	struct s_list	*next;
+	struct s_list	*goto_line;
+}				t_list;
+
+static int		ft_stc_free_list(t_list *lst)
+{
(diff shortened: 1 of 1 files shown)


chall11/edramire.c
```diff
@@ -0,0 +1,144 @@
+/* ************************************************************************** */
```
aguimbao, line 1: Not needed, but gives your file a lot of style 😎

chall11/edramire.c
```diff
@@ -0,0 +1,144 @@
...
+{
+	size_t len;
+
+	len = 0;
+	if (isdigit(*node->key))
+	{
+		node->goto_line = node->next;
+		return ;
+	}
+	while (*node->value == ' ')
```
aguimbao, line 78: `isblank()` ?

chall11/edramire.c
```diff
@@ -0,0 +1,144 @@
...
+	t_list	*node;
+	t_list	*last;
+
+	list = NULL;
+	while (*code != '\0')
+	{
+		node = malloc(sizeof(t_list));
+		if (node == NULL && ft_stc_free_list(list))
+			return (NULL);
+		bzero(node, sizeof(t_list));
```
aguimbao, line 53: `bzero()` is deprecated, I would use `memset()`, it's more standard

chall11/edramire.c
```diff
@@ -0,0 +1,144 @@
...
+static t_list	*ft_stc_load(const char *code)
+{
+	t_list	*list;
+	t_list	*node;
+	t_list	*last;
+
+	list = NULL;
+	while (*code != '\0')
+	{
+		node = malloc(sizeof(t_list));
```
aguimbao, line 50: Why not `calloc()` if you're gonna `bzero()` it ?

aguimbao (commented):
Your solution is really good. It's probably not the most performant cause of all the overhead of the structures setup, but it's definitely a smart way to do it, you could have maybe made a hashtable with the `goto` lines and the "normal lines" maybe, idk.

Couple things I would have done differently, is group initializations of the same kind, like:
```C
typedef	struct	s_list{
	char const		*key, *value;
	size_t			key_len, *value_len;
	int				visited;
	struct s_list	*next, *goto_line;
}	
```

Also I would initialize all pointers to NULL just in case, I think it's good practice.
And no need to free stuff, we don't check for leaks :) (but it's good practice too).

You pass everything, good job, really liked this solution !

PR #272, opened 2020-05-22
Ssacrist Chall11

esepunto:
My approach 

1.- Repeated blanks are removed.
2.- A structure is created to store different strings and variables.
3.- It looks for the first line and saves it in a temporary string ('temp').
4.- Check the temporary string to find out if it is "goto" or a line to be printed.
5.- If it has to be printed, it divides it. In a string it saves the line number (in this string all the line numbers will be saved consecutively). And in another string ('result') it saves the text to be printed.
6.- If it is a "goto" string, it compares the line with the string where the line numbers have been saved. If it is there, print "Infinite loop !" and leave the program. Otherwise, it searches in the copy of the source string ("duplicate") that line and continues printing.
7.- The program will continue printing until it reaches the end of the copy of the source string ("duplicate") .

The program does not work properly, I would need more time to debug it.


chall11/ssacrist.c
```diff
@@ -0,0 +1,173 @@
...
+            i++;
+        }
+        while (carrier->temp[i] != '\0')
+        {
+            carrier->result[carrier->positioin_result] = carrier->temp[i];
+            i++;
+            carrier->positioin_result++;
+        }
+    }
+    print_struct(carrier);
```
aguimbao, line 112: You forgot all this logs on the final version :(

chall11/ssacrist.c
```diff
@@ -0,0 +1,173 @@
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+#include <ctype.h>
+
+typedef struct	s_format
```
aguimbao, line 6: I would group this as 

```C
typedef struct	s_format
{
    char    *result, *num_lines, *str_go_to, *duplicate, *tmp = NULL;
    int     position, position_result, len;
}				t_format;
```

It looks more readable to me (not to the norm but yeah 😓 )

chall11/ssacrist.c
```diff
@@ -0,0 +1,173 @@
...
+    carrier->num_lines = strdup("");
+    carrier->str_go_to = strdup("");
+    carrier->temp = strdup("");
+    carrier->position = 0;
+    carrier->positioin_result = 0;
+}
+
+void	print_struct(t_format *carrier)
+{
+	printf("\n");
```
aguimbao, line 30: I would have put all of this in a single `printf()` statement

chall11/ssacrist.c
```diff
@@ -0,0 +1,173 @@
...
+
+    i = 0;
+    while (carrier->duplicate[i] != '\0')
+    {
+        while (isblank(carrier->duplicate[i]) && isblank(carrier->duplicate[i + 1]))
+        {
+            j = i;
+            while (j <= carrier->len)
+            {    
+                carrier->duplicate[j] = carrier->duplicate[j + 1];
```
aguimbao, line 54: There's a big overflow here, if 

```C
carrier->len == 10;
j = 9;
```

In this case when you do 
```C
carrier->duplicate[9] = carrier->duplicate[10];
```

You overflow.
And if `j == carrier->len`, you overflow x2 

Gives segfault on almost every test case.

aguimbao (commented):
Oh sorry I hadn't seen that you hadn't finished your submission :( that's a shame.

Couple things I saw, maybe use `static` keyword for your functions, group declarations of the same type of variable and avoid brackets when possible to save lines (there's no norminette 😝 ).

You had a good approach separating the string, you should take a look at `edramire`'s solution as he does this too with structs and it works really well. The only thing is that this kind of approach won't be super performant, you could have done going through the string a single time and concatenating the answer on the fly.

Nice try though ! Good luck on the next one ;)

chall11/ssacrist.c
esepunto: I didn't know this way to declaring variables. Thank you!

esepunto:
Sorry, I didn't realize you'd left more time to finish the chall :( 
Many thanks in any case! :)

PR #273, opened 2020-05-25
Chall12 migferna

migferna42:
We generate 2 iterators, we place the first iterator where the opening parenthesis starts, and we place the second iterator in the closing parenthesis that is within the same level as the opening parenthesis.

If we find nested parentheses, we raise and lower the level.
When finding the closing parenthesis, the string is reversed, the level is reset and the search continues from the next position.

_Note to the evaluator:
**Commits from 53b2f94343e, I know they are out of time, but it is a challenge that has caught my attention and I want it to work well**_

### Another way
I have been working is to do it by recursion, we must go down the tree to a base case where the string only has an opening and closing parenthesis.


clafoutis42:
Given the late commits I have to refuse the PR. Also, given the obfuscated code, I won't review it, I don't have that time. (If you think that doing `i[str]` is cool it isn't)

PR #274, opened 2020-05-25
edramire - AB Aurigae

edramir18:
To be able to resolve the challenge for today, i used queues and stacks to reverse the order of the string.
The procedure is silimar to the one used to evaluate prefix/postfix  expressions, each letter is pushed to the stack until a closing parentheses is found. In that case each letter already in the stack is pushed to an auxiliar stack until an opening parentheses not already used is found. After that each character in the auxiliar stack is extracted as a queue, reversing the order of the parentheses found along the way, and pushed again to the original stack.
I set a flag in each node to determine whenever an opening parentheses was already use or not. In addition a counter for each parentheses found is used to determine whenever exist a mismatch in the number of them.
Finally the original stack is traversed as a queue and each character is copied to the final string.

### DIFF
--- a/chall12/edramire.c
+++ b/chall12/edramire.c
@@ -1,0 +1,176 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   edramire.c                                         :+:      :+:    :+:   */
+/*                                                    +:+ +:+         +:+     */
+/*   By: edramire <edramire@student.42.fr>          +#+  +:+       +#+        */
+/*                                                +#+#+#+#+#+   +#+           */
+/*   Created: 2020/05/25 13:54:51 by edramire          #+#    #+#             */
+/*   Updated: 2020/05/25 16:46:16 by edramire         ###   ########.fr       */
+/*                                                                            */
+/* ************************************************************************** */
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+
+typedef struct		s_list
+{
+	char			letter;
+	int				operator;
+	struct s_list	*next;
+}					t_list;
+
+static int			ft_stc_free(t_list *list, char *response)
+{
+	t_list	*aux;
+
+	free(response);
+	while (list != NULL)
+	{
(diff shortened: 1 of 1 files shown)


chall12/edramire.c
```diff
@@ -0,0 +1,176 @@
...
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+
+typedef struct		s_list
+{
+	char			letter;
+	int				operator;
+	struct s_list	*next;
+}					t_list;
```
clafoutis42, line 22: That's in principle a nice idea, but really it's too complex, and overkill, also, you will lose a lot of time parsing and rebuilding the string.

chall12/edramire.c
```diff
@@ -0,0 +1,176 @@
...
+	char		*response;
+
+	list = NULL;
+	node = NULL;
+	operators = 0;
+	if (str == NULL)
+		return (NULL);
+	response = malloc(strlen(str) + 1);
+	if (response == NULL)
+		return (NULL);
```
clafoutis42, line 142: You should check for syntax errors BEFORE to avoid allocating things you will not use

chall12/edramire.c
```diff
@@ -0,0 +1,176 @@
...
+			ft_stc_reverse_stack(&list);
+		}
+		else if (node->letter == '(')
+		{
+			++operators;
+			ft_stc_push(&list, node);
+		}
+		else
+			ft_stc_push(&list, node);
+		++str;
```
clafoutis42, line 162: It is intelligent to use chained lists, but you are not getting anything from it here. As you basically are creating a node by characters, your chained lists behaves just like a string, but is more greedy in memory, and it takes more time to iterate on it.  
What you did here could have been done exactly the same with a string. But it would have been faster.

An intelligent way of using lists here would have been to create a node by substring (implied by the parentheses).

clafoutis42 (commented):
You have a good algorithm that works, but using chained lists here is overkill. Furthermore you did not take advantage of them.  
But good job!

PR #275, opened 2020-05-25
vde-dios.c

vdedios:
### SOLUTION

The program goes through the string. Whenever the first `(` is encountered it means we have reached a new level and, therefore, we now should find its closing brace to get the string to be swapped. 

In order to achieve it we use a counter. When we find an opening brace, counter is increased, if is a closing one, counter is decreased. If the counter reaches `zero` value, it means we have a string to swap and we pass it, along with the opening and close positions to  `swap_str` which will swap the string between these positions. (braces are kept the same way)


chall12/vde-dios.c
```diff
@@ -0,0 +1,55 @@
...
+        i++;
+    }
+    if ((finish - start + 1) % 2)
+        str[i] == ')' ? str[i] = '(' : (str[i] == '(' ? str[i] = ')' : str[i]);
+}
+
+char *ft_reverse_parenthesis(const char *str)
+{
+    int i, j, level;
+    char *aux = strdup(str);
```
clafoutis42, line 25: What if it fails?

chall12/vde-dios.c
```diff
@@ -0,0 +1,55 @@
...
+
+    i = -1;
+    level = 0;
+    while (str[++i])
+        str[i] == '(' ? level++ : (str[i] == ')' ? level-- : level);
```
clafoutis42, line 30: Using ternaries can help you gaining lines, but be careful not to hurt your code's readability!  
Also, with this syntax check, it seems you're not handling the parentheses being in that order `)    (` !  
Example of string that fails:
```
never close ) before opening (
```

chall12/vde-dios.c
```diff
@@ -0,0 +1,55 @@
...
+    {
+        aux = str[i];
+        str[i] = str[finish + start - i];
+        str[i] == ')' ? str[i] = '(' : (str[i] == '(' ? str[i] = ')' : str[i]);
+        aux == ')' ? aux = '(' : (aux == '(' ? aux = ')' : aux);
+        str[finish + start - i] = aux;
+        i++;
+    }
+    if ((finish - start + 1) % 2)
+        str[i] == ')' ? str[i] = '(' : (str[i] == '(' ? str[i] = ')' : str[i]);
```
clafoutis42, line 19: You could have swapped while copying to avoid writing the same string over over and over ;)

clafoutis42 (commented):
Your algorithm seems to work, but your syntax checker does not work, there are some edge cases you did not think about, better luck next time!

vdedios:
:(((

chall12/vde-dios.c
vdedios: I do not really understand what  do you mean here. Do you mean if str is an invalid input? Do you suggest checking its validity before duplicating it?

chall12/vde-dios.c
vdedios: I will account for that next time. I didn't feel very confy nesting several ternary but just wanted to try to see how it worked..

About the parenthesis error... There's ever enough tests, I am planning to tatoo it on my skin 😅

chall12/vde-dios.c
vdedios: I have been thinking about how to swap the strings without using an aux var and haven't got to any solution... What do you mean with this?

PR #276, opened 2020-05-25
chall12 | estina

xplore93:
My version of the solution. First, I check if the str is not NULL and if the parentesis are good. Then with strchr I find first '(' and search the last valid ')' so then I make a reverse of the string between this two parentesis. Repeat it with next all the string - first '(' until all parentesis end. Then return the reversed line.

### DIFF
--- a/chall12/estina.c
+++ b/chall12/estina.c
@@ -1,0 +1,78 @@
+#include <string.h>
+#include <stdio.h>
+#include <unistd.h>
+
+static void	ft_turn_parentesis(char *c)
+{
+	*c = *c == '(' ? ')' : *c == ')' ? '(' : *c;
+}
+
+static char* ft_reverse(char *str, int len)
+{
+	int		i;
+	char	c;
+
+	i = len / 2 - 1;
+	while (i >= 0)
+	{
+		ft_turn_parentesis((str + i));
+		c = str[i];
+		str[i] = str[len - i - 1];
+		str[len - i - 1] = c;
+		ft_turn_parentesis((str + i));
+		i--;
+	}
+	return (str);
+}
+
+static void	pass_flag(char c, int *flag, int *open)
+{
+	if (c == ')' && *open)
+	{
+		*open = 0;
+		(*flag)--;
+	}
+	else if (c == '(')
+	{
+		*open = 1;
+		(*flag)++;
+	}
+}
+
+static int	count_parentesis(const char *str)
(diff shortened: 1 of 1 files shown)


chall12/estina.c
```diff
@@ -0,0 +1,78 @@
+#include <string.h>
+#include <stdio.h>
+#include <unistd.h>
+
+static void	ft_turn_parentesis(char *c)
+{
+	*c = *c == '(' ? ')' : *c == ')' ? '(' : *c;
```
clafoutis42, line 7: Using ternaries can help you gaining lines, but be careful not to hurt your code's readability!  
Especially since you placed it in a function

chall12/estina.c
```diff
@@ -0,0 +1,78 @@
...
+	{
+		*open = 1;
+		(*flag)++;
+	}
+}
+
+static int	count_parentesis(const char *str)
+{
+	int		i;
+	int		flag, open;
```
clafoutis42, line 45: It's a bad practice to name a variable as a builtin (`open`)

chall12/estina.c
```diff
@@ -0,0 +1,78 @@
...
+	if (c == ')' && *open)
+	{
+		*open = 0;
+		(*flag)--;
+	}
+	else if (c == '(')
+	{
+		*open = 1;
+		(*flag)++;
+	}
```
clafoutis42, line 39: I'm not sure to get it, but it seems that if you open 2 parentheses, and close only one, open becomes `0` and you can't close any parentheses anymore?

chall12/estina.c
```diff
@@ -0,0 +1,78 @@
...
+}
+
+char		*ft_reverse_parenthesis(const char *str)
+{
+	char	*copy, *ptr;
+	int		i, flag, open;
+
+	if (!str || !*str || count_parentesis(str))
+		return (NULL);
+	copy = strdup(str);
```
clafoutis42, line 61: What if it fails?

chall12/estina.c
```diff
@@ -0,0 +1,78 @@
...
+	ptr = copy;
+	while (1)
```
clafoutis42, line 63: You could have just done `while (ptr)` or something like that, this is really ugly

clafoutis42 (commented):
Your code is a little odd, and doesn't work with the first example. Better luck next time.

PR #277, opened 2020-05-25
Chall12 - gsanchez

GonzSanch:
First I check if the string provided is valid, counting the parenthesis and checking if were opened before or doesn't have one.  Then for reverse I iterate and look for the first opened parenthesis and the last closed parenthesis, with those index I reverse the string  and iterate again till all parenthesis are gone.

Finally I changed the reverse algorithm because in some cases the last parenthesis doesn't match witth the first, so I used the similar for the missmatch to find the parenthesis till the end of the string.


chall12/gsanchez.c
```diff
@@ -0,0 +1,101 @@
...
+	char	*reverse;
+	int		start, end;
+	int		i = 0;
+	int		missmatch, paren;
+
+	paren = 0;
+	missmatch = 0;
+	if (!str || ft_searchunmatch(str))
+		return (NULL);
+	reverse = strdup(str);
```
clafoutis42, line 72: What if it fails?

chall12/gsanchez.c
```diff
@@ -0,0 +1,101 @@
...
+	if (!strchr(reverse, 40))
```
clafoutis42, line 73: Why not directly using `'('` instead of `40` ? It helps readability

chall12/gsanchez.c
```diff
@@ -0,0 +1,101 @@
...
+{
+	int paren = 0;
+	int open = 0;
+	char *str = (char*) stro;
+
+	while (*str)
+	{
+		if (*str == '(')
+			paren++;
+		else if (*str == ')' && paren > 0)
```
clafoutis42, line 27: Watch out here, you won't know if there is an unmmatch `)`, which means that you fail this test:
```
never close ) without opening
```

chall12/gsanchez.c
```diff
@@ -0,0 +1,101 @@
...
+			}
+		}
+		if (!missmatch && paren)
+		{
+			i = start;
+			paren = 0;
+			ft_strreverse(reverse, start, end);
+		}
+		i++;
+	}
```
clafoutis42, line 99: Nice algorithm. The problem I see here is that you are writing many times the same string. I would suggest you to reverse while reading the input string. This is a little more technical, but way faster!

clafoutis42 (commented):
Nice algorithm, however the syntax checking has some flaws. Good job anyway!

PR #278, opened 2020-05-25
chall12 - jbaringo

baringo11:
First I check if there is the correct number of parenthesis, if this is the case, I return the max number of left parenthesis and if there is not, I return NULL. Then I go over the string until I find the last '(' and then I also find the correspondent ')' to close the parenthesis. After that I send the string between that parenthesis to the "reverse_str" function. I reverse that sting and I add it into the original string. Finally I rest one to the max number of left parenthesis for doing the same steps to the previous parenthesis in the next iteration. When there is not more parenthesis I return the string reversed.

### DIFF
--- a/chall12/jbaringo.c
+++ b/chall12/jbaringo.c
@@ -101,7 +101,7 @@ char *ft_reverse_parenthesis(const char *str)
     int flag = 0;
     int j = 0;
 
-    if ((parenthesis = check_parenthesis(s)) == -1)
+    if (!str || (parenthesis = check_parenthesis(s)) == -1)
         return(strdup("NULL"));
     while (parenthesis > 0)
     {
--- a/chall12/jbaringo.c
+++ b/chall12/jbaringo.c
@@ -1,0 +1,134 @@
+#include <stdlib.h>
+#include <string.h>
+
+char *ft_substr(char *str, int start, int finish)
+{
+	char *s;
+	int i = 0;
+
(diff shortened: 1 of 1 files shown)


chall12/jbaringo.c
```diff
@@ -0,0 +1,134 @@
...
+        j++;
+        i++;
+    }
+    s2[j] = '\0';
+    return (s2);
+}
+
+char *ft_reverse_parenthesis(const char *str)
+{
+    char *s = strdup(str);
```
clafoutis42, line 97: What if it fails?

Also you should check the syntax before allocating the string you want to return, otherwise you allocate for nothing.

chall12/jbaringo.c
```diff
@@ -0,0 +1,134 @@
...
+    int parenthesis;
+    int i = 0;
+    int left = 0;
+    int flag = 0;
+    int j = 0;
+
+    if (!str || (parenthesis = check_parenthesis(s)) == -1)
+        return(strdup("NULL"));
```
clafoutis42, line 105: You should return `NULL`, not a string `"NULL"` :/

clafoutis42 (commented):
Your algorithm seems fine, however you return a string instead of returning `NULL`.

A better approach of the problem would have been to directly make the reverse while reading the string, you want rapidity and to avoid allocating too much memory (which you never free and this is wrong).

PR #279, opened 2020-05-25
szavarce chall12

szavarce42:
My approach:

1. I create a copy of the string
2. We call a function to count every close/open parenthesis. They must match. Otherwise the input will be considered invalid.
3. Then we loop through the string and stop at every opened parenthesis
4. Now we loop to find a matching ending parenthesis by counting every open parenthesis, and substracting every close parenthesis. Every time we reach 0, we are finding a completed parenthesis.
5. We call a function to reverse the text we found between those parenthesis.
6. We get back to the position where we found the previous parenthesis to keep looping, so we dont leave anything behind.

### DIFF
--- a/chall12/szavarce.c
+++ b/chall12/szavarce.c
@@ -53,7 +53,7 @@ char *ft_reverse_parenthesis(const char *str)
 
 	start_pos = 0;
 	end_pos = 0;
-	if (count_chars(str, '(') != count_chars(str, ')'))
+	if (!(str) || (count_chars(str, '(') != count_chars(str, ')')))
 		return (0);
 	if (!(new_str = malloc(sizeof(char) * strlen(str))))
 		return (0);
--- a/chall12/szavarce.c
+++ b/chall12/szavarce.c
@@ -27,6 +27,22 @@ void	reverse_text(char *str, int start, int end)
 	}
 }
 
+int		check_first_parenthesis(const char *str)
+{
+	int		pos;
+
+	pos = 0;
+	while (str[pos] && str[pos] != '\0')
+	{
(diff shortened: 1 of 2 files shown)


chall12/szavarce.c
```diff
@@ -0,0 +1,102 @@
...
+	int		start_pos;
+	int		end_pos;
+
+	start_pos = 0;
+	end_pos = 0;
+	if (!(str) || (count_chars(str, '(') != count_chars(str, ')')))
+		return (0);
+	if (check_first_parenthesis(str) == 0)
+		return (0);
+	if (!(new_str = malloc(sizeof(char) * strlen(str))))
```
clafoutis42, line 76: You are not allocating the space for the null terminator of the string (`\0`) this is causing a segfault...

chall12/szavarce.c
```diff
@@ -0,0 +1,102 @@
...
+    int		count;
+	char	*new_str;
+	int		start_pos;
+	int		end_pos;
+
+	start_pos = 0;
+	end_pos = 0;
+	if (!(str) || (count_chars(str, '(') != count_chars(str, ')')))
+		return (0);
+	if (check_first_parenthesis(str) == 0)
```
clafoutis42, line 74: That's cool to check the first parentheses, but what if I do this:
```
never close () before) opening (
```
You don't check the second one, which is invalid (this later cause your program to segfault as you did not expect this)

clafoutis42 (commented):
If you forget the little errors causing segfaults, your program seems fine. However I would have suggested doing the reverse whilst parsing to avoid multiple writes of the same things.

PR #280, opened 2020-05-25
chall12 - gbudau

gbudau:
**How i solved this problem**  /  **Short description of the code**
1. Check if all open parentheses have an closing pair, then copy the string on the heap using `strdup()`.
2. Save the indexes after an open parenthesis on a `stack` array.
3. Loop over the copy and if the current character is an open parenthesis save the `index + 1` on the `stack`.
4. If the current character is a closing parenthesis first invert all parentheses inside the current parentheses then reverse the string inside the parentheses.
5. Free the `stack` and return the modified string.


chall12/gbudau.c
```diff
@@ -0,0 +1,117 @@
+#include <stdlib.h>
+#include <stdbool.h>
+#include <string.h>
+
+static bool		valid_parenthesis(const char *str)
+{
+	size_t	open;
```
clafoutis42, line 7: It is bad practice to name a variable as `open` (as `open` is meant to open files)

chall12/gbudau.c
```diff
@@ -0,0 +1,117 @@
...
+
+	open = false;
```
clafoutis42, line 9: Here you are using `open` as a counter, assigning it `false` is confusing.

chall12/gbudau.c
```diff
@@ -0,0 +1,117 @@
...
+{
+	while (start <= end)
+	{
+		if (str[start] == ')')
+			str[start] = '(';
+		else if (str[start] == '(')
+			str[start] = ')';
+		start++;
+	}
+}
```
clafoutis42, line 77: You could do this while copying instead of going through you string all over again

clafoutis42 (commented):
Your algorithm is correct, congrats you seem to pass all the tests !  
However you could have optimized a little by reversing while reading, but this is really just optimization.

chall12/gbudau.c
gbudau: My bad, didn't think about it.

chall12/gbudau.c
gbudau: Thank you, will keep it in mind for next time.

gbudau:
Thank you Charles! I will try to optimize more in the future :+1: .

PR #281, opened 2020-05-26
chall12 - vsempere - Not working solution

victorsempere:
Hi all,

This is just to git flow and to show that I tried to solve it.

The idea behind the solution was:
Malloc an array.
Start copying linearly from the src until a parenthesis is found
Then I call swap.
This swap method will swap char/char, char/block, block/char, block/block.
A block is detected when found a ( in the left index or ) in the right index.
Each block is solved calling to swap recursivelly
After return a linear swap of the internal blocks will be done
If everything goes ok the malloced array it was supposed to contain the result.
If something goes wrong a NULL message will be returned

### DIFF
--- a/chall12/vsempere.c
+++ b/chall12/vsempere.c
@@ -1,0 +1,209 @@
+#include <stdlib.h>
+#include <string.h>
+
+char *ft_reverse_parenthesis(const char *str);
+static char *reverse(const char *str, int size);
+static int ft_reverse_parenthesis_left(char *const str, int pos, int *end_block_left);
+static int ft_reverse_parenthesis_right(char *const str, int pos, int len, int *ini_block_right);
+
+static void ft_swap_blocks(char *const str, int end_left_block, int ini_right_block)
+{
+    char *tmp;
+    int len;
+
+    len = strlen(str);
+    tmp = calloc(len + 1, sizeof(char));
+    strcat(tmp, &str[len - ini_right_block]);
+    strncat(tmp, &str[end_left_block], ini_right_block - end_left_block);
+    strncat(tmp, str, end_left_block);
+    memmove(str, tmp, len);
+    free(tmp);
+    tmp = 0;
+}
+
+static int ft_look_for_left_parenthesis(const char *str, int start)
+{
+    int open_parenthesis;
+    int i;
+
+    open_parenthesis = 0;
+    i = start;
+    while (i >= 0) {
+        if (str[i] == ')')
+            open_parenthesis++;
+        else if (str[i] == '(')
+        {
+            if (!open_parenthesis)
+                return (i);
+            else
+                open_parenthesis--;
+        }
+        
+        i--;
+    }
+    return (-1);
+}
+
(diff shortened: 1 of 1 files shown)


chall12/vsempere.c
```diff
@@ -0,0 +1,209 @@
+#include <stdlib.h>
+#include <string.h>
+
+char *ft_reverse_parenthesis(const char *str);
+static char *reverse(const char *str, int size);
+static int ft_reverse_parenthesis_left(char *const str, int pos, int *end_block_left);
+static int ft_reverse_parenthesis_right(char *const str, int pos, int len, int *ini_block_right);
```
clafoutis42, line 7: If you keep a good order on declarations, you don't have to do this (on the same file I mean)

chall12/vsempere.c
```diff
@@ -0,0 +1,209 @@
...
+    int i;
+    char *rev;
+    int end_block_left;
+    int ini_block_right;
+
+    if (!str)
+        return strdup("NULL");
+    len = strlen(str);
+    if (!(rev = calloc(len + 1, sizeof(char))))
+        return strdup("NULL");
```
clafoutis42, line 151: Why are you returning a string when you hust had to return `NULL` which is a macro to the `0x0` pointer ?

chall12/vsempere.c
```diff
@@ -0,0 +1,209 @@
...
+    memcpy(rev, str, len);
```
clafoutis42, line 152: You could have used `strdup` instead of allocating then copying !

chall12/vsempere.c
```diff
@@ -0,0 +1,209 @@
...
+    char block_chr;
+
+    *end_block_left = ft_look_for_right_parenthesis(&str[i + 1]);
+    if (*end_block_left < 0)
+        return (0);
+    block_chr = str[i + 1 + *end_block_left + 1];
+    str[i + 1 + *end_block_left + 1] = 0;
+    if (!(block = ft_reverse_parenthesis(&str[i + 1])))
+        return (0);
+    memcpy(&str[i + 1], block, *end_block_left);
```
clafoutis42, line 115: Why using `memcpy` when you are just treating strings ? You could have used `strncpy` to copy just a part of the string !

chall12/vsempere.c
```diff
@@ -0,0 +1,209 @@
...
+}
+
+static char *reverse(const char *str, int size)
+{
+    int i;
+    int block_end;
+    char *ret;
+
+    if (!(ret = calloc(strlen(str) + 1, sizeof(char))))
+        return 0;
```
clafoutis42, line 76: It seems easier to allocate the part you want to reverse, but it will lead to many allocation, and will utlimately be slow and greedy. You might want to simply clone the string once and reverse directly each time you encounter a parentheses. Or better you could directly do the reverse while reading and cloning the input.

clafoutis42 (commented):
Thanks for submitting your solution even if it's not over.  
Better luck next time !

PR #282, opened 2020-05-26
Chall12

abarral-nav:


abarral-nav:
Forget to make the description. My program copy the string and runs through that copy to found '(', count the parentheses and if the number of '(' and ')' is different it returns NULL. Swap characters inside parentheses except '(' and ')'.

clafoutis42:
You pushed a .txt file. I can't accept your PR, sorry :(

PR #283, opened 2020-05-26
Chall12 jtrancos

jtrancoso:
This is my try for Chall12. Tried to work around the `return NULL` rule since I couldn't make it work.

First we check if the string is valid, to do so we look for pairs of `(` and `)`, if we have an odd count, we mark it as invalid and return `NULL`.

After that we do an `strdup` of the initial string. We look for a `(`, if we don't find any we return the string just as it is.

If we have found a `(` we save its position and then start looking for its closing pair. Once we've got them both we reverse everything that's inside it and we reverse again each individual bracket so they face the right direction after that. Then we continue iterating for each pair of brackets we find until we get to the centre of the string.

### DIFF
--- a/chall12/jtrancos.c
+++ b/chall12/jtrancos.c
@@ -1,0 +1,98 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   jtrancos.c                                         :+:      :+:    :+:   */
+/*                                                    +:+ +:+         +:+     */
+/*   By: jtrancos <marvin@42.fr>                    +#+  +:+       +#+        */
+/*                                                +#+#+#+#+#+   +#+           */
+/*   Created: 2020/05/25 13:34:09 by jtrancos          #+#    #+#             */
+/*   Updated: 2020/05/26 11:32:43 by jtrancos         ###   ########.fr       */
+/*                                                                            */
+/* ************************************************************************** */
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+
+static void	swap(char *a, char *b)
+{
+	char	temp;
+
+	temp = *a;
+	*a = *b;
+	*b = temp;
+}
+
+static void	reversing(char *str, int start, int end)
+{
+	int	q;
+	int	i;
+
+	i = start;
+	q = end;
+	while (i < end)
+	{
+		swap(&str[i], &str[end]);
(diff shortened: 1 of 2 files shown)


jtrancoso:
RIP deadline, took to long writing the description.

chall12/jtrancos.c
```diff
@@ -0,0 +1,98 @@
...
+	int	end;
+	int	i;
+	int	q;
+
+	i = 0;
+	q = strlen(str);
+	if (check_error(str) == 0)
+		temp = strdup(str);
+	else
+		temp = strdup("NULL");
```
clafoutis42, line 82: What is this ? 😅You want to return `NULL` not `"NULL"`.

chall12/jtrancos.c
```diff
@@ -0,0 +1,98 @@
...
+	char	*temp;
+	int	start;
+	int	end;
+	int	i;
+	int	q;
+
+	i = 0;
+	q = strlen(str);
+	if (check_error(str) == 0)
+		temp = strdup(str);
```
clafoutis42, line 80: What if the dup fails and you get `NULL` ?

chall12/jtrancos.c
```diff
@@ -0,0 +1,98 @@
...
+	else
+		temp = strdup("NULL");
+	if (strchr(temp, '(') == NULL)
+		return (temp);
+	while (temp[i])
+	{
+		while (temp[i] != '(' && i < q)
+			i++;
+		start = i;
+		i++;
```
clafoutis42, line 90: Watch out, if you come to the end of the string on this subloop, you will increment `i` one more time and be passed the null terminator, which means that you will segfault on the main loop.

clafoutis42 (commented):
Nice try ! Too bad that you took too long to write the description :(  
Watch out for buffer overflows !

chall12/jtrancos.c
jtrancoso: Yeah, should have covered it with a `if strdup == NULL`

chall12/jtrancos.c
jtrancoso: Don't know why it would not return NULL so I tried to be creative about it 😅

jtrancoso:
Thanks for the feedback! Very knowledgeable as always :) Next time I will upload it and edit the description afterwards

